PR #414, opened 2014-10-03
Generate a `rails-api` app with `--app`

seanpdoyle:
Hangs locally, only pushing to let CI run it


ACPK:
@derekprior - Any chance I could check out the sd-api-generator branch? I absolutely love where you were going with it! :-)

Issue #427, opened 2014-10-31
[feature] Specify heroku region


lenart:
For those of us who live in Europe it would be great if we could pass an optional parameter to heroku apps:create command to choose region. Something like: `suspenders my_app --heroku true --heroku-region eu`

croaky:
Thanks, @lenart. This would be nice to support. I think we'll get this in via https://github.com/thoughtbot/suspenders/pull/468

PR #448, opened 2014-11-23
Allow --heroku-region to override default US region

delphaber:
I left US as default region. This could be useful for european suspender's users :)

Related Issue:
https://github.com/thoughtbot/suspenders/issues/427

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,10 @@ This:
 * Adds the [Rails 12 Factor](https://github.com/heroku/rails_12factor) gem
   to make running Rails 4 apps easier on Heroku
 
+If you want to specify a different region from US:
+
+    suspenders app --heroku true --heroku-region eu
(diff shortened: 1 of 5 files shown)


croaky:
@delphaber I like the idea. Thanks for opening it.

With a Stockholm office, this may be particularly useful for thoughtbot, too.

I'd prefer not to expand the `suspenders` API too much, though. To protect against wanting to add any other `heroku create` flags to the `suspenders` command, I edited this PR to take a `--heroku-flags` flag, instead. That adds the ability to change region, stack, addons, etc. without adding too many more flags to `suspenders`. What do you think?

https://github.com/thoughtbot/suspenders/pull/468

delphaber:
@croaky Yeah I think yours is a better approach :) Thank you!

PR #468, opened 2015-01-04
Allow additional Heroku flags

croaky:
Most useful are likely:

* `--addons`
* `--region`


mike-burns:
Good idea. OK to merge.

Issue #461, opened 2014-12-27
Use Host ENV in environment files


ACPK:
config.action_mailer.default_url_options = { host: ENV["HOST"] }

croaky:
I think `staging.myapp.com` is usually what we want but I take the 12factor point to [store config in the environment](http://12factor.net/config).

https://github.com/thoughtbot/suspenders/pull/469

Issue #407, opened 2014-09-25
Use `ENV["SMTP_DOMAIN"]` for `config.action_mailer.default_url_options`


jessieay:
Any reason not to use `ENV["SMTP_DOMAIN"]` here:
https://github.com/thoughtbot/suspenders/blob/29c9e26e717a1430696fc6b87b7b4ea1c2a1a5d4/lib/suspenders/actions.rb#L17 ?

We already required that variable to be set for https://github.com/thoughtbot/suspenders/blob/29c9e26e717a1430696fc6b87b7b4ea1c2a1a5d4/templates/smtp.rb#L4 so would be nice to re-use.

croaky:
`config.action_mailer.default_url_options` is used in our Mailer templates when we use `_url` named routes. Example: `upgrade_url` within Upcase might expand to `https://upcase.com/upgrade`. I think we want to keep these variables.

I think the SendGrid and Mandrill docs are showing `yourdomain.com` because they don't want their examples to be hosting-platform-specific. In Heroku's docs for [Mandrill](https://devcenter.heroku.com/articles/mandrill) and [SendGrid](https://devcenter.heroku.com/articles/sendgrid#ruby-rails), they suggest using `heroku.com`.

What if we just improve the comment?

https://github.com/thoughtbot/suspenders/pull/467

PR #469, opened 2015-01-05
Set default_url_options to ENV["HOST"] on servers

croaky:
https://github.com/thoughtbot/suspenders/issues/407
https://github.com/thoughtbot/suspenders/issues/461


mike-burns:
This should use `ENV.fetch("HOST")` so that we get runtime errors if it's unset.

croaky:
Agreed. Thanks.

PR #470, opened 2015-01-07
Create Hound-compliant config/unicorn.rb

croaky:
Fix:

> Line is too long. [91/80]

Fix:

> Unused block argument - server.
> Unused block argument - worker.
> You can omit all the arguments if you don't care about them.

### DIFF
--- a/templates/unicorn.rb
+++ b/templates/unicorn.rb
@@ -4,9 +4,9 @@ worker_processes (ENV["UNICORN_WORKERS"] || 3).to_i
 timeout (ENV["UNICORN_TIMEOUT"] || 15).to_i
 preload_app true
 
-before_fork do |server, worker|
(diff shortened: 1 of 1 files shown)


templates/unicorn.rb
```diff
@@ -4,9 +4,9 @@
 timeout (ENV["UNICORN_TIMEOUT"] || 15).to_i
 preload_app true
 
-before_fork do |server, worker|
+before_fork do |_, _|
```
gylaz: Thoughts on going with `_server`, `_worker` instead? Might be nice in the future to know what these variables were.

gylaz:
One question, :+1:

templates/unicorn.rb
croaky: Good idea.

PR #473, opened 2015-01-07
Use Travis container infrastructure

derekprior:
Builds start sooner and run faster. Open source builds on the container
infrastructure also get to use the bundle cache. See:
http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

This has worked extremely well on all projects I've tried it on thus far.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,4 @@ install:
   - bundle install
 notifications:
   email: false
+sudo: false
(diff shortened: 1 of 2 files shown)


croaky:
Awesome, thanks. Can the `cache: bundler` be removed? Not sure if the container itself acts as a caching mechanism.

derekprior:
No, those are independent. For instance, I enabled container infrastructure on this repo but did not enable bundle caching because this is a gem and that seems akin to checking in Gemfile.lock.

derekprior:
Looks like it took about 45% off the test runs for suspenders.

croaky:
Looks good.

PR #475, opened 2015-01-10
gitignore /public/assets

gabebw:


croaky:
Could you add a description to the git message? I'm guessing we don't want this because we only compile assets into `public/assets` at deploy-time? Generated files, not source files, that kind of thing?

gabebw:
[Derek Prior](https://twitter.com/derekprior/status/535172866359439361) strikes again.

Yes, good idea.

jferris:
This looks good to me. I can't see why we'd ever want to commit these.

Issue #471, opened 2015-01-07
requiring Rails.root.join("config/smtp") in production.rb fails in Ruby 2.2.0 and Rails 4.2.0


gabebw:
We should add `.to_s` to the end of it, because of this error:

> `require': no implicit conversion of Pathname into String (TypeError)

It seems to work in plain IRB, so I think it's a Rails problem.

jferris:
Can't we just use `require_relative`?

delphaber:
It works for me.

The `key not found` error is raised because I didn't set the ENV variable, but smtp.rb file was loaded

````
↳ $ RAILS_ENV=production rails s
=> Booting WEBrick
=> Rails 4.2.0 application starting in production on http://localhost:3000
=> Run `rails server -h` for more startup options
=> Ctrl-C to shutdown server
Exiting
/home/faber/weLaika/test-suspenders/config/smtp.rb:2:in `fetch': key not found: "SMTP_ADDRESS" (KeyError)
	from /home/faber/weLaika/test-suspenders/config/smtp.rb:2:in `<top (required)>'
````

It works also with

````ruby
require_relative "../smtp"
````

**Comment by croaky** [comment]
`require_relative` sounds good to me. Ready for review: https://github.com/thoughtbot/suspenders/pull/477

**Comment by gabebw** [comment]
I was actually wrong about the issue! I removed the `config/smtp` file, so `require` was failing:

```
> require Rails.root.join("nonexistent")                                                                                                                           
TypeError: no implicit conversion of Pathname into String
```

Sorry!

yangsec888:
require Rails.root.join("nonexistent").to_s

PR #477, opened 2015-01-12
Fix implicit conversion of Pathname into String

croaky:
https://github.com/thoughtbot/suspenders/issues/471


gabebw:
I was actually wrong about the issue in #471! I hadn't removed the `config/smtp` file, so `require` was failing:

```
> require Rails.root.join("nonexistent")                                                                                                                           
TypeError: no implicit conversion of Pathname into String
```

This can be closed. Sorry!

croaky:
No worries. Thanks for the follow-up.

PR #481, opened 2015-01-13
Linting factories

delphaber:
I think linting factories could be useful, as described [here].(http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md#Linting_Factories)

### DIFF
--- a/templates/factory_girl_rspec.rb
+++ b/templates/factory_girl_rspec.rb
@@ -1,3 +1,12 @@
 RSpec.configure do |config|
   config.include FactoryGirl::Syntax::Methods
+
+  config.before(:suite) do
+    begin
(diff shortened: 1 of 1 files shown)


croaky:
Thanks for the thought, @delphaber. We used to have this in Suspenders, and removed it in v1.13.0 in May of last year due to it slowing down the suite an average of ~300ms:

https://github.com/thoughtbot/suspenders/commit/0dbbc626e5da9f6ee94e1226487b4017cd2b8099

As far as I know, there's nothing that's since been added to Rails or Factory Girl or RSpec that would improve that slowdown.

delphaber:
Oh! I see :) No problem!

PR #483, opened 2015-01-14
Do run `rake dev:prime` on CI

gabebw:
We wrap the `dev:prime` task in `if Rails.env.development?`. CI runs in the test environment. Because the environments don't match, the `dev:prime` task is never defined in CI, and CI crashes.

### DIFF
--- a/templates/development_seeds.rb
+++ b/templates/development_seeds.rb
@@ -1,4 +1,4 @@
-if Rails.env.development?
+if Rails.env.development? || Rails.env.test?
   require "factory_girl"
 
   namespace :dev do


derekprior:
This is how I originally did it in the PR where I made travis use bin/setup but there was a feeling that if we really want bin/setup to be executable documentation, it should run prime. Should we just define it in test?

gabebw:
Usually `dev:prime` adds data that's convenient for developers (like an admin, plan levels, etc), but it shouldn't be necessary for tests. Tests should create their own data.

I don't think `dev:prime` should be necessary for tests to run.

derekprior:
I agree. It's not necessary for tests. Except that this also serves as a
test of bin/setup itself.

I don't feel strongly either way. I think it was @jferris That suggested
it. I'm inclined to agree.

jferris:
> I don't think dev:prime should be necessary for tests to run.

The idea of running it on CI isn't that it's necessary for tests, but that the build should fail is `dev:prime` doesn't work. We frequently forget to update `dev:prime` when updating the schema or factories, and running it on CI will catch that before merging.

gabebw:
OK, so then we should change how we define `rake dev:prime` so it's available in the test environment.

drapergeek:
One of the issues we face with dev seeds is making sure they're constantly useful and up to date. This doesn't just mean that they match the schema but also that we put in any information that people on the team need to do daily work.

One of the ways I've managed to keep them up to date is having the dev:prime task run at the end of the default rake task. This way each time you run your full suite, any extra information you've added to your dev environment is gone which forces you to start adding it to the dev prime task instead. This benefits the entire team by having good seed data to work with, specifically on new features you're currently adding. This is especially useful when you're building a feature in TDD and then handing it off to a designer. Once you've added the seeds you need to verify it in the browser, they're ready to go for the designer and they only have to run `rake` to make sure they're ready.

If we moved the dev:prime task to the end of the rake, then you would get the same benefit of testing it on CI but also improve the usefulness of the seeds in the future (IMO).

The only downside is that you'd need to run `rake` after the `bin/setup` when you first join a project. As a developer I know that is my first move but it may not be for designers.

croaky:
I think my preference would be to alter the existing `rake dev:prime` task to run in both `development` and `test` environments for the reason @jferris mentioned above.

The workflow @drapergeek describes is interesting to hear and I can see the appeal but feels like extra coupling that might be unexpected for folks. In the case where a teammate should re-run `bin/setup` after they pull master, it seems reasonable to me to mention that in the team chat room:

> "Merged abc123"
> @croaky ^ hey folks, you'll want to run `./bin/setup` after pulling that one.

Most commits and most `rake` runs don't need to re-prime the dev database but we still want to catch potential errors in it. I like the idea of letting CI handle that for us.

gabebw:
OK, changed.

croaky:
Looks good. Let's give it a whirl.

PR #484, opened 2015-01-14
Don't set default_url_options to a string

gabebw:
Before:

```ruby
config.action_mailer.default_url_options = { host: 'ENV.fetch("HOST")' }
```

After:

```ruby
config.action_mailer.default_url_options = { host: ENV.fetch("HOST") }
```

### DIFF
--- a/lib/suspenders/actions.rb
+++ b/lib/suspenders/actions.rb
@@ -10,8 +10,8 @@ module Suspenders
     end
 
     def action_mailer_host(rails_env, host)
-      host_config = "config.action_mailer.default_url_options = { host: '#{host}' }"
-      configure_environment(rails_env, host_config)
+      config = "config.action_mailer.default_url_options = { host: #{host} }"
+      configure_environment(rails_env, config)
     end
 
     def configure_environment(rails_env, config)
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -237,8 +237,8 @@ end
     end
 
     def configure_action_mailer
-      action_mailer_host "development", "localhost:#{port}"
-      action_mailer_host "test", "www.example.com"
+      action_mailer_host "development", %{"localhost:#{port}"}
+      action_mailer_host "test", %{"www.example.com"}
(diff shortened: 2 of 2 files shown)


lib/suspenders/app_builder.rb
```diff
@@ -235,7 +235,7 @@ def configure_simple_form
     def configure_action_mailer
       action_mailer_host "development", "localhost:#{port}"
       action_mailer_host "test", "www.example.com"
-      action_mailer_host "staging", %{ENV.fetch("HOST")}
+      action_mailer_host "staging", %q{ENV.fetch("HOST")}
```
gabebw: Oops!

lib/suspenders/actions.rb
```diff
@@ -10,7 +10,7 @@ def replace_in_file(relative_path, find, replace)
     end
 
     def action_mailer_host(rails_env, host)
-      host_config = "config.action_mailer.default_url_options = { host: '#{host}' }"
+      host_config = "config.action_mailer.default_url_options = { host: #{host} }"
```
gabebw: I think we can leave this as-is?

gabebw:
How can I test that this works?

lib/suspenders/actions.rb
croaky: Could name the variable `config`.

croaky:
Whoops, good call.

croaky:
I think the test failure is because in the test environment, we're now passing in `www.example.com`.

What if we passed in the whole config from `lib/suspenders/app_builder.rb`?

```shell
lib/suspenders/app_builder.rb
236:      action_mailer_host "development", %{{ host: "localhost:#{port}" }'}
237:      action_mailer_host "test", '{ host: "www.example.com" }'
238:      action_mailer_host "staging", '{ host: ENV.fetch("HOST") }'
239:      action_mailer_host "production", '{ host: ENV.fetch("HOST") }'
```

I think the spec could be similar to `spec/features/new_project_spec.rb`'s "generated en.yml is evaluated" spec.

croaky:
Good to squash and merge.

PR #485, opened 2015-01-14
Correctly indent config.static_cache_control line

gabebw:
Before:

config.static_cache_control = "public, max-age=#{1.year.to_i}"

After:

 config.static_cache_control = "public, max-age=#{1.year.to_i}"

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -95,7 +95,7 @@ module Suspenders
 
       inject_into_file(
         "config/environments/production.rb",
-        'config.static_cache_control = "public, max-age=#{1.year.to_i}"',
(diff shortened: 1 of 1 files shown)


croaky:
Good to merge.

gabebw:
@croaky How do I test that this actually does what I'm 85% sure it will do? Do I have to do `rake install` and generate a new Rails app? Can I write a test?

croaky:
I don't think this is worth a test for what it does and for the time needed in the setup phase of the feature specs.

gabebw:
Makes sense.

croaky:
Your time is too valuable to the world to write a spec for an instance of two-space indentation, Mr. Gabe Berke-Williams. Helping give that time back to humanity and its flora and fauna is my greatest achievement today.

PR #486, opened 2015-01-14
Prepend a new line before adding to README

gabebw:
Before:

```
* [Style](http://github.com/thoughtbot/guides/blob/master/style)
## Deploying
```
...and it renders "## Deploying" on the same line as the Style link.

After:

```
* [Style](http://github.com/thoughtbot/guides/blob/master/style)

## Deploying
```

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -340,6 +340,7 @@ fi
       copy_file "bin_deploy", "bin/deploy"
 
       instructions = <<-MARKDOWN
+
 ## Deploying
 
 If you have previously run the `./bin/setup` script,


gabebw:
This SHOULD work but I'm not sure how to test it.

teoljungberg:
I would prefer adding the newline at the end of the `[Style]` block instead, though it might be less obvious considering it's a template

gabebw:
Do you mean that the default `README.md` would have an extra newline in it? I'd prefer not to do that because, well, then  there's an extra newline in everyone's README. This way, we only add newlines when we're appending actual content.

teoljungberg:
Yeah, it's definitely more clear having a newline before the deploy section than an extra in the README

croaky:
Ready to merge.

PR #487, opened 2015-01-14
Remove unused require of production.rb

teoljungberg:


jferris:
We require `production.rb` to ensure staging has parity with production. Our default configuration uses production exactly and then prevents us from accidentally spamming real users from staging.

Can you explain why this isn't necessary?

teoljungberg:
My brain kept thinking that `production.rb` was a file in the repository that had been removed, a quick `git log` would have proved me wrong. My bad, sorry Joe!

PR #490, opened 2015-01-15
Turn off parameter wrapping

gabebw:
With this setting on, Rails will happily turn a POST to the UsersController 
from this:

    { name: "Ralph" }

Into this:

    { users: { name: "Ralph" }, name: "Ralph" }

This means that `params.require(:user).permit(:name)` will not raise an error,
even if the developer forgets to nest their parameters inside a top-level
`user` key.

It is enabled for JSON by default. This change turns it off for every format.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -118,6 +118,10 @@ end
       template 'secrets.yml', 'config/secrets.yml', force: true
     end
 
+    def disallow_wrapping_parameters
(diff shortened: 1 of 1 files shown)


gabebw:
We can also just remove the file - is that too much?

jferris:
If removing the file works, I think I prefer that. We generate less code, and the code for removing it in Suspenders will be simpler.

croaky:
Ready to squash and merge.

PR #494, opened 2015-01-20
Tell Heroku to serve static files

gabebw:
By default, Rails 4 will not serve your assets:
https://devcenter.heroku.com/articles/rails-4-asset-pipeline#serve-assets

We use an environment variable to tell Rails to serve static files:

```ruby
```

..but we were not automatically setting that environment variable on Heroku.
Now we are.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -300,9 +300,13 @@ end
     end
 
     def create_heroku_apps(flags)
+      rack_env = "RACK_ENV=staging RAILS_ENV=staging"
+      rails_serve_static_files = "RAILS_SERVE_STATIC_FILES=true"
+      staging_config = "#{rack_env} #{rails_serve_static_files}"
       run_heroku "create #{app_name}-production #{flags}", "production"
       run_heroku "create #{app_name}-staging #{flags}", "staging"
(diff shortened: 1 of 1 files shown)


lib/suspenders/app_builder.rb
```diff
@@ -302,7 +302,8 @@ def init_git
     def create_heroku_apps(flags)
       run_heroku "create #{app_name}-production #{flags}", "production"
       run_heroku "create #{app_name}-staging #{flags}", "staging"
-      run_heroku "config:add RACK_ENV=staging RAILS_ENV=staging", "staging"
+      run_heroku "config:add RACK_ENV=staging RAILS_ENV=staging config:add RAILS_SERVE_STATIC_FILES=true", "staging"
```
gabebw: arf

lib/suspenders/app_builder.rb
```diff
@@ -300,9 +300,12 @@ def init_git
     end
 
     def create_heroku_apps(flags)
+      staging_environment = "RACK_ENV=staging RAILS_ENV=staging"
```
croaky: How about naming the variable after the ENV variable like `serve_static_files`? `rack_env = = "RACK_ENV=staging RAILS_ENV=staging"`

croaky:
Looks good to merge pending CI.

gylaz:
:+1:

PR #495, opened 2015-01-21
Document how to use the #title helper

mike-burns:
When opening layout.html.erb to change the title, if can be off-putting
to traverse down a nest of discovery in order to know that one should
change the title in config/locales/*.yml. Add a note right there so that
people don't have to hunt.

### DIFF
--- a/templates/suspenders_layout.html.erb.erb
+++ b/templates/suspenders_layout.html.erb.erb
@@ -4,6 +4,11 @@
   <meta charset="utf-8" />
   <meta name="ROBOTS" content="NOODP" />
   <meta name="viewport" content="initial-scale=1" />
+  <%%#
+    Configure default and controller-, and view-specific titles in
+    config/locales/en.yml. For more see:
+    https://github.com/calebthompson/title#usage
+  %>
   <title><%%= title %></title>
   <%%= stylesheet_link_tag :application, media: "all" %>
   <%%= csrf_meta_tags %>


templates/suspenders_layout.html.erb.erb
```diff
@@ -4,6 +4,7 @@
   <meta charset="utf-8" />
   <meta name="ROBOTS" content="NOODP" />
   <meta name="viewport" content="initial-scale=1" />
+  <!-- change the title in config/locales/en.yml in the en.titles key -->
```
croaky: We probably don't want this in production. ERb comment?

Or, not use the title gem at all if it's confusing to use?

gabebw: > not use the title gem at all if it's confusing to use

For what it's worth, I have never customized the title. It sets the title to "Application Name", which is fine, but anything beyond that requires research and so I don't do it.

I believe we should remove the title gem.

drapergeek: :-1: I like this gem. It works great for setting custom titles depending on your current page.

I don't think that it is fair to say it is confusing, the usage is just a bit hidden and pointing it out would be a sufficient way to fix the problem.

derekprior: I like clarifying the use but am against removing. We *should* be setting
titles per page for a better user experience and for better (and
simple) SEO. It's bad UX to open several tabs to an app and have them all
say "Application Name". Title makes it really simple to fix this once you
know how to use it.

mike-burns:
Updated to use an Erb comment.

calebthompson:
Cool, that being there can we drop the `title.application` key, which is defaulted to the same behavior?

https://github.com/calebthompson/title/blob/master/app/helpers/title/title_helper.rb#L31

templates/suspenders_layout.html.erb.erb
```diff
@@ -4,6 +4,7 @@
   <meta charset="utf-8" />
   <meta name="ROBOTS" content="NOODP" />
   <meta name="viewport" content="initial-scale=1" />
+  <%%# change the title in config/locales/en.yml in the en.titles key %>
```
calebthompson: `en.titles.{controller}.{view}`

gabebw: How about capitalizing? Maybe:

> Change the title in config/locales/en.yml using the \`en.titles.application\` key.

croaky:
I'm good with this being merged whenever @calebthompson gives his thumbs up.

templates/suspenders_layout.html.erb.erb
calebthompson: Better yet,

`<%%# Configure default and controller/view specific titles in config/locales/en.yml. https://github.com/calebthompson/title#usage $>`

calebthompson: Happy to accept any improvements to that documentation to clarify this.

gabebw: Oh, yeah, link to the docs.

mike-burns:
Thanks for the wording help. How's it look now?

calebthompson:
Looks good.

PR #497, opened 2015-01-22
Faster, more correct bundler gem and Gemfile install

bf4:
- About 3x faster when gems are available
- Does not return false positive when other gems on system have 'bundler' in name

**Updated**: [per discussion](https://github.com/thoughtbot/suspenders/pull/497#discussion-diff-23411299), to use `gem install bundler --conservative` as it works with rbenv and will only install the gem if not installed.
Tested on my 2015 macbook pro, Mavericks, with RVM and Ruby 2.1.5 running via bash

Regarding the `gem uninstall -aIx gem_name` command used below

```plain
 -a  Uninstall all matching versions
 -I  Ignore dependency requirements while uninstalling.
 -x  Uninstall applicable executables without confirmation.
```
###  gem list bundler --installed, _0.619_*

```bash
gem uninstall -aIx bundler
time gem list bundler --installed; echo $?
```

```plain
true

real    0m0.619s
user    0m0.538s
sys     0m0.077s
0
```

*but note it is falsely returning 'true' because I have other gems on the system with 'bundler' in the name*

### gem list ^bundler$ --installed, _0.604_

```bash
time gem list ^bundler$ --installed; echo $?
```

```plain
false

real    0m0.604s
user    0m0.525s
sys     0m0.075s
1
```

### command -v bundle, _0.000_

```bash
time command -v bundle; echo $?
```

```plain
real    0m0.000s
user    0m0.000s
sys     0m0.000s
1
```

### gem list bundler --installed || gem install bundler, _0.621_*

```bash
gem uninstall -aIx bundler
time eval "gem list bundler --installed || gem install bundler"; echo $?
```

```plain
true

real    0m0.621s
user    0m0.539s
sys     0m0.076s
0
```

*again, it is falsely only executing the LHS because I have other gems on the system with 'bundler' in the name*

### gem list ^bundler$ --installed || gem install bundler, _4.695_

```bash
gem uninstall -aIx bundler
time eval "gem list ^bundler$ --installed || gem install bundler"; echo $?
```

```plain
false
Fetching: bundler-1.7.12.gem (100%)
Successfully installed bundler-1.7.12
Parsing documentation for bundler-1.7.12
Installing ri documentation for bundler-1.7.12
Done installing documentation for bundler after 1 seconds
1 gem installed

real    0m4.695s
user    0m2.983s
sys     0m0.348s
0
```

### command -v bundle || gem install bundler, _4.051_

```bash
gem uninstall -aIx bundler
time eval "command -v bundle || gem install bundler"; echo $?
```

```plain
Fetching: bundler-1.7.12.gem (100%)
Successfully installed bundler-1.7.12
Parsing documentation for bundler-1.7.12
Installing ri documentation for bundler-1.7.12
Done installing documentation for bundler after 1 seconds
1 gem installed

real    0m4.051s
user    0m2.459s
sys     0m0.269s
0
```

##### bundle, _2.706_

```bash
rm Gemfile.lock; gem uninstall -aIx dotenv-rails dotenv
time bundle; echo $?
```

```plain
Fetching gem metadata from https://rubygems.org/....
Resolving dependencies...
Installing dotenv 1.0.2
Installing dotenv-rails 1.0.2
Using bundler 1.7.12
Your bundle is complete!
Use `bundle show [gemname]` to see where a bundled gem is installed.

real    0m2.706s
user    0m0.739s
sys     0m0.103s
0
```

##### bundle check || bundle install, _2.423_

```bash
rm Gemfile.lock; gem uninstall -aIx dotenv-rails dotenv
time eval "bundle check || bundle install"; echo $?
```

```plain
Resolving dependencies...
Bundler can't satisfy your Gemfile's dependencies.
Install missing gems with `bundle install`.
Fetching gem metadata from https://rubygems.org/....
Resolving dependencies...
Installing dotenv 1.0.2
Installing dotenv-rails 1.0.2
Using bundler 1.7.12
Your bundle is complete!
Use `bundle show [gemname]` to see where a bundled gem is installed.

real    0m2.423s
user    0m1.264s
sys     0m0.171s
0
```

##### bundle, _3.176_

```bash
bundle; gem uninstall -aIx dotenv-rails dotenv
time bundle; echo $?
```

```plain
Fetching gem metadata from https://rubygems.org/...
Installing dotenv 1.0.2
Installing dotenv-rails 1.0.2
Using bundler 1.7.12
Your bundle is complete!
Use `bundle show [gemname]` to see where a bundled gem is installed.

real    0m3.176s
user    0m0.755s
sys     0m0.105s
0
```

##### bundle check || bundle install, _2.880_

```bash
bundle; gem uninstall -aIx dotenv-rails dotenv
time eval "bundle check || bundle install"; echo $?
```

```plain
The following gems are missing
 * dotenv (1.0.2)
  * dotenv-rails (1.0.2)
  Install missing gems with `bundle install`
  Fetching gem metadata from https://rubygems.org/...
  Installing dotenv 1.0.2
  Installing dotenv-rails 1.0.2
  Using bundler 1.7.12
  Your bundle is complete!
  Use `bundle show [gemname]` to see where a bundled gem is installed.

  real    0m2.880s
  user    0m1.277s
  sys     0m0.178s
  0
  ```

### Gems Installed

As the time to resolve and install gems increases, `bundle check || bundle`'s speed
advantage becomes more impressive. As a bonus, `bundle check` will update the `Gemfile.lock`
per changes in the Gemfile and all dependent gem versions are available locally.

##### bundle, _1.519_

```bash
bundle; rm Gemfile.lock
time bundle; echo $?
```

```plain
Fetching gem metadata from https://rubygems.org/....
Resolving dependencies...
Using dotenv 1.0.2
Using dotenv-rails 1.0.2
Using bundler 1.7.12
Your bundle is complete!
Use `bundle show [gemname]` to see where a bundled gem is installed.

real    0m1.519s
user    0m0.651s
sys     0m0.092s
0
```

##### bundle check || bundle install, _0.612_

```bash
bundle; rm Gemfile.lock
time eval "bundle check || bundle install"; echo $?
```

```plain
Resolving dependencies...
The Gemfile's dependencies are satisfied

real    0m0.612s
user    0m0.535s
sys     0m0.073s
0
```

#### bundle, _0.669_

``bash
bundle
time bundle; echo $?
```

```plain
Using dotenv 1.0.2
Using dotenv-rails 1.0.2
Using bundler 1.7.12
Your bundle is complete!
Use `bundle show [gemname]` to see where a bundled gem is installed.

real    0m0.669s
user    0m0.582s
sys     0m0.086s
0
```

#### bundle check || bundle install, _0.596_

```bash
bundle
time eval "bundle check || bundle install"; echo $?
```

```plain
The Gemfile's dependencies are satisfied

real    0m0.596s
user    0m0.521s
sys     0m0.071s
0
```

### DIFF
--- a/templates/bin_setup.erb
+++ b/templates/bin_setup.erb
@@ -7,8 +7,8 @@
 set -e
 
 # Set up Ruby dependencies via Bundler
-gem list bundler --installed > /dev/null || gem install bundler
-bundle install
+gem install bundler --conservative
+bundle check || bundle install
 
 # Set up configurable environment variables
 if [ ! -f .env ]; then


templates/bin_setup.erb
```diff
@@ -7,8 +7,8 @@
 set -e
 
 # Set up Ruby dependencies via Bundler
-gem list bundler --installed > /dev/null || gem install bundler
-bundle install
+command -v bundle > /dev/null || gem install bundler
```
derekprior: `command -v` is insufficient if you are an rbenv user. It will pass if you have rbenv installed in any ruby, as the shim for bundler will be in the path.

derekprior:
`bundle check` seems inline with the idempotentcy changes we've made recently, but `gem list` is still more likely to be correct for everyone. I'm not really sweating half a second in `bin/setup`.

jferris:
My local numbers are much faster for each case, and I agree with @derekprior that it's more important to be correct than half a second faster in this case.

One other option we haven't explored is the `--conservative` option:

`gem install bundler --conservative`

templates/bin_setup.erb
bf4: Ah, good to know.  I cargo-culted it from lower down in the script.  

I can change the PR to then 

1. add `bundle check`
2. apply the regex fix to `gem list`

@jferris suggested `gem install bundler --conservative` but I think that is unrelated to this PR, which was originally just about `bundle check`'ing "all the things" :)

jferris: I think it is related.

The `gem list bundler` line being changed here is to make sure Bundler gets installed. It looks like RubyGems has that functionality built-in (install a gem, only if it's missing), and the flag for it is `--conservative`.

croaky: We use `gem install --conservative bundler` in Paperback's `bin/setup`
script. Maybe @trevororeilly has thoughts on it.

bf4: Ah, I accidentally the gem docs.

bf4:
Updated per discussion.

trevororeilly:
I first noticed the `--conservative` flag in the default Rails `bin/setup` script:

https://github.com/rails/rails/blob/v4.2.0/railties/lib/rails/generators/rails/app/templates/bin/setup#L11

Looks like the same approach as what's proposed here.

jferris:
The current version looks pretty good to me. @croaky any thoughts?

croaky:
Thanks, @bf4. Merged as e5d5e43.

bf4:
Turn out, `bundle install --local` might be a better option ref: https://blog.engineyard.com/2011/bundler-pro-tip

tute:
I'm surprised `--local` is not default behavior.

jferris:
I think that, when generating a brand new Rails project, we probably want to fetch the gems index so that we install with the latest gems.

tute:
> I think that, when generating a brand new Rails project, we probably want to fetch the gems index so that we install with the latest gems.

Indeed, I thought we were talking about suspenders' CI.

jferris:
Hmm, yeah, it looks like generating a Suspenders app [does its own `bundle install`](https://github.com/thoughtbot/suspenders/blob/master/lib/suspenders/generators/app_generator.rb#L63).

Using `--local` will fail if the gems aren't available, though, so unless you're vendoring and committing gems, that flag will cause the `bundle install` to fail for other developers.

PR #498, opened 2015-01-23
Use new --skip-turbolinks flag

delphaber:
Rails 4.2 introduced -skip-turbolinks flag. We can remove some code and use that flag instead. What do you think?

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -140,12 +140,6 @@ end
         force: true
     end
 
-    def remove_turbolinks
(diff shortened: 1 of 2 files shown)


gabebw:
Well I love this idea.

JoelQ:
:+1:

teoljungberg:
I'm in favour of deleting code, :+1:

jferris:
Looks good to me.

croaky:
Thanks, @delphaber. Merged as 500f1a6.

PR #501, opened 2015-01-26
Replace Unicorn with Puma

calebthompson:
> Heroku now recommends using the Puma webserver. The previously
> recommended webserver, Unicorn, is susceptible to slow client attacks.

https://devcenter.heroku.com/changelog-items/594


templates/puma.rb
```diff
@@ -0,0 +1,20 @@
+# https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server
+
+# The environment variable WEB_CONCURRENCY may be set to a default value based
+# on dyno size. To manually configure this value use heroku config:set
+# WEB_CONCURRENCY.
+workers Integer(ENV["WEB_CONCURRENCY"] || 3)
+threads_count = Integer(ENV["MAX_THREADS"] || 5)
```
derekprior: Where did we get these settings from? Reccomended by heroku? Default?

calebthompson: https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#config

derekprior: :+1:

JoelQ:
:+1:

calebthompson:
Okay before merging I'm going to try actually running an app with this server and these config.

calebthompson:
And leave this at least a day.

templates/Procfile
```diff
@@ -1,2 +1,2 @@
-web: bundle exec unicorn -p $PORT -c ./config/unicorn.rb
+web: bundle exec rails serve Puma -p $PORT -c ./config/puma.rb
```
croaky: Why `rails serve Puma` rather than `bundle exec puma` per https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server?

calebthompson: I pulled it from https://github.com/puma/puma#rails.

templates/unicorn.rb
```diff
@@ -1,30 +0,0 @@
...
-after_fork do |_server, _worker|
-  Signal.trap "TERM" do
-    puts "Unicorn worker intercepting TERM, waiting for master to send QUIT"
-  end
-
-  if defined? ActiveRecord::Base
-    config = ActiveRecord::Base.configurations[Rails.env] ||
-      Rails.application.config.database_configuration[Rails.env]
-    config["reaping_frequency"] = (ENV["DB_REAPING_FREQUENCY"] || 10).to_i
-    config["pool"] = (ENV["DB_POOL"] || 2).to_i
```
croaky: It looks like there's a regression here, losing our `DB_POOL` setting. It looks like with Heroku and Puma, they suggest setting it in the `database.yml` file:

https://devcenter.heroku.com/articles/concurrency-and-database-connections#threaded-servers

templates/unicorn.rb
```diff
@@ -1,30 +0,0 @@
-# https://devcenter.heroku.com/articles/rails-unicorn
-
-worker_processes (ENV["UNICORN_WORKERS"] || 3).to_i
-timeout (ENV["UNICORN_TIMEOUT"] || 15).to_i
```
croaky: With `UNICORN_TIMEOUT` gone and no timeout built into Puma, should we increase our default `RACK_TIMEOUT` to 20?

https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#timeout

croaky:
Definitely want to hear reports of people using this config in production before we merge this in.

Otherwise, makes sense and thanks for the PR. Good to squash and merge at your discretion.

templates/Procfile
croaky: Maybe some details here: http://stackoverflow.com/a/21218048 That answer
makes sense to me but I wonder why Heroku's docs are different, if there's
anything platform-specific that makes the Heroku docs version better on
Heroku. Maybe @schneems or @hone knows?

schneems: I don't think there are differences really. There were some conflicts with some flags being parsed and passed to `rails`. The main reason we're doing it that way is so anyone running straight Rack can copy most of the article verbatim and still have it work. Same command on every project. 

Btw, i think `serve` should be `server`. Also have you considered providing default config values? So you can run without a port specified i.e. `-p ${PORT:-3000}`

templates/unicorn.rb
```diff
@@ -1,30 +0,0 @@
-# https://devcenter.heroku.com/articles/rails-unicorn
-
-worker_processes (ENV["UNICORN_WORKERS"] || 3).to_i
```
calebthompson: @croaky we lost something here too. Any idea what?

calebthompson: Found it.

calebthompson:
@croaky @schneems updated with most feedback addressed.

templates/unicorn.rb
calebthompson: Done in 211517e.

templates/unicorn.rb
calebthompson: Re-added in 76cf8a4. Much cleaner.

templates/postgresql_database.yml.erb
```diff
@@ -4,9 +4,14 @@ development: &default
...
-  pool: 2
+  pool: <%%= ENV.fetch("DB_POOL", 2) %>
+  reaping_frequency: <%%= ENV.fetch("DB_REAPING_FREQUENCY", 10) %>
   timeout: 5000
 
 test:
   <<: *default
   database: <%= app_name %>_test
+
+production:
```
croaky: Do we want this for staging and production? Not sure if there's a better way, but maybe something like:

```yaml
staging: &deploy
  <<: *default
  url:  <%%= ENV.fetch("DATABASE_URL") %>

production:
  <<: *deploy
```

templates/postgresql_database.yml.erb
```diff
@@ -4,9 +4,14 @@ development: &default
...
+  url:  <%%= ENV.fetch("DATABASE_URL") %>
+  pool: <%%= ENV.fetch("DB_POOL", ENV.fetch("MAX_THREADS", 5) %>
```
croaky: Do we want this `pool` setting in all environments? Worth moving to `default` / `development` and then using the default in the deploy environments? Looks like this `pool` setting is currently out of sync with the `development` version. Not sure if that could cause any unexpected bugs in development, but either way, seems like maintaining parity could be valuable.

templates/postgresql_database.yml.erb
calebthompson: I don't think so. We'd have to have that URL instead of the rest of this config.

This works because heroku provides this URL, Rails knows how to parse it, and we just override a single setting (or set it in the first place, in this case).

croaky: Actually, I wonder why we need to set `url` to `DATABASE_URL` at all. I thought Heroku already did that? Can we drop this line and only set `pool`?

templates/postgresql_database.yml.erb
calebthompson: I saw that and didn't change anything. I upped this from what we previously had, 2, and added the second ENV variable.

I also don't think we want production to use the defaults, in case they override something Heroku sets in production.

If anything, possibly this:

```
production: &deploy
  <<: *default # has the pool
  url: <%%= ENV.fetch("DATABASE_URL") %>

staging:
  <<: *deploy
```

@schneems ?

jferris:
This is great. I've been thinking we should be switching to Puma for a while, and I've always been uncomfortable with the slow client problem.

I think Upcase is a good mid-traffic app to try this on. I can hopefully try it out this week, but if you're looking for a test project, feel free to open a pull request there.

calebthompson:
Thanks, Joe.

templates/Procfile
```diff
@@ -1,2 +1,2 @@
-web: bundle exec unicorn -p $PORT -c ./config/unicorn.rb
+web: bundle exec rails server Puma -p ${PORT:-3000} -c ./config/puma.rb
```
pedrosmmoreira: This seems to break starting the app with Foreman. While running `bundle exec rails s puma' works, passing the config file will break with `undefined method `workers'`. This is a known issue (see https://github.com/puma/puma/issues/512)  and the example Procfile in the Heroku docs uses puma directly. Of course, the downside to that solution is losing logging...

Changing the Procfile to `web: bundle exec puma -C ./config/puma.rb` does boot the app loading the config.
but will also throw a thread warning related to the connection pool (discussion around similar issues can be found in https://github.com/puma/puma/issues/614).

templates/puma.rb
```diff
@@ -0,0 +1,20 @@
...
+# on dyno size. To manually configure this value use heroku config:set
+# WEB_CONCURRENCY.
+workers Integer(ENV.fetch("WEB_CONCURRENCY", 3))
+threads_count = Integer(ENV.fetch("MAX_THREADS", 5))
+threads(threads_count, threads_count)
+
+preload_app!
+
+rackup DefaultRackup
+port ENV.fetch("PORT", 3000)
```
pedrosmmoreira: Setting the port in the config seems to cause Puma to bind twice, throwing an `Address already in use - bind(2)` error.

templates/postgresql_database.yml.erb
schneems: It's a best practice. Rails does merge in DATABASE_URL into your config but explicit is better than implicit imho. As of 4.1 all this behavior is in Rails, Heroku isn't doing anything other than setting the env var

templates/postgresql_database.yml.erb
schneems: You need your pool to at least be as large as your puma thread count. Btw I recommend against using a "staging" specific environment: https://devcenter.heroku.com/articles/deploying-to-a-custom-rails-environment

templates/puma.rb
schneems: I've heard of this as an issue but never run into it myself, can you give me a repro? Does it happen every time?

pedrosmmoreira: @schneems It happens pretty consistently for me. I've pushed an app generate with suspenders and add puma and the config. Setting the port throws the error, link: https://github.com/pedrosmmoreira/puma_double_binding

@croaky, @calebthompson If you feel this should discussed further as an issue in the Puma repo and not here, please let me know.

templates/postgresql_database.yml.erb
calebthompson: Looks like you might be right:

> database.yml present
> DATABASE_URL present
> => Merged into `url` sub key. If both specify `url` sub key, the `database.yml` `url` 
>     sub key "wins". If other paramaters `adapter` or `database` are specified in YAML, 
>     they are discarded as the `url` sub key "wins".

https://github.com/rails/rails/pull/13463#issuecomment-31480799 linked from https://github.com/rails/rails/pull/13578

Which would mean that we can't use defaults in production if it specifies database names, etc. but can set things like pool to be merged in.

These are the keys we already have which I think we can set safely, without overriding DATABASE_URL:

```yaml
  encoding: utf8
  min_messages: warning
  pool: 2
  timeout: 5000
```

templates/puma.rb
schneems: I wonder if this double binding port issue is because you're specifying `-p` in the boot command and then again in the puma.rb config file? If you take out the `-p` flag does it work? If so we should open an issue with Puma.

templates/postgresql_database.yml.erb
croaky: @schneems That is really interesting. We'll have to give that a try. I created a separate issue to track that: https://github.com/thoughtbot/suspenders/issues/506

For the purpose of this PR, I think it's more incremental to continue to have our separate staging environment. Could end up getting removed soon thereafter but that seems fine.

NEWS.md
```diff
@@ -1,3 +1,9 @@
+Unreleased
+
+* Remove Unicorn in favor of Puma, as [recommended by Heroku]
```
gabebw: Period?

templates/Procfile
gabebw: Why default to `3000`? That seems like an additional (unnecessary) change. Is it something to do with Puma? Doesn't foreman handle `$PORT` for us?

templates/postgresql_database.yml.erb
gabebw: Looks like `ENV.fetch("DB_POOL", ENV.fetch("MAX_THREADS", 5)` is missing a closing parenthesis.

templates/Procfile
croaky: Good catch. We should either re-use `def port` from lib/suspenders/app_builder.rb or rely on Foreman.

templates/puma.rb
jayroh: Same issue as @pedrosmmoreira here. 

Removing the `-p` in the Procfile will allow it to boot without any problems though.

templates/postgresql_database.yml.erb
```diff
@@ -4,9 +4,14 @@ development: &default
...
+  pool: <%%= ENV.fetch("DB_POOL", 2) %>
+  reaping_frequency: <%%= ENV.fetch("DB_REAPING_FREQUENCY", 10) %>
   timeout: 5000
 
 test:
   <<: *default
   database: <%= app_name %>_test
+
+production:
+  url:  <%%= ENV.fetch("DATABASE_URL") %>
```
jayroh: When running this in my development env it chokes because I don't have `DATABASE_URL` set. Suffice it to say - on heroku this *will* be set, and locally it's not needed. A change to `<%%= ENV.fetch("DATABASE_URL", "") %>` fixed this for me.

templates/puma.rb
jayroh: It looks like removing either `-p $PORT` from the Procfile **or** `port ENV.fetch('PORT', 3000)` from `puma.rb` will fix this particular hiccup.

gabebw: I vote we remove `port ENV.fetch("PORT", 3000)` from `puma.rb` then, I think it's easier to reason about `-p $PORt`

templates/postgresql_database.yml.erb
croaky: Hi @jayroh. <3 <3

jayroh: Hollaaaaaaaaaa @croaky!

templates/puma.rb
jayroh: I'm with @gabebw!

templates/postgresql_database.yml.erb
stevenharman: If `fetching` with an empty default, perhaps just `ENV["DATABASE_URL"]`?

templates/Procfile
stevenharman: > the example Procfile in the Heroku docs uses puma directly. Of course, the downside to that solution is losing logging...

@pedrosmmoreira Curious what you mean by "losing logging"? The Rails app itself will still log to whatever mechanism you've set the log to (`STDOUT` when running on Heroku). Are there other logs I'm not aware of or missing?

pedrosmmoreira: I might not have explained myself correctly. Using Puma directly runs the server unaware of the rails environment: this modifies logging behaviour, if I'm not mistaken.

bf4: The problem with `-c ./config/puma.rb` is that the rails server command is expecting a rackup file there, e.g. `config.ru` [by default](https://github.com/rails/rails/blob/4-2-stable/railties/lib/rails/commands/server.rb#L117).

To do what you want, what you basically need to do is:

```ruby
require ::File.expand_path('../config/environment', __FILE__)
require "puma"
require "puma/configuration"
puma_config_file = ::File.expand_path('../config/puma.rb', __FILE__)
Puma::Configuration.new(config_file: puma_config_file).load
run Rails.application
```

which doesn't work, so we'd need to ask the puma peeps how to configure puma from a rackup file

schneems: I don't think the rackup file is supposed to know about the server:

```
# This file is used by Rack-based servers to start the application.

require ::File.expand_path('../config/environment',  __FILE__)
run CodeTriage::Application
```

What's the problem with manually using the `puma` command in the `web:` line of the `Procfile`?

calebthompson: Works for me. Done.

calebthompson: 9c7c2f8

templates/puma.rb
calebthompson: Dropped both, we'll let the default port win.

templates/postgresql_database.yml.erb
calebthompson: Fixed in a78dc1e

gabebw:
Since we now never specify the port in either puma.rb or the Procfile, will Puma still work with Foreman?

schneems:
Foreman uses port 5000 by default. Also even if you start using `rails s` rails sees puma is in the gemfile and uses it by default, but won't use your config/puma.rb setup.

templates/postgresql_database.yml.erb
calebthompson: Done in ba12e09

templates/postgresql_database.yml.erb
calebthompson: Done in ba12e09

gabebw:
So, this will work with Foreman even if I use a custom port like `foreman start -p 7001`?

jayroh:
@gabebw - Yes. It will. Working for me here, currently.

gabebw:
Great!

templates/Procfile
```diff
@@ -1,2 +1,2 @@
-web: bundle exec unicorn -p $PORT -c ./config/unicorn.rb
+web: bundle exec puma -C ./config/puma.rb
```
jayroh: Without the port here, `-p $PORT`, or set in puma.rb things go wrong pretty quick up on Heroku. I would have guessed Heroku would have assigned it as a default somewhere but that doesn't look to be the case. I deployed without `-p $PORT` and with, and it would only work *with*.

gabebw: I vote `-p $PORT` then. Thanks for checking, Joel.

schneems: > . I would have guessed Heroku would have assigned it as a default somewhere

heroku sets `PORT` environment variable. You're missing the port directive in your config/puma.rb file


```ruby
port ENV['PORT']     || 3000
```

Without it puma doesn't know what port to use.

jayroh: Right, @schneems  - as it is now, in this PR, neither the Procfile or puma.rb set the port. I should have been more specific.

@gabebw, I'm with you. I set it in the Procfile and all is good.

calebthompson: eef3ccd

templates/Procfile
```diff
@@ -1,2 +1,2 @@
-web: bundle exec unicorn -p $PORT -c ./config/unicorn.rb
+web: bundle exec puma -p ${PORT:-3000} -C ./config/puma.rb
```
gabebw: I thought we agreed to not set a default?

templates/postgresql_database.yml.erb
```diff
@@ -4,9 +4,19 @@ development: &default
...
   timeout: 5000
 
 test:
   <<: *default
   database: <%= app_name %>_test
+
+production: &deploy
+  encoding: utf8
+  min_messages: warning
+  pool: <%%= [ENV.fetch("MAX_THREADS", 2), ENV.fetch("DB_POOL", 2)].max %>
```
stevenharman: Curious why we'd default to `MAX_THREADS` or `2` for the database pool, while defaulting each Puma worker to `MAX_THREADS` or `5`?

kewubenduben:
(y)

calebthompson:
I am swamped with other stuff right now - if anyone would like to take this over you have my blessing.

pedrosmmoreira:
I'd happy to help out with this.

tute:
@calebthompson what's left to do in here? Rebase, squash and test? @pedrosmmoreira may do it, and I could help this Friday too.

calebthompson:
It would be better to generate an app and throw it up on heroku with some crud scaffold to test the full stack. Suspenders tests aren't very comprehensive and this could break a lot.

croaky:
@tute We should get a few thoughtbot projects using this config in the real-world before codifying it into Suspenders. Upcase, Hound, FormKeep, thoughtbot.com, Maitre-d are all good candidates. If they run well for a few days or a week, then we can feel good about swapping out a major component (web server) as the new default.

bf4:
The other thing you can do is add a `.foreman` file that specifies some stuff, maybe?

seanpdoyle:
We've run into issues using `ember-cli-rails` and a forking server (i.e. unicorn).

https://github.com/rwz/ember-cli-rails/issues/94

My understanding is that puma can be configured to have a single, non-forking worker in development, which would smooth over the difficulties with `ember-cli-rails`.

I'm willing to try out puma on my next `ember-cli-rails` project (heroku or otherwise).

croaky:
I've used Puma a few times in the last month or so, for example on a production app serving an API for an active iPhone app. This change in general seems well-vetted now in prod.

@seandoyle I think you can set WEB_CONCURRENCY and MAX_THREADS to 1 each in development? Maybe we add that to .sample.env?

https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#config

seanpdoyle:
@drapergeek and I tried setting `UNICORN_WORKERS=1` to no avail.

I'll try out puma with `WEB_CONCURRENCY=1` and `MAX_THREADS=1`.

templates/unicorn.rb
buren: Have you addressed the exception issue detailed in https://github.com/heroku/rack-timeout/issues/76?

croaky:
@seanpdoyle How has Puma been running for FormKeep? Any settings in this PR that you recommend as different defaults?

templates/postgresql_database.yml.erb
halogenandtoast: The DB_POOL needs to be greater than or equal to the MAX_THREADS so defaulting it to 2 will lead to connection issues. On FormKeep we had this happen and since rack-timeout is on the main thread it killed the connection and left a prepared statement in the pool causing tons of errors. I suggest upping this to 5 to be in sync with the default on MAX_THREADS

croaky:
@calebthompson Possible to push this over the finish line? Seems like we're ready. @halogenandtoast had one piece of feedback above.

l4u:
:+1:

templates/postgresql_database.yml.erb
```diff
@@ -4,9 +4,19 @@ development: &default
...
+  timeout: 5000
```
gylaz: Is this a reasonable timeout for production?

croaky:
@calebthompsen I think you're available tomorrow? Possible to do a final review, merge this, and release a new version?

Seen a couple of brand-new apps recently getting slow client errors with Unicorn, Heroku, two dynos, almost no traffic.

Might be something worth grabbing from https://github.com/thoughtbot/hound/pull/846/files

templates/postgresql_database.yml.erb
calebthompson: I can't think of a reason it wouldn't be. Can you?

calebthompson:
Ready for final review.

calebthompson:
I'm content with this. If it's a :+1: please go ahead and merge this.

croaky:
Looks good to me.

mcmire:
:tada:

PR #502, opened 2015-01-27
Remove old logs and tempfiles

croaky:
Better matches Rails' default `bin/setup`:

https://github.com/rails/rails/blob/v4.2.0/railties/lib/rails/generators/rails/app/templates/bin/setup#L11

### DIFF
--- a/templates/bin_setup.erb
+++ b/templates/bin_setup.erb
@@ -21,6 +21,10 @@ bundle exec rake db:setup dev:prime
 # Add binstubs to PATH via export PATH=".git/safe/../../bin:$PATH" in ~/.zshenv
 mkdir -p .git/safe
 
+# Remove old logs and tempfiles
+rm -f log/*
+rm -rf tmp/cache
+
 # Pick a port for Foreman
 if ! grep --quiet --no-messages --fixed-strings 'port' .foreman; then
   printf 'port: <%= config[:port_number] %>\n' >> .foreman


mike-burns:
I'm all for consistency, but I didn't realize that 1) `bin/setup` should be run more than once, and 2) it should delete files. With a `-f`, even!

Overall, this doesn't bother me, but I do find (1) and (2) surprising, and I follow the POLA.

croaky:
`bin/setup should be run more than once`

Yup, it's been idempotent for awhile. Things like the db seeds or ENV variables in `.sample.env` can change. I find myself running `bin/setup` on projects a few times a week, after pulling new code, as long as I don't have development data I'm currently working with.

croaky:
It may not be surprising to me because I think of "setup" as "I'm setting myself up for this programming session I'm about to sit down for." rather than "I'm setting up this project for the first time and I'll never need to think about seeds and ENV variables again."

> it should delete files. With a `-f`, even!

Yeah, I can see the concern with that.

I actually don't feel strongly about this one way or the other. Saw it in the default Rails `bin/setup` and wanted to put it up for discussion.

gabebw:
I know that bin/setup is idempotent, but I run it once per project unless one of the other devs specifically asks me to run it again because they changed something.

I'm not against this, but I also don't think this gains us much - is it so bad to have old logs and tmp files?

croaky:
Yeah, I don't feel strongly. I think I'd rather not maintain this code or open up possibility for breakages. Closing without merging.

jferris:
For what it's worth, I also run `bin/setup` regularly, and my answer to many local development question is "did you try running `bin/setup` again?"

Issue #499, opened 2015-01-24
config.active_record.default_timezone default is already :utc


delphaber:
In this commit https://github.com/thoughtbot/suspenders/commit/d8af03a30921d192910b786d7d56745bb2b70639 you set config.active_record.default_timezone to :utc but rails default is already :utc

References:
http://guides.rubyonrails.org/configuring.html

> config.active_record.default_timezone determines whether to use Time.local (if set to :local) or Time.utc (if set to :utc) when pulling dates and times from the database. The default is :utc.

Could it be removed or am I missing something?

croaky:
Maybe our commit came before that was the default? Agreed, it could be removed.

PR #504, opened 2015-01-27
default_timezone value is :utc by default in Rails

delphaber:
Reference: https://github.com/thoughtbot/suspenders/issues/499


croaky:
Thanks, Merged as 605094e.

PR #505, opened 2015-01-27
Switch to Honeybadger

calebthompson:
In a recent conversation with @masonforest, we talked about how much more we
enjoyed using Honeybadger than any other exception trackers.

A lot of this stems from its great UI, but it also has a ton of integrations
for things we use in our stack (and those our clients are already using), and
some great features we haven't seen elsewhere.

* Super nice backtrace that can either mimic the Ruby/Rails backtrace or
  use their custom format which has nice features like links to view
  code from a gem on GitHub at the specific version you're using for
  each line in the stack:
  
* Offers integrations with much of our own stack: Trello,
  Slack/Campfire and GitHub; and with tools our clients are already
  using: HipChat, Flowdock, PagerDuty, and Pivotal Tracker.
* Has support for also tracking front-end JavaScript errors in concert
  with Rails errors https://github.com/honeybadger-io/honeybadger-js
* Awesome timeline of when errors occur, which has helped me see
  patterns that can be matched to user or admin actions or deploys,
  which has allowed me to easily determine causes for errors:


lib/suspenders/app_builder.rb
```diff
@@ -54,6 +54,10 @@ def configure_newrelic
       template 'newrelic.yml.erb', 'config/newrelic.yml'
     end
 
+    def configure_honeybadger
+      copy_file 'honeybadger.rb', 'config/initializers/honeybadger.rb'
```
calebthompson: Matches surrounding style.

lib/suspenders/generators/app_generator.rb
```diff
@@ -209,8 +210,10 @@ def remove_routes_comment_lines
     end
 
     def outro
-      say 'Congratulations! You just pulled our suspenders.'
-      say "Remember to run 'rails generate airbrake' with your API key."
+      say "Congratulations! You just pulled our suspenders."
+      say "Remember to run 'rails generate honeybadger --api-key' with your API key."
```
calebthompson: Should be 80 cols in output.

drapergeek:
@calebthompson is there a research card for this? I used honeybadger for a short period of time and enjoyed it but I haven't seen anyone else using it.

calebthompson:
I remember there having been one, but not the outcome. I wasn't able to find it in the archives.

jferris:
I don't see a Research card for this. Want to create one? It would be good to evaluate it as a group before standardizing around it in Suspenders.

calebthompson:
See above. There was a card, but it was lost in Trello history.

lib/suspenders/app_builder.rb
gylaz: I believe the general guideline is to convert the code we touch/change to the current style guide.

gylaz:
We've also been using Sentry on Hound, and it has been a positive experience. Not sure if this is worth any weight, but this was the first Google hit on "Sentry vs Honeybadger": http://stackshare.io/stackups/honeybadger-vs-sentry

drapergeek:
> We've also been using Sentry on Hound, and it has been a positive experience.

I think this reinforces the need for a research card before making this change.

croaky:
I'd be okay with just removing Airbrake and letting each project team decide which error tracker to use. It seems like there are multiple good choices and we're all good about picking one. Removing Airbrake also doesn't remove error tracking from Suspenders because Suspenders already has New Relic, which can handle errors.

gabebw:
I believe Formkeep uses https://www.skylight.io/ and likes it.

drapergeek:
We've also enjoyed skylight which is more of a competitor to newrelic than to airbrake. Maybe if we're considering dropping airbrake newrelic could also be reconsidered?

I don't have much opinion, I prefer sane defaults that can be overridden instead of having to choose each time.

jferris:
> I don't have much opinion, I prefer sane defaults that can be overridden instead of having to choose each time.

I agree with this, but I'm not sure how much we gain from having the Airbrake gem in there, since you have to set up an Airbrake project anyway.

calebthompson:
Okay. I'll close this and open another PR to only remove Airbrake.

PR #507, opened 2015-01-28
Only display user-facing flashes

croaky:
We sometimes use the flash for things such as
"sign up" in order to send data to analytics systems
such as Mixpanel.

We've then ended up doing the following on a number of projects.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -135,7 +135,8 @@ end
     end
 
     def create_shared_flashes
-      copy_file '_flashes.html.erb', 'app/views/application/_flashes.html.erb'
+      copy_file "_flashes.html.erb", "app/views/application/_flashes.html.erb"
+      copy_file "flashes_helper.rb", "app/helpers/flashes_helper.rb"
     end
 
     def create_shared_javascripts
--- a/templates/flashes_helper.rb
+++ b/templates/flashes_helper.rb
@@ -1,0 +1,5 @@
+module FlashesHelper
+  def user_facing_flashes
+    flash.to_hash.slice("alert", "error", "notice", "success")
+  end
+end
--- a/templates/_flashes.html.erb
+++ b/templates/_flashes.html.erb
@@ -1,6 +1,6 @@
 <% if flash.any? %>
   <div id="flash">
-    <% flash.each do |key, value| -%>
+    <% user_facing_flashes.each do |key, value| -%>
       <div class="flash-<%= key %>"><%= value %></div>
     <% end -%>
   </div>


templates/flashes_helper.rb
```diff
@@ -0,0 +1,5 @@
+module FlashesHelper
+  def user_facing_flashes
+    flash.to_hash.slice("alert", "error", "notice", "success")
```
derekprior: Is there a way to read from this api instead? http://blog.remarkablelabs.com/2012/12/register-your-own-flash-types-rails-4-countdown-to-2013

Seems like that'd be likely to be in line with what we want and more dry (register a flash type and you are all set). As long as you didn't register a type for the other uses of flash... Which I think is a fair tradeoff.

mcmire:
Why is this necessary? If you decided to store something in flash using a custom key, wouldn't you be surprised to find that it wasn't showing up on the page?

templates/flashes_helper.rb
croaky: I may be misunderstaing, but I just tried the following in an app:

* Back to `flash.each` in the `_flashes` partial.
* `add_flash_types :track_event` in the controller.
* `redirect_to edit_team_path(@team), track_event: "Created Team"` in the controller
* "Created Team" displays to the user, which it shouldn't, as I'm using the flash only for non-user facing event tracking

Is that what you were suggesting, or a different implementation?

croaky:
@mcmire I've been using it like:

```
app/views/application/_analytics.html.erb
19:    <% if flash[:track_event] %>
20:      window.analytics.track("<%= flash[:track_event] %>")
```

That key is not intended to be user-facing.

templates/flashes_helper.rb
tute: @derekprior do you mean grab the list of user facing flash keys from Rails? I couldn't find in https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/flash.rb a way to do that.

tute:
@mcmire I indeed found the opposite surprise in Upcase, when an iterator was showing many flash messages, some with non-useful data for the user. I'm in favor of using flash this way.

@croaky what's the status of your PR? Would you like to rebase on top of master?

mcmire:
What's the difference between `alert` and `notice`?

templates/flashes_helper.rb
derekprior: Yes. That's what I meant. Shame it doesn't exist but now I know what my
next rails pr is...

tute:
ping @croaky

lib/suspenders/app_builder.rb
```diff
@@ -127,7 +127,8 @@ def create_partials_directory
     end
 
     def create_shared_flashes
-      copy_file '_flashes.html.erb', 'app/views/application/_flashes.html.erb'
+      copy_file "_flashes.html.erb", "app/views/application/_flashes.html.erb"
```
tute: Do you want to introduce another commit normalizing all quotes in this file, rather than just this line here? Or use single quotes for this method?

tute:
@croaky this PR is getting stale, almost two months after your last activity. Should we rebase and merge this PR, or disregard it? I'm [gardening](http://words.steveklabnik.com/how-to-be-an-open-source-gardener) here. :)

croaky:
Sorry for the delay. Thanks for merging.

tute:
Thank you for rebasing!

Issue #510, opened 2015-02-03
Switch from `--heroku` flag to `--no-heroku` flag


croaky:
We've heard from a number of folks at thoughtbot that they forget to use the `--heroku` flag and then they need to manually add what Suspenders does for Heroku config.

gylaz:
:+1:

teoljungberg:
Makes sense, the anomaly would be to not use heroku in our case

tute:
I'd wait on https://github.com/thoughtbot/suspenders/pull/511#issuecomment-72885271, that will handle this case. Should we close this issue in favor of that PR?

croaky:
Yup, sounds good.

tute:
Closing in favor of https://github.com/thoughtbot/suspenders/pull/511.

Issue #488, opened 2015-01-14
Add a flag to turn off email


gabebw:
When an app doesn't send email (which is often), I have to rip out recipient interceptor and `SMTP_SETTINGS`, otherwise the `ENV.fetch` in `SMTP_SETTINGS` crashes deploys on Heroku.

croaky:
Agreed. I like this idea.

gylaz:
Ran into Heroku deploy issue today, and wondered why `STMP_*` settings aren't in the `.sample.env` file, so I can at a glance see what config vars I need to pre-define on Heroku before trying to deploy.

tute:
@gylaz `ASSET_HOST` seems to be needed in `.sample.env` too, probably there's an issue/PR waiting there.

Should we close this issue in favor of https://github.com/thoughtbot/suspenders/pull/511?

croaky:
Sounds good to close in favor of #511.

tute:
Closing in favor of #511.

toobulkeh:
Any hints in .sample.env towards what these variables should be would help too!
I like the sendgrid hint, but being new to heroku too not sure if I should set the ASSET_HOST manually to the Heroku URL or just jump into a CDN.

Also, setting heroku variables was a pain. couldn't .env.staging be used for this? Maybe I'm just missing something. 'scuse my ignorance.

jferris:
> Also, setting heroku variables was a pain. couldn't .env.staging be used for this? Maybe I'm just missing something. 'scuse my ignorance.

Storing the environment-specific configuration in environment variables is one of the tenants of Heroku.

Heroku says this in their [dev center](https://devcenter.heroku.com/articles/config-vars):

> The traditional approach for handling such config vars is to put them under source - in a properties file of some sort. This is an error-prone process, and is especially complicated for open source apps which often have to maintain separate (and private) branches with app-specific configurations.
>
> A better solution is to use environment variables, and keep the keys out of the code.

Practically, using `.env.staging` has the following disadvantages:

* You need to commit and deploy when changing any configuration.
* You'll need to store sensitive configuration (like passwords and API keys) in a Git repository.

Heroku has some more to say about config in their [Twelve-Factory App](http://12factor.net/config) essay.

toobulkeh:
Fair enough jferris, that makes a lot of sense! Thanks for the intelligent writeup and sources!

CDainMiller:
Wow, I had no idea Heroku authored that 12factor site. 

[...]

PR #511, opened 2015-02-04
Opt in to the Rails differences

mike-burns:
This changes Suspenders to be an opt-in system instead of opt-out. That is, when you create a fresh Suspenders project, it is ready for you to uncomment and otherwise activate all the fancy Suspenders features -- but none of those features are active by default.

The idea in this branch is: if a gem can be unused, then it is.

This is a work in progress pull request, intended to get early discussion and some help.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -196,13 +196,11 @@ end
     end
 
     def configure_i18n_tasks
-      run "cp $(i18n-tasks gem-path)/templates/rspec/i18n_spec.rb spec/"
       copy_file "config_i18n_tasks.yml", "config/i18n-tasks.yml"
     end
 
     def configure_background_jobs_for_rspec
       copy_file 'background_jobs_rspec.rb', 'spec/support/background_jobs.rb'
-      run 'rails g delayed_job:active_record'
     end
 
     def configure_action_mailer_in_specs
@@ -216,10 +214,6 @@ end
 
     def configure_rack_timeout
       copy_file 'rack_timeout.rb', 'config/initializers/rack_timeout.rb'
-    end
-
-    def configure_simple_form
-      bundle_command "exec rails generate simple_form:install"
     end
 
     def configure_action_mailer
@@ -258,10 +252,6 @@ end
 
     def install_bitters
       run "bitters install --path app/assets/stylesheets"
-    end
-
-    def install_refills
-      run "rails generate refills:import flashes"
     end
 
     def gitignore_files
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -349,7 +349,7 @@ you can deploy to staging and production with:
 
     def setup_bundler_audit
       copy_file "bundler_audit.rake", "lib/tasks/bundler_audit.rake"
-      append_file "Rakefile", %{\ntask default: "bundler:audit"\n}
+      append_file "Rakefile", %{\n# task default: "bundler:audit"\n}
     end
 
     def copy_miscellaneous_files
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -90,12 +90,6 @@ feature 'Suspend a new project with default configuration' do
     end
   end
 
-  scenario "specs for missing or unused translations" do
-    run_suspenders
-
-    expect(File).to exist("#{project_path}/spec/i18n_spec.rb")
-  end
-
   scenario "config file for i18n tasks" do
     run_suspenders
 
@@ -109,12 +103,6 @@ feature 'Suspend a new project with default configuration' do
     app_name = SuspendersTestHelpers::APP_NAME
 
     expect(locales_en_file).to match(/application: #{app_name.humanize}/)
-  end
-
-  scenario "config simple_form" do
-    run_suspenders
-
-    expect(File).to exist("#{project_path}/config/initializers/simple_form.rb")
   end
 
   def analytics_partial
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -4,8 +4,11 @@ ruby "<%= Suspenders::RUBY_VERSION %>"
 
 # gem "airbrake"
 
-gem "bourbon", "~> 4.1.0" ### TODO
-gem "coffee-rails", "~> 4.1.0" ### TODO
+# Enable and then edit:
+#   app/assets/application.css.scss
+# gem "bourbon", "~> 4.1.0"
+
+# gem "coffee-rails", "~> 4.1.0"
 
 # Enable and then run:
 #   rails generate delayed_job:active_record
@@ -23,13 +26,18 @@ gem "coffee-rails", "~> 4.1.0" ### TODO
 #   config/i18n-tasks.yml
 # gem "i18n-tasks"
 
-gem "jquery-rails" ### TODO
-gem "neat", "~> 1.7.0" ### TODO
+# gem "jquery-rails"
+
+# Enable and then edit:
+#   app/assets/application.css.scss
+# gem "neat", "~> 1.7.0"
 
 # Enable and then edit:
 #   config/newrelic.yml
 # gem "newrelic_rpm"
 
+# Enable and then edit:
+#   app/assets/application.css.scss
 # gem "normalize-rails", "~> 3.0.0"
 
 gem "pg"
@@ -42,11 +50,13 @@ gem "rails", "<%= Suspenders::RAILS_VERSION %>"
 
 # gem "recipient_interceptor"
 
-# Enable and then run:
+# Enable and then edit:
+#   app/assets/application.css.scss
+# And then run:
 #   rails generate refills:import flashes
 # gem "refills"
 
-gem "sass-rails", "~> 5.0" ### TODO?
+# gem "sass-rails", "~> 5.0"
 
 # Enable and then run:
 #   rails generate simple_form:install
@@ -54,7 +64,7 @@ gem "sass-rails", "~> 5.0" ### TODO?
 
 gem "title"
 
-gem "uglifier" ### TODO?
+# gem "uglifier"
 
 gem "unicorn"
 
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -4,8 +4,8 @@ ruby "<%= Suspenders::RUBY_VERSION %>"
 
 # gem "airbrake"
 
-gem "bourbon", "~> 4.1.0"
-gem "coffee-rails", "~> 4.1.0"
+gem "bourbon", "~> 4.1.0" ### TODO
+gem "coffee-rails", "~> 4.1.0" ### TODO
 
 # Enable and then run:
 #   rails generate delayed_job:active_record
@@ -23,8 +23,8 @@ gem "coffee-rails", "~> 4.1.0"
 #   config/i18n-tasks.yml
 # gem "i18n-tasks"
 
-gem "jquery-rails"
-gem "neat", "~> 1.7.0"
+gem "jquery-rails" ### TODO
+gem "neat", "~> 1.7.0" ### TODO
 
 # Enable and then edit:
 #   config/newrelic.yml
@@ -46,7 +46,7 @@ gem "rails", "<%= Suspenders::RAILS_VERSION %>"
 #   rails generate refills:import flashes
 # gem "refills"
 
-gem "sass-rails", "~> 5.0"
+gem "sass-rails", "~> 5.0" ### TODO?
 
 # Enable and then run:
 #   rails generate simple_form:install
@@ -54,7 +54,7 @@ gem "sass-rails", "~> 5.0"
 
 gem "title"
 
-gem "uglifier"
+gem "uglifier" ### TODO?
 
 gem "unicorn"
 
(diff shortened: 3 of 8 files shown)


spec/features/new_project_spec.rb
```diff
@@ -111,12 +105,6 @@
     expect(locales_en_file).to match(/application: #{app_name.humanize}/)
   end
 
-  scenario "config simple_form" do
-    run_suspenders
-
-    expect(File).to exist("#{project_path}/config/initializers/simple_form.rb")
-  end
-
```
mike-burns: Instead of removing these tests, let's enable the required gem inside the test and make sure it still works.

jferris:
This is an interesting approach, but I'm not sold that it will be an improvement.

I like that Suspenders is opinionated now. I like that I don't need to do anything if we're adopting our default opinions. I also don't think this would save me work, because the commented out code would be distracting, and I think the first thing I'd do would be to uncomment everything or remove it.

If there are things we want to change to be opt-in, I'd be interested to see if we could add some generators to Suspenders. As much as possible, we try to make things in Suspenders just configuration. When that works, the basic pattern for activating something should be adding a gem to `Gemfile`, `bundel install`, running a generator, and then pulling in our opinionated config. I think we could make generators that wrap up those steps for each gem we don't always need (delayed_job, for example).

mike-burns:
The end goal is: I want to use Suspenders to get a base Rails app, but when I want to add something (e.g. background jobs) I can just turn on the thoughtbot-approved setting and gem for it.

A commented-out Gemfile seems easiest, but I'm open to other solutions.

croaky:
I've been thinking we'd go with the generator approach to solve this problem as well.

suspenders add email
suspenders add background_jobs
etc.

mike-burns:
My only concern with a generator-based approach (aside from the fact that it's more code) is that it's yet another thing to learn: people will need to learn `suspenders add X` and `suspenders list` (and `suspenders remove X`?). Uncommenting code in a Gemfile has a lower mental overhead.

What's the benefit of a generator?

jferris:
Pro-generators:

* You don't need to hunt down the related snippets and uncomment them together.
* You can't uncomment a gem without realizing that there's useful config for it or that you should run a generator.
* Using a generator keeps all the related elements (dependencies, config, setup) together, so it's obvious where all the Suspenders code for a particular concern is.
* Running generators is a familiar action to Rails developers.
* Generators are easier to test.
* You can update Suspenders and get new generators in an existing app.

mike-burns:
Joe are you picturing something like this:

```sh
rails g suspenders:add email
```

And a suspenders gem in the Gemfile providing that?

jferris:
I think just `rails g suspenders:email` would be fine as a generator. I think we could also hide it behind `suspenders add email` like @croaky suggested. I'm not sure what's more intuitive. I guess Rails generators already have a known method of discovery and execution, so Suspenders would just have to add itself to the Gemfile.

mike-burns:
Well, the work done on this branch is required for the generator, so that's good.

How about merging the comment-based version as I've done it here, then we can work on the generator that automates it?

(Not that this PR is ready to be merged, but I'm trying to make small diffs.)

jferris:
> How about merging the comment-based version as I've done it here, then we can work on the generator that automates it?

I think the comment-based version is a step backwards, because it creates a lot of work for new applications in order to get the common Suspenders requirements. I also believe the commented out code is unlikely to be used, because people will want to delete it.

If we moved to a generator-based model, I think that would be acceptable, because you could quickly pull in all of the generators you wanted to use.

croaky:
Couple more generator benefits over code comments:

* The comments are generated days, or weeks, before they are turned on, and can be out of date.
* A generator is a small, well-defined API to manage all the setup and config. Comments across the project have no API, which can introduce the kinds of mistakes that Suspenders is trying to address.

> How about merging the comment-based version as I've done it here, then we can work on the generator that automates it?

It's possible that the original code (before the comment change) could be more easily extracted to individual generators. The lines that were deleted use Thor's API.

> I'm trying to make small diffs

I like this goal. What if we took it in feature-based steps? Extract the `add heroku` command first, then `add email`, then `add background_jobs`, etc.?

> I guess Rails generators already have a known method of discovery and execution

I think gems installed out of the `Gemfile` also show up in the `rails generate` list as an option but I could be mistaken. Might be nice to not have a gem in the Gemfile that only adds generators. Then again, maybe it's a sign that Suspenders should be used to add new features. Could potentially just add that to the `README.md` or `CONTRIBUTING.md`, though? No strong opinion from me at this point.

Side note: moving to generators also addresses https://github.com/thoughtbot/suspenders/issues/510 and https://github.com/thoughtbot/suspenders/issues/488 in an elegant, consistent way.

edwardloveall:
> The basic pattern for activating something should be adding a gem to `Gemfile`, `bundle install`, running a generator, and then pulling in our opinionated config.

I really like the idea of this. Especially if there's a way to specify a config before hand. I was working on a PR that ran your own local modification file after suspenders finished, similar to [~/.laptop.local](https://github.com/thoughtbot/laptop#customize-in-laptoplocal) but it seems like this idea would replace that.

lib/suspenders/app_builder.rb
```diff
@@ -196,13 +196,11 @@ def configure_i18n_for_missing_translations
     end
 
     def configure_i18n_tasks
```
jessieay: seems like we shouldn't include the `yml` file if we are removing the spec file that uses it?

templates/Gemfile.erb
```diff
@@ -2,53 +2,104 @@ source "https://rubygems.org"
...
-gem "delayed_job_active_record"
-gem "email_validator"
-gem "flutie"
-gem "high_voltage"
-gem "i18n-tasks"
-gem "jquery-rails"
-gem "neat", "~> 1.7.0"
-gem "newrelic_rpm"
-gem "normalize-rails", "~> 3.0.0"
+# gem "airbrake"
```
jessieay: hmmm I don't love generated files with tons of comments. I would personally rather see gems that are required and then nothing else.

mcmire:
Yeah, I like the idea of generators too. Many projects may start out the same -- I probably want Bourbon and Neat, I don't want Turbolinks, I probably want to use SimpleForm -- but require different things as time goes on -- I may want Sidekiq over DelayedJob, I may not want Spring, I may want Carrierwave over Paperclip, I may want Devise over Clearance (god help me), I may want to add Konacha/Mocha/Chai or Teaspoon/Jasmine, I may want to add Bower... you get the idea. If there were a way to drop in these things after you create the initial foundation, that would be pretty sweet. I think the generator route makes a lot of sense -- it certainly works for plenty of gems.

templates/Gemfile.erb
tute: Agreed we shouldn't have commented stuff in the Gemfile.

mike-burns:
I will not be offended if someone uses the knowledge from this branch to make the generators happen. I won't be able to get to it for a couple of weeks.

jferris:
>  I may want Sidekiq over DelayedJob, I may not want Spring, I may want Carrierwave over Paperclip, I may want Devise over Clearance

Part of the purpose of this repo is to have an opinion. It's been helpful to have defaults, and one of the ways we codify our majority opinions as a group is to add things to Suspenders. That way, when somebody starts a project, they don't have to ask, "does thoughtbot generally prefer Sidekiq or DelayedJob?" The gem will already be installed.

If a project team makes a conscious decision to override that default choice, I'm fine with that. However, I'd prefer to make the default, majority choice zero-effort and make it so people who stray from the default at least have to realize they're doing it.

mike-burns:
I do not want it to be zero effort to use the recommended choice, but I do want to to be like this:

(using non-recommended choice) is harder than (using recommended choice) is harder than (using `rails new`).

This was my goal in this branch, and I'd like to see it be the goal of the branch that replaces it.

edwardloveall:
Part of the problem I think is it's starting to become more effort to use Suspenders with all the defaults. Setting up things like Airbrake and New Relic or an asset host at the start of a project just puts up barriers to entry.

Maybe `suspenders add defaults` could install the opinionated thoughtbot defaults to a project without adding too much complexity.

mike-burns:
Yes, that is the idea that was proposed in this very PR.

jferris:
> Setting up things like Airbrake and New Relic or an asset host at the start of a project just puts up barriers to entry.

Ideally, these only put up barriers to deployment. Out of the box, Rails applications are not really deployable; you need to tweak the configuration considerably. What's possibly worse is that there are things you can forget which seem to work, but are actually mistakes.

The two you mentioned are good examples. In our opinion, every production application should have an error notifier and performance monitor in place. Suspenders will crash if you fail to address this issue when deploying, so you can never make the mistake of releasing an app into the wild without essential services in place.

edwardloveall:
>every production application should have an error notifier and performance monitor in place

Yeah that's a good point and something I hadn't really considered. I guess my hope is that there's a way to make suspenders more modular and customizable for elements that are unnecessary for some apps. An example could be not including Bourbon for JSON APIs. I've always seen the opinion that suspenders has less as "all Rails apps should include Bourbon" and more "when a SCSS library is needed, Bourbon is the one to use".

I think a one-size-fits-all rails apps isn't as useful as it once was.

jferris:
>  My hope is that there's a way to make suspenders more modular and customizable for elements that are unnecessary for some apps. An example could be not including Bourbon for JSON APIs. I've always seen the opinion that suspenders has less as "all Rails apps should include Bourbon" and more "when a SCSS library is needed, Bourbon is the one to use".

Definitely, and there's been some discussion of this. The idea (as I understand it) has evolved to:

* Suspenders will install fewer things by default.
* Suspenders will come with generators for adding things you might want.
* The generators will be generic (`rails g install suspenders:auth`) but will use a particular library (Clearance).

I think this goes some way towards providing low-effort defaults for common needs, but removes the bloated feeling of using Suspenders for a JSON API.

I think this will be harder than it sounds, though, because there is likely some interaction between some of the things we consider orthogonal, like the choice of background jobs and deploying to Heroku.

edwardloveall:
A common base with optional add-ons sounds like a great solution. My thought on generic naming for the components is the same. `auth` instead of `clearance`. That way it's task based rather than gem based.

:+1:

toobulkeh:
:+1: as someone who just touched suspenders for the first time today, much needed here.

templates/Gemfile.erb
mcmire: I think this is fine for now, until we work out the generator solution.

lib/suspenders/app_builder.rb
```diff
@@ -196,13 +196,11 @@ def configure_i18n_for_missing_translations
...
-      run "cp $(i18n-tasks gem-path)/templates/rspec/i18n_spec.rb spec/"
       copy_file "config_i18n_tasks.yml", "config/i18n-tasks.yml"
     end
 
     def configure_background_jobs_for_rspec
       copy_file 'background_jobs_rspec.rb', 'spec/support/background_jobs.rb'
-      run 'rails g delayed_job:active_record'
```
mcmire: I don't think this PR is shippable with these lines completely removed, but with lines in the Gemfile merely commented out. It doesn't seem very helpful if you wanted to use Refills, for instance (you'd have to remember to un-comment out the line in the Gemfile, but then run `rails generate refills:import flashes`).

mcmire:
@mike-burns:

In this PR we are opting into the differences, but then later you said:

> I do not want it to be zero effort to use the recommended choice, but I do want to to be like this:

> (using non-recommended choice) is harder than (using recommended choice) is harder than (using rails new).

That sounds like you want it to be opt-out rather than opt-in. That actually makes more sense to me. So to get our standard stack, you'd just say:

```
suspenders create my-awesome-app
```

and you'd get DelayedJob, Refills, SimpleForm, etc.

But if you knew you wanted Sidekiq over DelayedJob right from the get-go, for instance, you could say:

```
suspenders create my-awesome-app --without-delayed-job
```

and then you *might* could say 

```
suspenders add sidekiq
```

or, you could just add it yourself.

The alternative is:

```
suspenders create my-awesome-app   # installs few things
suspenders add delayed_job
suspenders add simple_form
suspenders add refills
...
```

That seems like a lot of commands to run for the standard Rails app case.

Perhaps this problem could be solved by adding "meta-generators". So, for instance:

```
suspenders create my-awesome-app
suspenders add rails   # also could use "api" for an API-only app, for instance
```

toobulkeh:
What about an interactive middle ground?

Choose your asynchronous handler:
1. ActiveJob
2. Sidekiq
3. None

jferris:
> Choose your asynchronous handler:

I mentioned this in an [earlier comment](https://github.com/thoughtbot/suspenders/pull/511#issuecomment-76731523), but we want Suspenders to be opinionated. If we provide configuration for a background runner, we should provide configuration for the one we think we'll use most of the time.

I'm also not sure it's realistic for us to maintain configuration for several solutions for each of the problems Suspenders caters to.

toobulkeh:
Maybe just an interactive mode with YES/NO for a single configuration then?

Include an asynchronous handler? [yn]

mike-burns:
There is 0% chance I'll use an interactive thing.

@mcmire no, I meant the opposite: I want suspenders to make a vanilla Rails app, but provide guidance for how to add new libraries -- Suspenders should answer the question "I want to add an authentication library; which should I use?".

I haven't seen what you call "the standard case" since we extracted Suspenders (from Hoptoad?). And even then, I'm not sure I saw the standard case.

I'm going to close this PR now. I look forward to the one that adds generators.

PR #512, opened 2015-02-05
travis_retry the bin/setup script

Neson:
The `bundle install` command has possibilities facing network timeouts, causing build failures. By default, Travis will retry install commands three times if it fails. 

However, this behavior isn't enabled while using a custom setup script, only through running `travis_retry bin/setup` instead of `bin/setup` manually tells Travis use this functionality.

http://blog.travis-ci.com/2013-05-20-network-timeouts-build-retries/

### DIFF
--- a/templates/travis.yml.erb
+++ b/templates/travis.yml.erb
@@ -4,7 +4,7 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 install:
-  - bin/setup
+  - travis_retry bin/setup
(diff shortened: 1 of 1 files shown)


tute:
@joshk does this look good to you? Thank you! :)

joshk:
instead of using `travis_retry`, why not use the built in bundler goodness of `--jobs` and `--retry`?

Neson:
I didn’t know that bundler has these features until now, thanks for the advise!

PR #513, opened 2015-02-07
Remove unused method

delphaber:


tute:
Won't Rails run `bundle install` before suspender does, running it twice?

drapergeek:
Yes I think this method is called by the normal rails app generator and we're just putting it off until later.

croaky:
Yes, I think the intent of this method to prevent Bundler from running twice. If this commit were to be merged, I think we'd revert to that, slower behavior.

We invoke `bundle install` at the moment we want it, inside the `customize_gemfile` method, on line 61 of `lib/suspenders/generators/app_generator.rb`

lib/suspenders/generators/app_generator.rb
```diff
@@ -212,10 +212,6 @@ def outro
       say "Remember to run 'rails generate airbrake' with your API key."
     end
 
-    def run_bundle
-      # Let's not: We'll bundle manually at the right spot
-    end
-
```
bf4: I think this is meant to over-write `run_bundle` defined in `app_base`

- https://github.com/rails/rails/blob/4-2-stable/railties/lib/rails/generators/app_base.rb#L334
- https://github.com/rails/rails/blob/4-2-stable/railties/lib/rails/generators/rails/app/app_generator.rb#L259

Although I don't know why you don't just set skip_bundle to true https://github.com/rails/rails/blob/4-2-stable/railties/lib/rails/generators/app_base.rb#L32

delphaber:
Oh I'm sorry. Didn't think about inheritance,

tute:
If you'd like to update the PR, `skip_bundle` could be an improvement, making the code clearer IMO.

croaky:
Agreed on `skip_bundle`. Clearer, more intentional API.

delphaber:
Ok I'm gonna update it :)

delphaber:
Updated :)

mcmire:
:+1:

croaky:
Thanks!

Issue #491, opened 2015-01-17
Comment out ASSET_HOST setting?


gabebw:
I think this is good to have but is unlikely to be set up within the first, say, 2 months of a project's existence (if ever). Since suspenders' target audience is new projects, the following line crashes Heroku pretty consistently and project owners are unlikely to have something to set `ASSET_HOST` to:

```
config.action_controller.asset_host = ENV.fetch("ASSET_HOST")
```

Can we generate it but keep it commented? Thoughts?

croaky:
I usually set it to HOST (staging.project.com) at first, then switch to a CDN at "prep for public launch" time.

gabebw:
Ah, that makes sense. Maybe worth a comment above that line saying that?

gabebw:
Oh, or even:

```ruby
= ENV.fetch("ASSET_HOST", ENV.fetch("HOST"))
```

croaky:
Clever. I think I like it.

tute:
If now we require `ASSET_HOST` for production, shouldn't it exist in `.sample.env` file?

I like `ENV.fetch("ASSET_HOST", ENV.fetch("HOST"))`.

croaky:
> If now we require `ASSET_HOST` for production, shouldn't it exist in `.sample.env` file?

Yup, sound good.

PR #517, opened 2015-02-14
Add ASSET_HOST to sample.env

tute:
It defaults to `HOST`.

Related with issue: https://github.com/thoughtbot/suspenders/issues/491


tute:
@gabebw would this PR solve https://github.com/thoughtbot/suspenders/issues/491?

gabebw:
Yes, I think so.

PR #518, opened 2015-02-14
Add mongoid support;

alexkravets:


tute:
Suspenders is a script thoughtbot uses as defaults for our projects. We default to Postgres for now, not Mongo.

It would be counterproductive to accept this option as I don't use this DB/ORM, I wouldn't know how to test or keep up to date, and I couldn't then maintain it.

Thank you very much for your work.

alexkravets:
Got it, makes sense! Thank you.

PR #521, opened 2015-02-16
Set email deliver method to :test for development

### LINKED ISSUE #516: Provide a better default ActionMailer `delivery_method` in development env
It seems that Rails 4.2 defaults to `:deliver` as the default ActionMailer `delivery_method`. Thus, emails will try to deliver via `sendmail`. Seems not very useful in most dev setups. Should we default that to `:test` in development environment instead?

gylaz:
`:test` method will not send the email, but instead print it to the logs. This 
will allow faster development cycle, since there is no need to setup 
smtp/sendmail when email is involved.

Closes #516

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -11,6 +11,14 @@ module Suspenders
         'raise_delivery_errors = false', 'raise_delivery_errors = true'
     end
 
+    def set_test_delivery_method
+      inject_into_file(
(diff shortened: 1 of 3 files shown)


mcmire:
:+1:

tute:
Merged in https://github.com/thoughtbot/suspenders/commit/8b45fd42b29b71269ca239534f7dc77b8215d8ab. Thank you!

PR #522, opened 2015-02-17
Include missing word in the Flutie description.

aaronagray:
I had a hard time comprehending the Flutie gem's description before I realized it must be missing the `page_title` bit.


tute:
Thank you!

PR #523, opened 2015-02-19
Require the spec_helper file within rails_helper

dgalarza:
With the switch over to a rails_helper file in RSpec 3.x, the
spec_helper file should be required from within the rails_helper as
well.

### DIFF
--- a/templates/rails_helper.rb
+++ b/templates/rails_helper.rb
@@ -1,5 +1,7 @@
 ENV["RAILS_ENV"] = "test"
 
+require "spec_helper"
+
 require File.expand_path("../../config/environment", __FILE__)
 
 require "rspec/rails"


JoelQ:
:+1:

JoelQ:
I'm surprised this wasn't already happening?

BlakeWilliams:
:+1:

gylaz:
> I'm surprised this wasn't already happening?

It actually is for `suspenders` apps, because the default `.rspec` requires `spec_helper.rb` by default. See: https://github.com/thoughtbot/suspenders/pull/410/files#diff-3c4866bd3bcfb6d6f37077dee0aefc95R3

dgalarza:
It appears that this is not longer the case @gylaz. Looking at the `rails_helper` that is now generated by `rspec-rails` it looks like they've moved the require directly into the `rails_helper`.

https://github.com/rspec/rspec-rails/commit/49138fb5415f6d89b3d74d7052b0acc3563f3fd3

gylaz:
FWIW, the current release of the `rspec-rails` gem (version 3.2.0), my generated `.rspec` contains:
```
--color
--require spec_helper
```

pedrosmmoreira:
It might help to refer to the discussion on https://github.com/thoughtbot/suspenders/pull/410

dgalarza:
Thanks for the heads up @gylaz and @pedrosmmoreira. I had recently started a project where we didn't have the `.rspec` file generated which is why I ended up making this PR. I just tried again with a fresh app built from suspenders and the `.rspec` file was generated just fine. Closing this out.

Issue #313, opened 2014-04-07
Remove SimpleCov in favor of CodeClimate


croaky:
https://codeclimate.com/changelog/5313a93ce30ba060970056fb

croaky:
SimpleCov was removed from Suspenders in June in 6ece3c4, partially due to that Travis bug that appears to be fixed now.

bf4:
@croaky let me know if I can help re: simplecov in any way.

tute:
Is there anything suspenders can do to easy the set up of SimpleCov with CC? Should we have SimpleCov compute coverage locally, anyway, however we run CI?

jferris:
> Should we have SimpleCov compute coverage locally, anyway, however we run CI?

I think that would be the best solution. The project needs SimpleCov set up correctly to support Code Climate, so we may as well make it easy to see coverage locally. Code Climate just makes a convenient public report and notifier when it comes to coverage.

bf4:
Check out the bin dir for how you can dump reports to the file system,
then send along, like on tddium

tute:
Please see PR: https://github.com/thoughtbot/suspenders/pull/524.

tute:
Closing in favor of PR.

PR #524, opened 2015-02-20
Add SimpleCov

tute:
It was removed due to a return status bug which would mark failed builds as
successful, and is now fixed.

This reverts commit 6ece3c4bbea3825802e16c41724fa062b91b7c29.

Related with: https://github.com/thoughtbot/suspenders/issues/313

### DIFF
--- a/templates/spec_helper.rb
+++ b/templates/spec_helper.rb
@@ -1,3 +1,8 @@
+if ENV.fetch("COVERAGE", false)
+  require "simplecov"
+  SimpleCov.start "rails"
+end
+
 require "webmock/rspec"
 
 # http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    suspenders (1.25.0)
+    suspenders (1.26.0)
       bitters (~> 1.0.0)
       bundler (~> 1.3)
       rails (= 4.2.1)
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -47,6 +47,7 @@ group :test do
   gem "formulaic"
   gem "launchy"
   gem "shoulda-matchers", require: false
+  gem "simplecov", require: false
   gem "timecop"
   gem "webmock"
 end


tute:
> You can use the simplecov multiformatter to still use html AND codeclimate.

@bf4 how does this work?

pedrosmmoreira:
I believe something like this should work: 
```ruby
if ENV.fetch("CODECLIMATE_REPO_TOKEN") { false }
  require "codeclimate-test-reporter"
  CodeClimate::TestReporter.configure do |config|
    config.logger.level = Logger::WARN
  end
  CodeClimate::TestReporter.start
end

if ENV.fetch("COVERAGE") { false }
  require "simplecov"
  SimpleCov.start "rails"
end

**Comment by bf4** [comment]
I've done something like this (relevant part is `SimpleCov.formatters =`, but I thought the rest would be useful to look at as well):

```ruby
# spec/spec_helper.rb
require "simplecov"

# .simplecov
# https://github.com/colszowka/simplecov#using-simplecov-for-centralized-config
# see https://github.com/colszowka/simplecov/blob/master/lib/simplecov/defaults.rb
# vim: set ft=ruby
SimpleCov.profiles.define "suspenders" do
  load_profile "rails"

  add_group "Long files" do |src_file|
    src_file.lines.count > 100
  end
  class MaxLinesFilter < SimpleCov::Filter
    def matches?(source_file)
      source_file.lines.count < filter_argument
    end
  end
  add_group "Short files", MaxLinesFilter.new(5)

  # Exclude these paths from analysis
  add_filter "bundle"
  add_filter "bin"
  add_filter "tasks"
end

## RUN SIMPLECOV
if defined?(@running_tests)
  @running_tests = false
else
  @running_tests = caller.any? { |line| line =~ /exe\/rspec/ }
end

if @running_tests
  SimpleCov.start "suspenders"
end

if ENV["COVERAGE"] =~ /\Atrue\z/i
  if ENV['CODECLIMATE_REPO_TOKEN']
    begin
      require "codeclimate-test-reporter"
      # We run in parallel, so output results to disk via TO_FILE
      # and send at end, like tddium config
      # https://github.com/codeclimate/ruby-test-reporter/blob/master/lib/code_climate/test_reporter/formatter.rb#L18
      ENV['TO_FILE'] = 'true'
      SimpleCov.formatters = [
          SimpleCov::Formatter::HTMLFormatter,
          CodeClimate::TestReporter::Formatter
      ]
    rescue LoadError
      STDERR.puts "Could not load CodeClimate SimpleCov Reporter"
      puts "[COVERAGE] Running with SimpleCov HTML Formatter"
      SimpleCov.formatters = [SimpleCov::Formatter::HTMLFormatter]
    end
else
  SimpleCov.formatters = []
end

SimpleCov.at_exit do
  @minimum_coverage = ENV.fetch("COVERAGE_MINIMUM") { 100.0 }.to_f.round(2)
  SimpleCov.result.format!
  percent = Float(SimpleCov.result.covered_percent)
  if percent < @minimum_coverage
    puts "Spec coverage was not high enough: #{percent.round(2)} is < #{@minimum_coverage}%"
    exit 1
  else
    puts "Nice job! Spec coverage (#{percent.round(2)}) is still above #{@minimum_coverage}%"
  end
end
```

templates/spec_helper.rb
```diff
@@ -1,3 +1,6 @@
+require "simplecov"
+SimpleCov.start "rails"
```
jessieay: should this be in `rails_helper`  that's where I've seen it in a recent project that used SimpleCov. (although it looks like we were almost always requiring `rails_helper` rather than `spec_helper`, which is not ideal)

jessieay:
One comment, but just a clarification. Good catch! :balloon:

templates/spec_helper.rb
bf4: Before you load any app code, so at the top of spec_helper and Rakefile

PR #525, opened 2015-02-24
Provide EXECJS_RUNTIME env variable

gylaz:
This is necessary for compiling assets (when needed to run with compiled assets
in dev).

### DIFF
--- a/templates/sample.env
+++ b/templates/sample.env
@@ -3,3 +3,4 @@ ASSET_HOST=localhost:3000
 HOST=localhost:3000
 RACK_ENV=development
 SECRET_KEY_BASE=development_secret
+EXECJS_RUNTIME=Node


derekprior:
Doesn't execjs detect an available runtime?

gylaz:
@derekprior Probably, but in my case whatever it detects throws an error like:
```
ExecJS::RuntimeError: Unrecognized node in FTL backend
```
Is there a way to see what it's detecting?

The error goes away when I set `EXECJS_RUNTIME=Node`.

derekprior:
https://github.com/sstephenson/execjs/blob/master/lib/execjs/runtimes.rb#L71

It looks to me like it's likely choosing JavaScriptCore which comes with OS X. Wonder if there's something you're doing it can't handle? Further wonder what heroku uses when it compiles. I can see an argument for being sure to use the same if you are testing things localy.

gylaz:
I just checked two different projects and heroku console reports Node as the runtime:
```
irb(main):001:0> ExecJS::Runtimes.best_available
<ExecJS::ExternalRuntime:0x007faf6cc26190 @name="Node.js (V8)", @command=["nodejs", "node"] ...
```

bf4:
I have, in the past, used code such as

```ruby
  if defined?(JRUBY_VERSION) && defined?(ExecJS) && system('which node')
    STDOUT.puts "Using Node ExecJS runtime"
    ExecJS.runtime = ExecJS::Runtimes::Node
  end
```

And I think the node engine compares pretty well to therubyracer.

jferris:
Heroku [installs node](https://github.com/heroku/heroku-buildpack-ruby/blob/0714d7e2eb0091de7b0af8d632343386854d94d2/lib/language_pack/helpers/node_installer.rb) for Ruby apps.

tute:
@gylaz is this still an issue?

gylaz:
@tute To me it seems like a good idea to have parity with Heroku. So, since it's `Node` on Heroku, I think we should be using that locally.

tute:
> @tute To me it seems like a good idea to have parity with Heroku. So, since it's Node on Heroku, I think we should be using that locally.

Makes sense to me. Feel free to rebase and merge @gylaz.

tute:
Ping @gylaz

PR #526, opened 2015-02-27
Configure ActiveJob's queue adapter to DelayedJob

### LINKED ISSUE #520: ActiveJob
What are thoughbot's thoughts (how it sounds!) on using `ActiveJob`? Do you think it could bring some value over using `DelayedJob` directly?

I was thinking to open a pull request in which `ActiveJob`'s queue adapter is configured to `:delayed_job` but I want to hear some of your opinions first.

ventsislaf:
This removes the step to manually configure the `ActiveJob`'s adapter.

Closes #520


mcmire:
This seems good to me, anyone have any thoughts?

tute:
:+1: from me

tute:
@mcmire do we have any delayed job specifics in other parts of the project? See Derek's comment: https://github.com/thoughtbot/suspenders/issues/520#issuecomment-74528389

mcmire:
That's a good question. I don't know as I haven't studied the Suspenders code in a long time.

tute:
Looping in @derekprior for that. Next week I'll test and merge!

ventsislaf:
Thank you for the fast feedback guys.

What about this piece of code? https://github.com/thoughtbot/suspenders/blob/master/templates/background_jobs_rspec.rb
Do you think we could make this not adapter specific? Also I was wondering whether to set the `ActiveJob`'s adapter to `:test` in the test environment.
http://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/TestAdapter.html

 I think I need to look more into `ActiveJob`.

ventsislaf:
Another option would be to use the [inline adapter](http://edgeapi.rubyonrails.org/classes/ActiveJob/QueueAdapters/InlineAdapter.html) when running features which will execute the jobs immediately.

Any thoughts are welcome!

jferris:
> Another option would be to use the inline adapter when running features which will execute the jobs immediately.

I think this would have the same effect as the current `background_jobs_rspec.rb` file. The purpose is just to make it so that background jobs will run during integration tests.

ventsislaf:
> I think this would have the same effect as the current background_jobs_rspec.rb file. The purpose is just to make it so that background jobs will run during integration tests.

I agree. I'm just wondering here... is there any reason to make it only for integration tests and not for the whole test environment?

jferris:
> I agree. I'm just wondering here... is there any reason to make it only for integration tests and not for the whole test environment?

I think the theory is:

In integration tests, we assume that you generally want to test everything working together, so it makes sense to just run the background jobs and treat the fact that they're backgrounded as an implementation detail. When your test is "the credit card is charged," you don't care if that happens in an HTTP request or a background job.

In unit tests, we assume that you're testing a specific class, and any interactions with other classes are managed by the test. That means that you probably don't need the actual background job to run; you just need to verify that it did run. We usually use mocks to verify this.

In reality, tests tend to be at least somewhat of a blend of both, but this default approach to testing background jobs seems to be reasonably intuitive and has worked out well for us so far.

ventsislaf:
> In unit tests, we assume that you're testing a specific class, and any interactions with other classes are managed by the test. That means that you probably don't need the actual background job to run; you just need to verify that it did run. We usually use mocks to verify this.

I know. That's why I asked whether is worthy to set this only for the integration tests. We would anyway mock the background jobs in the unit tests and I couldn't think of a reason not to set the `inline` adapter in the `config/test/environment.rb`.

Don't take me wrong. I don't mind just to change the existing test helper if that's what we want.

jferris:
> I know. That's why I asked whether is worthy to set this only for the integration tests.

Ah, I see. I misunderstood your original question.

I'd be open to trying `:inline` globally in tests to see how that works out. It's simpler and I can't think of situations where it would be likely to cause problems.

ventsislaf:
Any thoughts?

jferris:
This looks good to me.

@tute any thoughts?

lib/suspenders/app_builder.rb
```diff
@@ -237,6 +236,13 @@ def configure_action_mailer
       action_mailer_host "production", %{ENV.fetch("HOST")}
     end
 
+    def configure_active_job
+      configure_application_file(
+        "config.active_job.queue_adapter = :delayed_job"
+      )
+      configure_environment "test", "config.active_job.queue_adapter = :inline"
```
jessieay: this is cool!

jessieay:
This looks good to me! I didn't know about `ActiveJob` so thanks for the educational PR!

ventsislaf:
@jessieay I'm glad that you find my PR is useful.

tute:
Merged in https://github.com/thoughtbot/suspenders/commit/fce1401954d1c2e175ef8a855b87e516475c63ee. Thank you! :)

PR #529, opened 2015-03-02
Add Autoprefixer

tysongach:
Should we add a default config file, as well? It would allow us to [define which browser versions](https://github.com/ai/autoprefixer-rails#ruby-on-rails) we want to support…where would I put that `browserslist` here in Suspenders?

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -388,6 +388,7 @@ you can deploy to staging and production with:
     end
 
     def copy_miscellaneous_files
+      copy_file "browserslist", "browserslist"
       copy_file "errors.rb", "config/initializers/errors.rb"
       copy_file "json_encoding.rb", "config/initializers/json_encoding.rb"
     end
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 ruby "<%= Suspenders::RUBY_VERSION %>"
 
 gem "airbrake"
+gem "autoprefixer-rails"
 gem "bourbon", "~> 4.2.0"
 gem "coffee-rails", "~> 4.1.0"
 gem "delayed_job_active_record"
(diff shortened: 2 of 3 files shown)


tute:
@wardpenney have you worked with autoprefixer before?

tysongach:
@tute Autoprefixer Rails simply has you put a config file called `browserslist` in the app’s root. In that, you can define the browsers you want to support.

With Suspenders, where can we put that `browserslist` file and have it placed in the correct spot when running a new `suspenders` command?

tute:
@tysongach see how we use `copy_file` and the `templates/` folder: https://github.com/thoughtbot/suspenders/blob/master/lib/suspenders/app_builder.rb#L35-L37 to add our `browserlist` version.

Does thoughtbot use this in the majority of the projects? Did autoprefixer go through the experiments Trello board?

tysongach:
Thanks, @tute, I have a look and add to this PR!

And yes, we’ve now begun adding Autoprefixer to all projects. The next version of Bourbon (v5.0) will remove its prefixing mixins, in favor of Autoprefixer doing that. This is why I want to add it here.

We have a research card already: https://trello.com/c/udp7Bp3i/397-autoprefixer-for-prefixing

tute:
ping @tysongach :)

tysongach:
I just added a commit to add a `browserslist` config file for Autoprefixer. Per our [Playbook](http://playbook.thoughtbot.com/#web-apps):

> We support Internet Explorer 9.0+ and the latest versions of Firefox, Chrome, and Safari. (…)

> On mobile devices, we support iOS Safari 7.1+, Android Browser 4.4+, and the latest version of Chrome for Android.

This config file tells Autoprefixer to support:

- Android: 37, 4.4, 4.4.3-4.4.4
- Chrome: 40, 39
- Firefox: 35, 34
- IE: 11, 10, 9
- IE Mobile: 11, 10
- iOS: 8.1, 8, 7.0-7.1
- Opera: 27, 26
- Safari: 8, 7.1

cc: @tute (appreciate the help!)

tute:
Looking good!

We'll need a test and add an `invoke :setup_autoprefixer` in `lib/suspenders/generators/app_generator.rb`; tomorrow we'll wrap it up.

PR #532, opened 2015-03-05
Remove SMTP assumption

andyw8:
The production config assumes that the SMTP environment variables will be present, but many apps don't need to send mail.

I'd welcome any suggestions for a more elegant way of handling this. There are four env vars related to SMTP, so `SMTP_ADDRESS` is an arbitrary choice.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -66,12 +66,14 @@ module Suspenders
       copy_file 'smtp.rb', 'config/smtp.rb'
 
       prepend_file 'config/environments/production.rb',
-        %{require Rails.root.join("config/smtp")\n}
+        %{require Rails.root.join("config/smtp") if ENV["SMTP_ADDRESS"].present?\n}
 
       config = <<-RUBY
 
-  config.action_mailer.delivery_method = :smtp
-  config.action_mailer.smtp_settings = SMTP_SETTINGS
+  if ENV["SMTP_ADDRESS"].present?
+    config.action_mailer.delivery_method = :smtp
+    config.action_mailer.smtp_settings = SMTP_SETTINGS
+  end
       RUBY
 
       inject_into_file 'config/environments/production.rb', config,


jessieay:
One tricky think about this PR is that someone generating an app with suspenders is starting from scratch, so there would be no `ENV` vars present while `app_builder.rb` is running. Which means that, if we incorporated this PR, the SMTP settings would never be included.

I agree with the point that there are many apps that don't do email, so being automatically opted into these configs isn't an ideal situation. 

See the conversation [here](https://github.com/thoughtbot/suspenders/pull/511) about turning Suspenders into more of an opt-in template (rather than opt-out). I think this PR is a good example of why going that direction would be helpful...but I am not sold on going that direction just yet.

andyw8:
The conditional is part of the generated output, so it doesn't matter if the `ENV` var is present while the app builder is running, only that it's present or absent when the built app is started.

tute:
Agreed with @jessieay that this is a perfect fit for an opt-in option, if we had such an architecture.

I guess the question is to thoughtbot developers: how many times do we need to strip out this config from generated apps vs how many add it in?

If it's even, the conditional makes sense to me.

gabebw:
I routinely strip it out. Every app I've built over the past 6 months, I've removed SMTP settings in the first week.

BlakeWilliams:
I've also removed it from every app I've started.

tute:
If we use it for some projects, I'm not against the conditional. I never had to use them either though.

@croaky introduced the `SMTP` behavior in https://github.com/thoughtbot/suspenders/commit/44c01dbc. Dan, did you use this configurations in a few projects?

JoelQ:
I've found myself removing this from all my suspenders-generated procjects as well

tute:
We just had a discussion on this, and decided to leave the SMTP assumption there. Some points of the discussion:

* Many people remove/comment out/disable SMTP config in the first day of a project.
* Others deploy to Heroku on day one, and set all the services. If they add user authentication on day one, they'll typically need to set up SMTP on day one too.
* Sooner or later most projects _do_ use SMTP.
* `suspenders` documents the correct configuration, and SMTP configuration isn't completely straightforward.
* `suspenders new` is not "plug and play": some changes are often needed right after generation of the app (we minimize those, but don't completely remove that need). Users can comment out this block, or remove it in its own commit and revert it when it's needed.

So, thank you all for your feedback and work, closing this PR for now!

PR #535, opened 2015-03-09
Remove New Relic memory management variable

reshleman:
The `NEW_RELIC_AGGRESSIVE_KEEPALIVE` environment variable was was added
in #440 to reduce New Relic's memory consumption for idling applications
on Heroku. As of version 3.9.8, New Relic defaults this setting to
`true`, so it's no longer necessary to set it explicitly.

From the New Relic [CHANGELOG]:

> Reduced memory usage for idling applications
>
> Idling applications using the agent could previously appear to leak
> memory because of native allocations during creation of new SSL
> connections to our servers. These native allocations didn't factor
> into triggering Ruby's garbage collector.
>
> The agent will now re-use a single TCP connection to our servers for
> as long as possible, resulting in improved memory usage for
> applications that are idling and not having GC triggered for other
> reasons.

Related commit: newrelic/rpm@e04d5cedf17bc1df032eabf56529ed4d4a0ddcb8

[CHANGELOG]: https://github.com/newrelic/rpm/blob/master/CHANGELOG


tute:
:+1:

tute:
Thank you!

PR #536, opened 2015-03-09
Enable verify_partial_doubles for RSpec

andyw8:
https://relishapp.com/rspec/rspec-mocks/v/3-0/docs/verifying-doubles/partial-doubles

"When the verify_partial_doubles configuration option is set, the same
argument and method existence checks that are performed for
object_double are also performed on partial doubles. You should set
this unless you have a good reason not to. It defaults to off only for
backwards compatibility."


mcmire:
This looks good to me. :+1:

tute:
Thank you!

PR #538, opened 2015-03-12
Mention coaching

jferris:


tute:
Seems great.

halogenandtoast:
Is this really the correct venue for this? Do we have examples of others taking this approach. Even if not it's fine to be innovative but this seems a little desperate.

jferris:
> I'm interested in the motivation behind this approach.

I think this fills a niche and will help scratch some itches. We do our best to debug things in GitHub issues and provide support over mailing lists, Stack Overflow, and other venues. However, I can't always offer several hours of time for free, and I'm sure others are in the same situation. If we can get people to pay for that support, we can put more people in a position to get the most out of free software.

> Do we have examples of others taking this approach?

There are numerous examples of open source projects which use paid support, commercial licenses, or paid-only features, such as Sidekiq, MySQL, and Qt. I'm not sure I've seen the exact coaching approach, but I think it fits well with our general process and the consultant mindset.

> Is this really the correct venue for this?

Oops, no! I meant to submit this as pull request to Clearance. Excellent question.

jferris:
https://github.com/thoughtbot/clearance/pull/540

tute:
It was surprising we'd talk about auth only, but it makes sense too, as suspenders uses those gems. We just need to update a bit the copy here, I'm in favor of this PR.

halogenandtoast:
I do think it makes sense to some degree. Unlike other venues for communicating this information this is a bit more permanent. I also think it makes sense in context with this gem, just phrased differently.

jferris:
https://github.com/thoughtbot/suspenders/pull/539

PR #542, opened 2015-03-16
Only use rack-timeout in staging or production environments

jakecraige:
This causes intermittent failures in capybara test suites where js is enabled.

See https://github.com/heroku/rack-timeout/issues/55 for more information.

### DIFF
--- a/templates/rack_timeout.rb
+++ b/templates/rack_timeout.rb
@@ -1,1 +1,0 @@
-Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -222,7 +222,11 @@ end
     end
 
     def configure_rack_timeout
-      copy_file 'rack_timeout.rb', 'config/initializers/rack_timeout.rb'
+      rack_timeout_config = <<-RUBY
+Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
+      RUBY
+
+      append_file "config/environments/production.rb", rack_timeout_config
     end
 
     def configure_simple_form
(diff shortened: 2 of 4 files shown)


reshleman:
FWIW, I had been experiencing an issue like the one linked (intermittent failures on Capybara feature specs w/ `js: true`) on a Suspenders-generated app. Making a change similar to the one in this PR -- at least temporarily, until the linked issue is resolved -- has resulted in 0 non-deterministic spec failures in dozens of runs. Previously, with `rack-timeout` loaded, the test suite was timing out > 50% of the time on CI.

jakecraige:
That's pretty much what we were experiencing. In our case it would *almost* always fail when running the whole suite, but running individual specs or files was *usually* fine. 

There seems to be open issues on pretty much all the drivers about something similar. This is the one we finally tracked down in our case and it would be great to add this so it doesn't cause others the same headache.

mcmire:
This is interesting. I wouldn't think that a timeout of 10 seconds would cause any tests to fail, so that seems weird to me. But, there shouldn't be any timeout in tests anyway, it's really not that important. So I'm :+1:  on this.

tute:
This file was introduced in https://github.com/thoughtbot/suspenders/commit/3f444c82a265865a3294a90d5e6bde86234aa0d8 by @croaky. Would like him to peek this, but I feel positive about this change too.

templates/rack_timeout.rb
```diff
@@ -1 +1,3 @@
-Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
+if Rails.env.production? || Rails.env.staging?
```
croaky: Rather than conditionally checking the environment here, would it be better to put this inside `config/environments/production.rb`? We `require_relative "production" in `config/environments/staging.rb`.

tute: :+1: to this comment. @jakecraige can you please add that change, and rebase to latest master?

jakecraige: Yeah that sounds good. I made the change.

tute:
Thank you!

Issue #537, opened 2015-03-10
Refills flashes


tysongach:
There seems to be two sets of `_flashes` partials in `views`: one in `views/application` and the other in `views/refills`.

I believe the `views/refills` partial can be removed (along with the whole `views/refills` directory) and Refills should stop generating that, if it is.

tute:
Can you verify this @tysongach please? I never worked with Rails + Bourbon yet (I'll suspenderize my "breakable toys"), and thus you will have a more educated view of it.
Thank you!

tute:
I see what you mean @tysongach. It seems it's generated somewhere, it shouldn't.

tute:
Closing in favor of PR. Thanks Tyson!

PR #546, opened 2015-03-25
Avoid generation of extra _flashes view

tute:
This step generates an unneeded `app/views/refills/_flashes.html.erb`, but there's already `app/views/application/_flashes.html.erb`. Contents of the seemingly unneeded component:

```erb
<div class="flash-success">
  <span>This is a success message <a href="#">with a link</a></span>
</div>

<div class="flash-error">
  <span>This is an error message <a href="#">with a link</a></span>
</div>

<div class="flash-notice">
  <span>This is an notice message <a href="#">with a link</a></span>
</div>

<div class="flash-alert">
  <span>This is an alert message <a href="#">with a link</a></span>
</div>
```

Related: https://github.com/thoughtbot/suspenders/issues/537

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -274,10 +274,6 @@ end
       run "bitters install --path app/assets/stylesheets"
     end
 
-    def install_refills
-      run "rails generate refills:import flashes"
-    end
-
     def gitignore_files
       remove_file '.gitignore'
       copy_file 'suspenders_gitignore', '.gitignore'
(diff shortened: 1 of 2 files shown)


tute:
@croaky is there a reason why you added this component in https://github.com/thoughtbot/suspenders/commit/6988eeba? Should it still be there?

croaky:
Our goal should be that the flashes in Suspenders out of the box use the flash styles in Refills. div, class name, span, a tags, etc. If it's broken or duplicated now, definitely change it to not run the generator but we should also make sure that `app/views/application/_flashes.html.erb`'s markup works with Refills' styles.

tute:
It does, thanks Dan!

Issue #547, opened 2015-04-06
Nitrous.io


lawgixlaw:
Got an error that suspenders requires Ruby 2.2.1.  That's the version that's installed and running in the Nitrous box by default.  Any ideas?

mrlee-io:
A nitrous.io box, by default, is running ruby 2.1.1p76 ( I just started a new rails box ).

In order to install a different version you use what they call 'parts'.

```
parts search ruby
parts install xyz
```
Restart the console and you should be good to go with whatever version you installed.

note 1:  They do not currently have a 'part' for ruby 2.2.1
note 2:  You can also use the GUI to manage packages under 'autoparts'.
note 3:  If you can't wait for nitrous to release 2.2.1 you can try downloading suspenders and building it manually requiring only 2.2.0 (no idea if it will build)

tute:
Suspenders should be more permissive with it's ruby version for generating apps I think. A user should be able to use "good" Ruby versions without the need for services to be all running latest Ruby (while still generating apps that require the latest version available, as a good default).

lawgixlaw:
I'm a beginner at this.   What you're saying is that I can't use suspenders on Nitrous.io until they launch Ruby 2.2.1, correct?  

I work on a Chromebook and am really liking the idea of developing completely in the cloud between Nitrous and Heroku.   Suspenders looks like a good way for me to build apps by piecing together code parts that others have written for much of the basic stuff like user accounts, etc.   

Am I right in that assumption?   Is there another gem or tool that would do the same thing as suspenders?

tute:
If @jferris doesn't disagree, I'll relax the Ruby Version needed for suspenders gem to >= 2.1.

tute:
Closing in favor of PR. Thanks for your input, I understand your issue.

PR #548, opened 2015-04-08
Relax ruby version for generator

tute:
suspenders recommends and will always generate apps for latest Ruby
version. But the generator doesn't need to be too picky about which Ruby
is it being run on.

See an example of why this is more convenient:
https://github.com/thoughtbot/suspenders/issues/547


mcmire:
I disagree with this change. I think that it makes sense that Suspenders uses the latest version of Ruby. What if we provided an an option to change that version instead?

jferris:
I think we want to make sure that the `suspenders` command runs using the same version of Ruby that we're going to run on, because otherwise it will run `bundle install` for the wrong version.

Also, in the case of #547, it seems like this won't really help, as the app will still refuse to run except on the latest version of Ruby; it will just let you generate an app that you can't run.

tute:
I guess I feel the pain because last year there were 6 minor or patch Ruby releases, and I spent some time bundling apps for different rubies unnecessarily for development, as these apps run in any Ruby >= 2.0 (sometimes >= 2.1).

My opinion is that most times this error message is not worth it.

```
% suspenders new app
rbenv: suspenders: command not found

The `suspenders' command exists in these Ruby versions:
  2.1.1
  2.1.3
  2.1.4
  2.1.5
  2.2.0

fatal
```

While I prefer running latest and most secure software, I think it's not _always_ needed. 2.1.5 is all right. And while thoughtbot doesn't need to care, I don't think latest Ruby should be enforced to the point where it's not even an option in some platforms.

Generating a suspenders app that doesn't run in your Ruby version allows you to update the generated `.ruby-version` accordingly and then run it. But it's not possible to run suspenders if you don't have that version unless you build suspenders locally.

An option for defining the Ruby version would probably be the best solution for this.

jferris:
I think it's unfortunate that we need to compile Ruby every time they release a new version, but I also feel that, as Ruby developers, we should be running the latest stable release.

The fact that the suspenders command is distributed as a Ruby gem combined with the fact that we using rbenv creates the unfortunate "command exists in these Ruby versions" issue. I would be open to exploring other avenues of distributing suspenders.

tute:
Thanks Elliot and Joe for your input.

I still think that relaxing the Ruby version the suspenders command requires to run is not against our goal of working on latest versions of everything (generated app requires it), with the additional option of using another one if we ever choose to.

mcmire:
Ohh I see... this is just the minimum version in the gemspec.

Then, yes, I approve of this, that seems reasonable to me.

jferris:
What would the actual effect of loosening the gem version be? We're still going to write the same version into the generated `Gemfile`, right? Have you test this out locally, using a different installed version of Ruby than the one that gets added to the project?

tute:
> We're still going to write the same version into the generated `Gemfile`, right? Have you test this out locally, using a different installed version of Ruby than the one that gets added to the project?

Whoops, it fails on the `bundle install` step (`Your Ruby version is 2.1.5, but your Gemfile specified 2.2.2`), and in follow-up steps. For the generator to work, we either need the latest ruby version (as we enforce in latest master), or we need to allow the generation of apps for older rubies.

Can't think now of a way this PR can solve this issue.

PR #552, opened 2015-04-14
We still need to import refills scss

tabfugnic:
* refills/flashes does not exist and fails import
* As per issue #537 and PR #546

I recognize that this is not the ultimate solution, this is still broken but this could be a possible solution and would have to change the refills gem.

We may want to consider scrapping the current flashes partial in favor of the refills defined ones instead.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -293,6 +293,12 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
                 "app/assets/stylesheets/application.scss"
     end
 
+    def install_refills
+      run "rails generate refills:import flashes"
+      run "rm app/views/refills/_flashes.html.erb"
+      run "rmdir app/views/refills"
+    end
+
     def install_bitters
       run "bitters install --path app/assets/stylesheets"
     end
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -40,6 +40,7 @@ module Suspenders
       invoke :configure_app
       invoke :setup_stylesheets
       invoke :install_bitters
+      invoke :install_refills
       invoke :copy_miscellaneous_files
       invoke :customize_error_pages
       invoke :remove_routes_comment_lines
@@ -149,6 +150,11 @@ module Suspenders
     def install_bitters
       say 'Install Bitters'
       build :install_bitters
+    end
+
+    def install_refills
+      say "Install Refills"
+      build :install_refills
     end
 
     def setup_git


tute:
> We may want to consider scrapping the current flashes partial in favor of the refills defined ones instead.

Do you think this would be cleaner? My goal while removing it on https://github.com/thoughtbot/suspenders/pull/546 was to avoid the duplication and unused partial, we can as well remove the other one and reference the refills one.

Question: how does it fail when it's not there? I didn't see it fail before, and that's why I removed it.

Thank you @tabfugnic!

tabfugnic:
This creates an immediate 500 error for the first page that you create because refills/flashes does not exist.

Steps to reproduce:

1. Create new suspenders app
2. Run `bin/setup`
3. Create new page, for example a new high voltage page
4. start server
5. navigate to that page

pedrosmmoreira:
I can confirm the issue and reproduction steps, thanks for this @tabfugnic!

lib/suspenders/app_builder.rb
```diff
@@ -275,6 +275,10 @@ def setup_stylesheets
         'app/assets/stylesheets/application.css.scss'
     end
 
+    def install_refills
+      run "rails generate refills:import flashes --only-stylesheets"
```
tute: Eric added `--only-stylesheets` but that's not an option to refills.

I think it does make sense for refills to generate markup and CSS, as they are coupled by definition, but we generate our own version of the markup in suspenders. @tabfugnic would you provide a step here to remove the refills-generated flashes partial?

Thank you!

mcmire:
@tute @tabfugnic Hey, what's the status of this PR? Do we think we can get this merged in soon?

tute:
@mcmire I just pushed a bugfix, which runs `run "rm app/views/refills/_flashes.html.erb"`. We don't want the view, but we do want the refills CSS. Can you please review this, and @tysongach too?

Thank you!

lib/suspenders/app_builder.rb
```diff
@@ -289,6 +289,11 @@ def setup_stylesheets
                 "app/assets/stylesheets/application.scss"
     end
 
+    def install_refills
+      run "rails generate refills:import flashes"
+      run "rm app/views/refills/_flashes.html.erb"
```
tute: We could remove the directory too.

tysongach:
:+1: This looks good and I think it’s nice quick win.

It does create a new issue—for me, at least—which is Refills making a `refills` directory inside `stylesheets`. I’d much prefer that directory be called `components`, because `refills` doesn’t _mean_ anything, but a flash message _is_ a component. This is more of an issue for Refills, though and not Suspenders.

tute:
Indeed we should fix it in refills I think, so it's clear what it is right now. Will merge on Friday unless someone finds a better way of fixing this.

mcmire:
@tute Can we merge this in?

tute:
Thanks for poking @mcmire! Will release soon.

Issue #220, opened 2013-08-21
Set up staging and production for Cloudfront CDN


croaky:
`config/environments/{staging,production}.rb`:

``` ruby
config.action_controller.asset_host = ENV['ASSET_HOST']
config.assets.version = ENV['ASSETS_VERSION']
config.static_cache_control = "public, max-age=#{1.year.to_i}"
```
## ASSET_HOST

The value of `ASSET_HOST` will look like `//123abc.cloudfront.net`
## ASSETS_VERSION

In production every asset has a hash added to its name. Whenever the file changes, the browser requests the latest version as the hash and therefore the whole filename changes.

Putting the assets version in an environment variable makes it possible to invalidate the cache without a re-deploy, if necessary. Simply re-set the `ASSETS_VERSION` as a Heroku config variable to change the hash globally and make the browser request every asset again.
## Already in Suspenders

Stuff we already have in Suspenders that makes this all work:

`Gemfile`:

``` ruby
gem 'coffee-rails'
gem 'sass-rails'
gem 'uglifier'
```

`config/environments/{staging,production}.rb`:

``` ruby
config.assets.compile = false
config.assets.digest = true
config.assets.js_compressor = :uglifier
```

We don't have to set `config.serve_static_assets = true` because the `rails_12factor` gem does it for us.
## Cloudfront settings

We probably want the following settings in Cloudfront:
- "Download" Cloudfront distribution
- "Origin Domain Name" as www.example.com (our app's URL)
- "Origin Protocol Policy" to "Match VIewer"
- "Object Caching" to "Use Origin Cache Headers"
- "Forward Query Strings" to "No (Improves Caching)"
- "Distribution State" to "Enabled"

PR #553, opened 2015-04-17
All used ENV variables with sample value in `.sample.env`

Dorian:
Easier discovery of ENV variables


tute:
I wouldn't add variables with sensible defaults which are not so eager to change:

* `DB_POOL`
* `ASSETS_VERSION`
* `DB_REAPING_FREQUENCY`

Where are the following vars used? I don't think I'd add them.

* `TIMEOUT_IN_SECONDS`
* `WEB_CONCURRENCY`
* `WEB_TIMEOUT`

The change for `SECRET_KEY_BASE` is not necessary.

`SEGMENT_IO_KEY` should read `SEGMENT_KEY`.

I would leave `SMTP_*` empty, or commented out, so we force real values. See related discussion in https://github.com/thoughtbot/suspenders/pull/532#issuecomment-86976291.

Thanks for your work!

tute:
ping @Dorian

Dorian:
Variables were outdated because I used an older version of suspenders.

> I would leave SMTP_* empty, or commented out, so we force real values. See related discussion in
> #532 

Maybe there is a nicer way to introduce the `SMTP_*` than by throwing an exception when deploying on a production env :smile:, but I understand.

(agree on the rest)
(Understood `ASSETS_VERSION` with #220 (interesting))

Thanks for your work, learned a lot from suspenders.

tute:
I'm glad it's useful for you, thanks for your input!

PR #561, opened 2015-04-30
Swap flash id for a class

tysongach:
Generally speaking, it’s best practice to style off `class`’s instead of `id`’s. There are exceptions, but in this case, I would consider the `id` of `flash` not very useful, so this swaps it for a `class` instead. Also, the flash key on Line 4 was already being output as a class; this unifies the two.

### DIFF
--- a/templates/_flashes.html.erb
+++ b/templates/_flashes.html.erb
@@ -1,5 +1,5 @@
 <% if flash.any? %>
-  <div id="flash">
+  <div class="flashes">
     <% user_facing_flashes.each do |key, value| -%>
       <div class="flash-<%= key %>"><%= value %></div>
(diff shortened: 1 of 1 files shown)


tute:
Looks good! Can you please add context/the PR description into the commit message before merging?
Thank you!

andyw8:
Since this wraps a collection (`user_facing_flashes`), would `flashes` be more descriptive?

tysongach:
@andyw8 Great point. `flashes` makes more sense when multiple flashes are being shown. Updated.

tute:
If we don't need to update any CSS or JS for this, feel free to rebase and merge @tysongach! Thank you.

PR #563, opened 2015-05-01
Ignore `.env.local` instead of `.env`

lpender:
Here is an common state that the current `.gitignore` supports:

- The `.sample.env` (or like) file is updated with a set of environment variables after a feature is merged.
- A developer pulls the latest from master and runs the test suite.
- It fails with an opaque error message.

A more robust workflow:

- Version development environment variables in the `.env` file.
- Overrides these locally with the `.env.local` file..

See:

https://github.com/bkeepers/dotenv#should-i-commit-my-env-file
https://github.com/thoughtbot/dotfiles/pull/369

### DIFF
--- a/templates/suspenders_gitignore
+++ b/templates/suspenders_gitignore
@@ -3,7 +3,7 @@
 *.swo
 *.swp
 /.bundle
-/.env
+/.env.local
 /.foreman
 /coverage/*
 /db/*.sqlite3
@@ -12,3 +12,4 @@
 /public/assets
 /tags
 /tmp/*
+!.env


lpender:
Been using this for the last two weeks at SkilledUp Academy without issue.

mcmire:
I do think this is a good idea in theory, although it does add one more file that you have to maintain.

templates/suspenders_gitignore
```diff
@@ -12,3 +12,4 @@
 /public/assets
 /tags
 /tmp/*
+!.env
```
tute: Why would we need this line?

tute:
We'll hold off on this PR until we try in projects at thoughtbot. Thanks Lee!

templates/suspenders_gitignore
lpender: @tute That line is to override a `.env` ignore in your dotfiles.

See https://github.com/thoughtbot/dotfiles/pull/369.

Sounds good, thanks Tute!

mcmire:
I used this on my last project and found this to be a good change. I think this is something that makes sense.

templates/suspenders_gitignore
gylaz: We should remove this line, and the open a PR in our dotfiles **not** to ignore this file.

gylaz:
I'm a fan of this approach. :+1:

templates/suspenders_gitignore
mcmire: Oh, good call.

tute:
I'm :+1: on this too. @lpender: want to reopen the dotfiles PR, and remove that override from this PR? Then I'll merge. Thank you! :clap:

tute:
Will wait on @lpender for two more weeks, and do it myself if he can't do it. Thank you!

thorncp:
One thing to note is that Foreman [dropped Dotenv](https://github.com/ddollar/foreman/commit/19e9806c4cdf348325fe8312fac827774af4affa) and started parsing .env files itself. Because of this, it does not behave the same with regards to .env.local, .env.development, etc. (it completely ignores them). And because Foreman spawns processes by specifying the environment variables it knows about, they take precedence over those in the dotenv files when the app boots and parses them.

I've tried a couple ways of getting around this:

* Specifying the files in `.foreman`, e.g. `env: .env,.env.local`
* Not letting Foreman set up any env variables, e.g. `env: ""`

I don't know if either of those are ideal, but I think this PR needs to include a solution before merging.

tute:
> those weird capybara-webkit issues I was experiencing on my laptop (secondary machine) were related to differing .env files between my two machines. And because .env isn’t checked into git they weren’t being updated. Let me know if you know a way to share some things with git and other things with one other machine. I’m not aware of an easy solve.

Will merge these PRs this Friday.

tute:
> I don't know if either of those are ideal, but I think this PR needs to include a solution before merging.

@thorncp what is a case where we'd want to use an override like `.env.local`? I didn't need that yet.

Also, if we do, `env: .env,.env.local` in suspenders sound good to me.

Thank you!

thorncp:
@tute A common pattern I've seen is putting placeholder variables for sensitive info in _.env_, then overriding them _.env.local_. The placeholder signifies what variables the app expects to be there. Here's an [example from Hound](https://github.com/thoughtbot/hound/blob/master/.env#L7-L11).

tute:
> A common pattern I've seen is putting placeholder variables for sensitive info in .env, then overriding them in .env.local. The placeholder signifies what variables the app expects to be there. Here's an example from Hound.

Thanks! Then this PR makes sense for that workflow: `.env` has the dummy values, and we add the `env: .env,.env.local` configuration to do the sometimes needed override. Do I follow?

thorncp:
@tute Yes, that's right. Though, I think it's worth pointing out that using `env: .env,.env.local` _won't_ pickup any variables that may be defined in .env.development or .env.test (which dotenv supports). I have not used those files before, so `env: .env,.env.local` has worked for me.

mcmire:
I think that's fine. I've never used `.env.development` and `.env.test`, either -- `.env.local` serves my needs just fine.

tute:
I have not used nor needed those either. Thanks!

tute:
> `.env.local` serves my needs just fine.

And this continues to work. Merging!

tute:
Thank you, Lee! :)

lpender:
Thanks guys! Sorry to just be seeing this now. Glad it got merged :)

PR #564, opened 2015-05-05
Do not touch $HOST in .env

mike-burns:
The .env should not touch $HOST, since that is controlled by the shell
and changing it can lead to surprises, failures, and even crashes.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -113,7 +113,7 @@ module Suspenders
     def setup_asset_host
       replace_in_file 'config/environments/production.rb',
         "# config.action_controller.asset_host = 'http://assets.example.com'",
-        'config.action_controller.asset_host = ENV.fetch("ASSET_HOST", ENV.fetch("HOST"))'
+        'config.action_controller.asset_host = ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))'
 
       replace_in_file 'config/initializers/assets.rb',
         "config.assets.version = '1.0'",
--- a/templates/sample.env
+++ b/templates/sample.env
@@ -1,6 +1,6 @@
 # http://ddollar.github.com/foreman/
 ASSET_HOST=localhost:3000
-HOST=localhost:3000
+APPLICATION_HOST=localhost:3000
 RACK_ENV=development
 SECRET_KEY_BASE=development_secret
 EXECJS_RUNTIME=Node


arunagw:
I think we need to remove this from here as well. https://github.com/thoughtbot/suspenders/blob/599b8932e98f700a7178c3bcbae1a1ec2ce4c42a/lib/suspenders/app_builder.rb#L98

mike-burns:
No that seems correct; it pulls `$ASSET_HOST` from `$HOST`. Or is `$HOST` not set on OS X?

gabebw:
`HOST` is not set on OS X, which is why we manually set (or, in non-OSX's case, override) it.

gabebw:
Is it truly a problem to override `$HOST` in just the foreman process? It's not like we're adding it to `.bashrc`.

tute:
My Mac does set HOST. I personally don't mind it either way in suspenders.

gabebw:
Oh wait, my mac does set HOST:

```
$ echo $HOST
calvin
```

mike-burns:
For those with access to tbot chat: https://thoughtbot.slack.com/archives/dev/p1430792571001997

And in general, `$HOST` is the hostname of the computer. We make use of this in rcm, and I'm sure other projects make use of it too (I don't have access to an OS source tree right now but I can check OpenBSD later).

gylaz:
What should be used instead of `HOST` to set the proper url in Rails apps? Checking a few of the recent apps, I found several instances of `HOST` usage, which would not work with my default mac `HOST`.

mike-burns:
What is your `$HOST` such that it will not work, @gylaz?

gabebw:
Mine is my `hostname`: `calvin`. `http://calvin/` doesn't load up the app running on `localhost:9000`.

mike-burns:
@gabebw does http://calvin:9000/ work?

gabebw:
It does not.

On Mon, May 11, 2015 at 11:20 AM, Mike Burns <notifications@github.com>
wrote:

> @gabebw <https://github.com/gabebw> does http://calvin:9000/ work?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/thoughtbot/suspenders/pull/564#issuecomment-100941624>
> .
>



-- 
Gabe Berke-Williams
http://gabebw.com/

gylaz:
```
~ % echo $HOST
greg-mbp.local
```
The behavior is same as @gabebw.

jferris:
It sounds like what we're learning here is that `$HOST` is not actually a reachable host on OS X by default?

mike-burns:
I've also learned, on GNU/systemd/Linux:

```
~% echo $SHELL
/usr/bin/zsh
~% echo $HOST
autumn
~% echo $HOSTNAME

~% bash
mike@autumn:~$ echo $HOST

mike@autumn:~$ echo $HOSTNAME
autumn
mike@autumn:~$ 
```

This all makes me think even more that 1) touching `$HOST` is potentially dangerous and going to bite us in surprising ways, and 2) we should not reference `$HOST` or `$HOSTNAME` from our Ruby.

jferris:
> We should not reference `$HOST` or `$HOSTNAME` from our Ruby.

We generally set and reference `HOST` in the Heroku environment, and I think it's useful; otherwise, we need to commit the host names to Git. I'd be open to renaming `HOST` to something else, though.

mike-burns:
I've changed the name of the env var.

mike-burns:
Anyone opposed?

reshleman:
Would naming this var `RAILS_HOST` or `APPLICATION_HOST` make it clearer what its purpose is?

It also looks like are a few other references to `HOST` that will need to be updated:

* https://github.com/thoughtbot/suspenders/blob/13f9450ceab5a41319dc523254d6652c963d0c7e/lib/suspenders/app_builder.rb#L85
* https://github.com/thoughtbot/suspenders/blob/13f9450ceab5a41319dc523254d6652c963d0c7e/lib/suspenders/app_builder.rb#L254-255

mcmire:
Other names for `THIS_HOST`: `LOCAL_HOST`? Or would that be easily confused with the idea of localhost?

I like APPLICATION_HOST.

mcmire:
Actually, why do we even have a separate environment variable here? Isn't `ASSET_HOST` good enough?

gabebw:
We might store assets on another host, like a CDN. I like `APPLICATION_HOST`.

mcmire:
Yeah but if we store assets on another host... then set `ASSET_HOST` to that host. If we don't, then `ASSET_HOST` is already set so that asset files are served from localhost. This seems to be the only place where we'd use a separate `APPLICATION_HOST` (or whatever).

jferris:
The current `HOST` is used by mailers, background jobs, and other classes where we might want to generate a URL without an incoming request. We also need a hostname in the config, because we [redirect all traffic to the canonical host](https://github.com/thoughtbot/suspenders/blob/a29f2a63836cc38d25ad1161768a0a87e62b26e5/lib/suspenders/app_builder.rb#L89).

mike-burns:
Aside from bikesheds about the name of this variable, is this idea good to merge? Alternatively we can close this PR.

mcmire:
Alright. Yes I think it's a good idea to merge.

tute:
`THIS_HOST` feels awkward to me, but I can't propose a better name. Merge at will @mike-burns.

jferris:
I personally find `THIS_HOST` odd enough that I'd prefer not to merge with that name. It's like having "post" and "do_post" as method names.

Is there any objection to `APPLICATION_HOST`?

gabebw:
I like `APPLICATION_HOST` much better.

gylaz:
Would `APP_HOST` be a decent, shorter alternative?

gabebw:
Since this isn't something we'd write a lot (we'd set it once per app), and will be something we read a lot, I'd prefer something longer and more explicative.

PR #565, opened 2015-05-06
Use rack-canonical-host

jferris:
Redirects web traffic so that all requests are served from the canonical
hostname.


tute:
I like this! :+1:

calebthompson:
I don't think I've ever run into a case where this wasn't happening / where I needed this behavior. What's the benefit to most apps that this has?

jferris:
> What's the benefit to most apps that this has?

The case I see on most apps is redirecting `www.example.com` to `example.com`. I think it's good practice to register and redirect `www`, because people are used to typing it, and some browsers even have shortcuts for adding it to a domain.

calebthompson:
Ah, good point. Isn't that usually handled with a [CNAME record](http://en.wikipedia.org/wiki/CNAME_record), and more a DNS concern than an app concern?

jferris:
> Ah, good point. Isn't that usually handled with a CNAME record, and more a DNS concern than an app concern?

If you use a CNAME, the browser uses that as the host name in HTTP requests, which means relative URLs, cookies, and other things will be based on the non-canonical domain. This leads to every resource on your site having two public URLs: one with `www`, and one without. It can also cause issues when users come back to a different domain, because cookies will be different, and tools like 1Password won't always know it's the same site.

Some DNS providers like dnsimple have URL-redirect support, but they can't redirect https traffic, as they don't have certificates for your `www.` domain. This means you need to serve the redirect yourself if you want it served over https.

As long as TLS and DNS are completely separate, there's no way to handle this kind of thing in DNS.

mjankowski:
You need to do both the DNS side and the application side.

Without the canonical host in place on the app side, you wind up with a scenario where all of:

* example.com
* www.example.com
* example-com-production.herokuapp.com

are valid hostnames for the application, are all potentially sent in links places, and are all potentially indexed by google.

This changes forces all of that to one spot.

:deciduous_tree:

calebthompson:
> If you use a CNAME, the browser uses that as the host name in HTTP requests, which means relative URLs, cookies, and other things will be based on the non-canonical domain.

TIL, thanks!

> As long as TLS and DNS are completely separate, there's no way to handle this kind of thing in DNS.

Cool.

> :deciduous_tree:

Agreed.

derekprior:
Having a canonical host everything is directed to can also help avoid weird issues where google page rank decreases because it sees duplicate content. I support the idea of a canonical host.

> The case I see on most apps is redirecting www.example.com to example.com

I've banged this drum a few times and no one seems to care, but what's one more more? I'd strongly suggest redirecting your zone apex to `www` rather than the other way around. Per DNS spec, you cannot have a `CNAME` at your zone apex, but `CNAME`ing is the most common way to put something behind a CDN. 

We get around this by using DNSimple, which has an `ALIAS` record that effectively simulates a `CNAME` at the apex. This isn't standard though, and most providers do not offer it. Providers that do, often offer it as an `ANAME` instead -- also non-standard. By relying on one of these record types, we are locking ourselves in to these providers and more importantly making it more difficult to be fault tolerant.

DNSimple went down hard last year and took all of our sites and customers with them. Had we been using standard DNS servers it would have been a best practice to have a backup provider listed as one of our name servers. Backup providers can sync your DNS records from your primary provider via zone transfer (`AXFR`). But the proprietary nature of DNSimple's `ALIAS` was at least in-part responsible for this not being possible with DNSimple. Even if we had access to our full zone in record, uploading to another provider would have meant temporarily compromising the `ALIAS` feature.

DNSimple has since worked out secondary DNS offerings provided by a limited number of providers. If those providers don't support `ALIAS` then DNSimple pushes the IP that `ALIAS` resolves to as an `A` record. So you're again throwing out the `ALIAS` functionality.

Most of this should "just work" now that they've put some effort behind it, but it's still a lot of magic for something that can just be avoided by forgoing the vanity of an apex domain. Yes, it looks prettier, but I don't think it's worth all of this hullabaloo. I've learned to stop worrying and love my `www`.

jferris:
@derekprior I think there are a lot of good points there, and I'd be interested in having that discussion. There's nothing in rack-canonical-host or other code in this pull request which will remove `www`, and I'm not sure how many people follow Suspenders, so I'm not sure this is the best place to have this discussion.

I think you may have success by:

* Opening a pull request to the guides which recommends against using non-standard DNS entries
* Opening a pull request to the handbook which recommends using `www` on Heroku

I believe that Heroku agrees with you. They seem to reluctantly acknowledge the existence of `ALIAS` and `ANAME` in their documentation, but they've historically recommended that people use subdomains.

PR #566, opened 2015-05-07
Require spec/support files in alphabetical order

gabebw:
This ensures that the `require` order is the same on all machines, including CI.

For example, we mainly use OS X as our local development machines, while CI runs
on Linux. We've run into ordering issues before due to the different OS's, and
this changes has historically fixed it.

### DIFF
--- a/templates/rails_helper.rb
+++ b/templates/rails_helper.rb
@@ -5,7 +5,7 @@ require File.expand_path("../../config/environment", __FILE__)
 require "rspec/rails"
 require "shoulda/matchers"
 
-Dir[Rails.root.join("spec/support/**/*.rb")].each { |file| require file }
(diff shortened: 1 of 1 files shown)


BlakeWilliams:
:+1:

tute:
:+1:

gabebw:
@tute Are you the leader of suspenders now? Is your thumbs up the final thumbs up?

tute:
Yes Gabe. In my non-thoughtbot OSS I rebase/merge them myself (should I here, or should I let you do it?). And, while I'm responsible for suspenders, we have the Experiments Trello board that Joe keeps in mind and I don't visit often, so my sign off here would be enough I'd say for this PR, but for others I ask for his thumbs up too, so we make sure it's the right step forward before merging.

Feel free to merge this!

gabebw:
I'll merge this RIGHT NOW.

PR #567, opened 2015-05-19
Fix bundler caching on travis to speed up builds

iwz:
Travis can't cache bundler if you use a custom install step (which we're
doing when we call bin/setup). This specifically bundles into a
directory that will then be cached on Travis and used for each build.

* bundle into `vendor/bundle`
* Add `vendor/bundle` to `.gitignore`
* Tell travis to cache `vendor/bundle`

See:

* http://docs.travis-ci.com/user/caching/#Caching-directories-(Bundler%2C-dependencies)
* http://docs.travis-ci.com/user/caching/#With-a-custom-install-step

### DIFF
--- a/templates/bin_setup.erb
+++ b/templates/bin_setup.erb
@@ -8,7 +8,7 @@ set -e
 
 # Set up Ruby dependencies via Bundler
 gem install bundler --conservative
-bundle check || bundle install --path=vendor/bundle
+bundle check || bundle install
 
 # Set up configurable environment variables
 if [ ! -f .env ]; then
--- a/templates/bin_setup.erb
+++ b/templates/bin_setup.erb
@@ -8,7 +8,7 @@ set -e
 
 # Set up Ruby dependencies via Bundler
 gem install bundler --conservative
-bundle check || bundle install
+bundle check || bundle install --path=vendor/bundle
 
 # Set up configurable environment variables
 if [ ! -f .env ]; then
--- a/templates/suspenders_gitignore
+++ b/templates/suspenders_gitignore
@@ -12,3 +12,4 @@
 /public/assets
 /tags
 /tmp/*
+/vendor/bundle
--- a/templates/travis.yml.erb
+++ b/templates/travis.yml.erb
@@ -9,7 +9,8 @@ branches:
   only:
     - master
 cache:
-  - bundler
+  directories:
+    - vendor/bundle
 language:
   - ruby
 notifications:
--- a/templates/travis.yml.erb
+++ b/templates/travis.yml.erb
@@ -3,6 +3,7 @@ before_install:
   - "echo 'gem: --no-document' > ~/.gemrc"
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
+  - bundle config --local path vendor/bundle
 install:
   - bin/setup
 branches:


templates/bin_setup.erb
```diff
@@ -8,7 +8,7 @@ set -e
 
 # Set up Ruby dependencies via Bundler
 gem install bundler --conservative
-bundle check || bundle install
+bundle check || bundle install --path=vendor/bundle
```
JoelQ: Do we want this to run in development on our local machines?

jferris: This means that, locally, we'll need to install every gem for every project, right? I think that's okay, I just want to think it through.

iwz: You're right, this will affect local dev machines, as well. That's a side effect, and not necessary to fix travis. Ideas on how to implement the custom bundler path only on Travis?

mcmire: Does Travis set an environment variable when it runs, and if so, could we check that here?

derekprior: As part of a step before install, could we configure the bundler path? http://bundler.io/v1.3/man/bundle-config.1.html

Failing that, travis sets `ENV["CI"]` which we can detect and set the path option here if necessary.

teoljungberg: IIRC, Travis has configured `bundler` to install to `vendor/bundle`. I think the `--path` argument can be omitted

iwz: > IIRC, Travis has configured bundler to install to vendor/bundle. I think the --path argument can be omitted

That only works if you don't use a custom `install` config for Travis http://docs.travis-ci.com/user/caching/#With-a-custom-install-step

iwz: > As part of a step before install, could we configure the bundler path?http://bundler.io/v1.3/man/bundle-config.1.html

Great idea, didn't know about that. Will try it out.

bf4:
IMHO, the only changes in this PR should be to the travis.yml

```yaml
bundler_args: --path vendor/bundle --jobs=3 --retry=3
cache: 
  - bundler
  directories:
    - vendor/bundle
```

mcmire:
@bf4 I like that... nice and simple.

@iwz Can we update this PR and :sheep: it?

iwz:
will try to find some time soon!

mcmire:
I misspoke -- the change that @bf4 will not work with a custom install step. We can't just use `bundler_args`, that's why this PR is the way it is.

bf4:
@mcmire True, I missed that.  You could 

- modify the before_install to run `bundle config --local path vendor/bundle` 
- or maybe pass an arg to bin/setup that does the above
-or  

```bash
mkdir -p .bundle
cat <<EOF>.bundle/config
---
BUNDLE_PATH: vendor/bundle
EOF
```

mcmire:
I'd be in favor of `bundle config`. I'm trying this out on a project, will report back soon.

mcmire:
Okay, that seems to work. So basically, we need to:

* Remove `--path=vendor/bundle` from `bin/setup`
* Add `bundle config --local path vendor/bundle` to `before_install` in `.travis.yml`

iwz:
@mcmire 
> Okay, that seems to work. So basically, we need to:

> * Remove --path=vendor/bundle from bin/setup
> * Add bundle config --local path vendor/bundle to before_install in .travis.yml

resolved in 0bbd8b0

tute:
This PR might get irrelevant if we merge https://github.com/thoughtbot/suspenders/pull/592.

iwz:
@tute true, but i'd love to get this in before that hehe

tute:
Already merged. :|

iwz:
Closing as this is now obsolete. #592 wins!

PR #572, opened 2015-05-23
Inject spring into binstubs

### LINKED ISSUE #569: Generate binstubs with spring
suspenders includes spring and spring-commands-rspec but binstubs haven't been generated.
I think `bundle exec spring binstub --all` should be executed somewhere (in `suspenders new` or `bin/setup.sh` ?)

What do you think? I can try to submit a pr if this is useful :)

delphaber:
This should close #569

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,14 @@ This has the same effect as running:
 
     hub create organization/project
 
+## Spring
+
+Suspenders uses [spring](https://github.com/rails/spring) by default.
+It makes Rails applications load faster, but it might introduce confusing issues
+around stale code not being refreshed.
+If you think your application is running old code, run `spring stop`.
+And if you'd rather not use spring, add `DISABLE_SPRING=1` to your login file.
+
 ## Dependencies
 
 Suspenders requires the latest version of Ruby.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+
+* Generate rake, rails and rspec binstubs with Spring
+* Remove Capybara and use RSpec 3.2 for development
+
 1.28.0 (May 9, 2015)
 
 * Require spec/support files in a certain order
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -401,6 +401,10 @@ you can deploy to staging and production with:
       append_file "Rakefile", %{\ntask default: "bundler:audit"\n}
     end
 
+    def setup_spring
+      bundle_command "exec spring binstub --all"
+    end
+
     def copy_miscellaneous_files
       copy_file "browserslist", "browserslist"
       copy_file "errors.rb", "config/initializers/errors.rb"
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -49,6 +49,7 @@ module Suspenders
       invoke :create_github_repo
       invoke :setup_segment
       invoke :setup_bundler_audit
+      invoke :setup_spring
       invoke :outro
     end
 
@@ -187,6 +188,11 @@ module Suspenders
     def setup_bundler_audit
       say "Setting up bundler-audit"
       build :setup_bundler_audit
+    end
+
+    def setup_spring
+      say "Springifying binstubs"
+      build :setup_spring
     end
 
     def init_git
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -139,6 +139,18 @@ RSpec.describe "Suspend a new project with default configuration" do
     )
   end
 
+  it "adds spring to binstubs" do
+    run_suspenders
+
+    expect(File).to exist("#{project_path}/bin/spring")
+
+    spring_line = /^ +load File.expand_path\("\.\.\/spring", __FILE__\)$/
+    bin_stubs = %w(rake rails rspec)
+    bin_stubs.each do |bin_stub|
+      expect(IO.read("#{project_path}/bin/#{bin_stub}")).to match(spring_line)
+    end
+  end
+
   def analytics_partial
     IO.read("#{project_path}/app/views/application/_analytics.html.erb")
   end


tute:
This is great. Would you rebase on top of latest master, and update with both PRs the NEWS.md file?
Thank you!

delphaber:
Sure :)

delphaber:
Done :) I didn't know if you want to do a release now and the version number, so I didn't write it in NEWS.md

Thanks for the fast feedback in PR!

gylaz:
Does this marry all of further projects created with suspenders to Spring? I know there are a lot of folks who are not fans of Spring.

gylaz:
Perhaps Spring should be an opt-in thing?

delphaber:
@gylaz yes. Since `spring`and `spring-commands-rspec`are in `templates/Gemfile.erb`I thought that binstubs would be useful.

Rails 4.2 has a `--skip-spring` option. I could change this PR: if you pass `--skip-spring` to suspenders, then I won't insert any spring gem in your Gemfile and no binstubs at all. But this seems more like an opt-out thing :(

mcmire:
Should we add an option to `suspenders` itself?

mcmire:
Or should we comment out `spring-commands-rspec` in Gemfile.erb, to go along with the "opt in" mentality as introduced by #511, and expect you to run `bundle exec spring binstub --all` yourself, if you really want it?

derekprior:
Folks that really don't like spring can disable it system wide with an environment variable. Is that sufficient?

jferris:
> Folks that really don't like spring can disable it system wide with an environment variable. Is that sufficient?

Yes, I think this is vastly preferable over generating different code based on the preference of whoever started the project.

gabebw:
> Folks that really don't like spring can disable it system wide with an environment variable. Is that sufficient?

Can we document that? Or even print it out when we create a new project? I just got bitten by Spring, on a React project of all things (it didn't reload the JS VMs and so had old JSX available to tests).

tute:
> Can we document that? Or even print it out when we create a new project? I just got bitten by Spring, on a React project of all things (it didn't reload the JS VMs and so had old JSX available to tests).

Let's add it the README, @delphaber?

delphaber:
ok!

bf4:
And then add `DISABLE_SPRING=1` to your login file and you won't have any bizarre errors that take hours to debug... :)

tute:
We just discussed at thoughtbot and we were split: some people consider it useful (sometimes necessary); some people would rather not use it, and install it when it makes sense.

@delphaber Let's add to your PR a paragraph in the README along the lines of:

> suspenders uses [spring] by default. It makes Rails applications load faster, and it might introduce confusing issues around stale code not being not being refreshed. If you think your application is running old code, run `spring stop`. And if you'd rather not use spring, add `DISABLE_SPRING=1` to your login file.

Please improve it at will!

gylaz:
I feel like Spring is an anti-pattern. Usually Spring is used to speed up the tests, but the idea should be not to write slow tests in the first place. If your tests are slow, write code that isn't. For example, instead of testing a lot through feature specs, or specs that touch the database, we should be pulling out logic into POROs and testing that in isolation.

The idea of not using Spring is that *it's good* to feel the slowness, it will make you think about what is slow, and why it's slow and how to make it fast the *right* way, rather than adding a band-aid, which is Spring.

tute:
Spring saves Rails boot time (time spent loading code AFAIK). If you have hundreds of files there's not much you can do to make that faster, more than splitting the project in parts. PORO do help, but I never worked in an application where most of the classes and specs would load Rails, yet.

tute:
Spring saves Rails boot time (time spent loading code AFAIK). If you have hundreds of files there's not much you can do to make that faster, more than splitting the project in parts. PORO do help, but I never worked in an application where most of the classes and specs wouldn't load Rails, yet.

gylaz:
> Spring saves Rails boot time (time spent loading code AFAIK). If you have hundreds of files there's not much you can do to make that faster, more than splitting the project in parts.

The idea is *not* to load all of Rails, when running/writing most specs. Feeling the pain/slowness (which Spring is supposed to mitigate) drives us write better, more decoupled (from Rails) code.

IMO, we shouldn't start a brand new project, relying on Spring. If we structure the code correctly, and are conscious about the slowness of Rails, we shouldn't need Spring. [hound](https://github.com/thoughtbot/hound) is a mature app that uses that philosophy -- our test suite runs in under 30 seconds, and we've never felt the need to add Spring. For example, feature specs load all of Rails, but the slowness of those specs make us want to write very little of them, while unit testing the logic at lower levels.

The argument I'm trying to make is that we shouldn't *need* Spring. It should be a situational tool, and used for existing, legacy projects, where you can't decouple logic from Rails easily. Thus, I don't think Spring is a good candidate for suspenders.

mike-burns:
If you made a PR to remove Spring from Suspenders, I would review it and err on the side of giving it a thumbs-up.

(I tried and failed in #511, but didn't look into why.)

tute:
I can agree with that. And the CodeClimate 10-second boot time would be an outlier for which it makes sense to manually add it.

bf4:
Some choices as I see it

1. keep spring
2. comment out spring
3. keep spring, but auto export DISABLE_SPRING=1 in dotfiles

jferris:
We currently do include Spring in Suspenders, and Spring is part of Rails. If you generate a new Rails app, you'll get Spring binstubs. Including Spring but not having the binstubs means that somebody will need to generate them for each new project whenever they happen to notice.

I think the discussion of whether or not Spring is a good idea at all is separate from the one of "should Spring projects use Spring binstubs."

mcmire:
> I think the discussion of whether or not Spring is a good idea at all is separate from the one of "should Spring projects use Spring binstubs."

Well, I think the question is, should we accept this PR or should we remove Spring from Suspenders. 

Keep in mind that putting Spring in the Gemfile or having Suspenders generate binstubs doesn't prevent you from opting out of using Spring on a project -- unless you have `./bin` in your PATH.

So, two questions:

1) Do most people at thoughtbot have `./bin` in their PATH? (or `.git/safe/../../bin`, but you know what I mean)
2) Are we concerned that someone will see that Suspenders includes Spring and think "oh, everyone at thoughtbot uses Spring"?

jferris:
> unless you have `./bin` in your `PATH`.

You can still opt out by setting `DISABLE_SPRING`.

derekprior:
> Well, I think the question is, should we accept this PR or should we remove Spring from Suspenders.

I think removing spring would be a question better suited for its own PR. In the world we're in right now, suspenders has spring, has bin stubs, but does not have springified binstubs. This PR is a way to get consistent at a very low cost (we're not somehow tied to spring moving forward).

graysonwright:
@derekprior that makes a lot of sense to me. I'm in favor of merging this PR and opening a separate discussion about whether or not to remove Spring.

delphaber:
@tute I added the spring section in readme :)

README.md
```diff
@@ -157,6 +157,14 @@ This has the same effect as running:
 
     hub create organization/project
 
+## Spring
+
+Suspenders uses [spring](https://github.com/rails/spring) by default.
+It makes Rails applications load faster, and it might introduce confusing issues
```
andyw8: Should this be 'but' rather than 'and'?

andyw8: Extra space before 'not'.

README.md
```diff
@@ -157,6 +157,14 @@ This has the same effect as running:
...
+around stale code  not being refreshed.
```
andyw8: Extra space before 'not'

delphaber:
@andyw8 fixed thx :)

tute:
Will merge this PR, our binstubs will use spring, and if a PR is submitted removing spring altogether I'll give it my :+1:. Thank you all!

PR #573, opened 2015-05-23
Remove capybara (dev dep) and use rspec 3.2

delphaber:
I removed `capybara` from development dependencies because it was only used for `feature`and `scenario`dsl in specs. I converted them in `describe`and `it`.

Then I run transpec to convert rspec 2.x tests to rspec 3.x, but no convertion was needed.

Finally, I updated rspec to 3.2


tute:
This is great, thank you!

PR #576, opened 2015-06-05
Add view and helper spec helpers

joshuaclayton:
This disables verification of partial doubles in view and helper specs.

What does this mean for you?

1. You can stub locals in view partial specs instead of doing the funky `render "partial/partial"...` dance
2. You can stub `current_user`, `signed_in?`, or other methods defined in other controllers/helpers that you don't normally have access to

Potential downsides:

1. You can stub methods on view that aren't normally defined. Apart from what RSpec considers not available (e.g. `current_user`, etc.) this means you can allow the view to receive literally whatever you want. This may be bad form, although @derekprior made a valid point that, within views, you're very typically interacting with various objects, so the fact that something may or may not be defined (and further stubbing of it) may not be that bad.

### DIFF
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -49,6 +49,18 @@ RSpec.describe "Suspend a new project with default configuration" do
     expect(File).to exist("#{project_path}/spec/support/i18n.rb")
   end
 
+  it "adds support file for view specs" do
+    run_suspenders
+
+    expect(File).to exist("#{project_path}/spec/support/view_spec_helpers.rb")
+  end
+
+  it "adds support file for helper specs" do
+    run_suspenders
+
+    expect(File).to exist("#{project_path}/spec/support/helper_spec_helpers.rb")
+  end
+
   it "ensures newrelic.yml reads NewRelic license from env" do
     run_suspenders
 
--- a/templates/helper_spec_helpers.rb
+++ b/templates/helper_spec_helpers.rb
@@ -1,0 +1,17 @@
+module HelperSpecHelpers
+  def stub_current_user(as: nil)
+    allow(view).to receive(:current_user).and_return(as)
+  end
+end
+
+RSpec.configure do |config|
+  config.include HelperSpecHelpers, type: :helper
+
+  config.around(:each, type: :helper, allow_helper_stub: true) do |example|
+    config.mock_with :rspec do |mocks|
+      mocks.verify_partial_doubles = false
+      example.run
+      mocks.verify_partial_doubles = true
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


templates/helper_spec_helpers.rb
```diff
@@ -0,0 +1,17 @@
+module HelperSpecHelpers
+  def stub_current_user(as: nil)
```
jferris: Should this be part of Clearance's RSpec support? I think we should try to keep Suspenders as just configuration whenever possible. Otherwise, we're essentially copying and pasting code into every project.

templates/view_spec_helpers.rb
```diff
@@ -0,0 +1,21 @@
+module ViewSpecHelpers
+  def page
```
derekprior: I do this on every project. I think this makes a ton of sense.

templates/helper_spec_helpers.rb
derekprior: What percentage of our projects are using clearance to provide `current_user`?

tute:
:+1:

templates/helper_spec_helpers.rb
jferris: A good number, I hope!

Clearance was the most obvious place to me, but I'd like to find homes for methods like this (even though they're small) in other projects. It's really easy for Suspenders to become bloated, and I don't want our new Rails apps to start out with a hundred lines of actual code we need to maintain. People complain enough already about the number of gems we include.

If we want a Clearance-agnostic approach for this helper, maybe something like Shoulda makes sense for this?

templates/view_spec_helpers.rb
jferris: Capybara has some Rails integration already. Maybe they'd be interested in a pull request which includes this method in view specs?

joshuaclayton: @jferris @derekprior I opened a [PR on Capybara](https://github.com/jnicklas/capybara/pull/1533); please provide any feedback there, especially if you think this would be a helpful change.

gabebw:
Whoa, neat: https://github.com/jnicklas/capybara/pull/1533

spec/features/new_project_spec.rb
```diff
@@ -49,6 +49,18 @@
     expect(File).to exist("#{project_path}/spec/support/i18n.rb")
   end
 
+  it "adds support file for view specs" do
+    run_suspenders
```
croaky: Each `run_suspenders` takes about 30-60s, I believe? Should we use a `before` block in this spec? Could save 90+% of the test time.

croaky: Attempting in https://github.com/thoughtbot/suspenders/pull/579

croaky:
Looks good to me.

tute:
@joshuaclayton do we still need this PR?

joshuaclayton:
@tute nope, I'll close it; thanks everyone!

PR #578, opened 2015-06-12
Add default .hound.yml file

croaky:
* Turn on JavaScript but not CoffeeScript by default.
* Help avoid configuration mistakes by providing this stub
  and a link to the configuration page.

### DIFF
--- a/templates/hound.yml
+++ b/templates/hound.yml
@@ -1,0 +1,17 @@
+# See https://houndci.com/configuration for help.
+coffeescript:
+  # config_file: .coffeescript-style.json
+  enabled: true
+haml:
+  # config_file: .haml-style.yml
+  enabled: true
+javascript:
+  # config_file: .javascript-style.json
+  enabled: true
+  # ignore_file: .javascript_ignore
+ruby:
+  # config_file: .ruby-style.yml
+  enabled: true
+scss:
+  # config_file: .scss-style.yml
+  enabled: true
--- a/README.md
+++ b/README.md
@@ -97,7 +97,8 @@ Suspenders also comes with:
 * [t() and l() in specs without prefixing with I18n][i18n]
 * An automatically-created `SECRET_KEY_BASE` environment variable in all
   environments
-* Configuration for [Travis Pro][travis] continuous integration
+* Configuration for [Travis][travis] Continuous Integration (tests)
+* Configuration for [Hound][hound] Continuous Integration (style)
 * The analytics adapter [Segment][segment] (and therefore config for Google
   Analytics, Intercom, Facebook Ads, Twitter Ads, etc.)
 
@@ -107,6 +108,7 @@ Suspenders also comes with:
 [binstub]: https://github.com/thoughtbot/suspenders/pull/282
 [i18n]: https://github.com/thoughtbot/suspenders/pull/304
 [travis]: http://docs.travis-ci.com/user/travis-pro/
+[hound]: https://houndci.com
 [segment]: https://segment.com
 
 ## Heroku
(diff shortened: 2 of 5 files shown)


templates/hound.yml
```diff
@@ -0,0 +1,11 @@
+# See https://houndci.com/configuration for configuration details.
+coffeescript:
+  enabled: true
+haml:
+  enabled: true
+javascript:
+  enabled: false
```
gylaz: Why is this false?

templates/hound.yml
```diff
@@ -0,0 +1,11 @@
...
+ruby:
+  enabled: true
+scss:
+  enabled: true
```
gylaz: Since these are enabled in Hound by default, is the idea here to be explicit anyway?

gylaz:
> Help avoid configuration mistakes by providing this stub and a link to the configuration page.

In the spirit of ^, would it be useful to add a commented out `config_file:` option?
```yaml
coffeescript:
  enabled: true
  # config_file: .coffeescript-style.json
```

templates/hound.yml
croaky: I was thinking thoughtbot's guides recommend CS https://github.com/thoughtbot/guides/tree/master/best-practices#javascript. Suspenders provides defaults for what a thoughtbot project would be most of the time and if you want something different, you can be explicit. Should we flip it now that ES6 is becoming more common?

templates/hound.yml
croaky: Yup, explicitness, easy to toggle on and off by editing the file rather than looking up the docs.

croaky:
> would it be useful to add a commented out config_file: option?

I had at that first, thought it was maybe a bit too noisy, but agree part of the benefit here is being able to comment/uncomment quickly. Just pushed 6243245 to bring these back.

templates/hound.yml
```diff
@@ -0,0 +1,19 @@
...
+  # config_file: .javascript-style.json
+  enabled: true
+  # ignore_file: .javascript_ignore
+ruby:
+  # config_file: .ruby-style.yml
+  enabled: true
+scss:
+  # config_file: .scss-style.yml
+  enabled: true
+  # exclude:
```
gylaz: The `exclude` should be in `.scss-style.yml`.

templates/hound.yml
```diff
@@ -0,0 +1,17 @@
+# See https://houndci.com/configuration for help.
+coffeescript:
+  # config_file: .coffeescript-style.json
+  enabled: false
```
gylaz: How about having them all be enabled by default, and it's up to the project wether they are using CS or JS (since Hound will review based on file extensions)? Possible that some projects might have a mix of CS and JS for whatever reasons.

croaky: Suspenders creates a new project, so it seems unlikely that the plan would be to mix CS and JS. Suspenders has a solid history of providing a good default opinion. I think we should pick one or the other and run with that.

croaky:
@tute @jferris When you have a moment. No rush. Interested in your thoughts.

templates/hound.yml
gylaz: I guess what I'm trying to say is that there is no cost to enabling all. The benefit to having everything enabled, is that *if* there is CoffeeScript file, it will be reviewed. Is there ever a need *not* to review CS if it's in the project?

I think CoffeeScript is still the winner according to our guides, so if anything JS should be disabled.

croaky: Cool, enabled all.

tute:
> Turn on JavaScript but not CoffeeScript by default.

Both look enabled, and that's what I'd expect.

I like this!

PR #579, opened 2015-06-12
Improve test suite speed

croaky:
* Run Suspenders only once in `new_project_spec.rb`.
* Should be about 17x faster.

### DIFF
--- a/spec/features/github_spec.rb
+++ b/spec/features/github_spec.rb
@@ -1,6 +1,11 @@
 require "spec_helper"
 
 RSpec.describe "GitHub" do
+  before do
+    drop_dummy_database
+    remove_project_directory
+  end
+
   it "suspends a project with --github option" do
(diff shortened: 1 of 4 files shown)


tute:
This makes infinite sense. Thanks so much, Dan!

joshuaclayton:
Huge :+1: from me, the test suite was a huge pain point when I just got started. I like the benefits of isolating individual assertions in separate tests, but I prefer speed :-)

My other thought was to try out FakeFS to attempt to alleviate this, but I haven't done so yet.

croaky:
Average successful test run times before this change: 10-15m. Afterward: 5m. So, not as large a change as I was thinking but still nice.

tute:
"Still nice"? I think this is a big deal!

gabebw:
Whoa. Great job, Dan.

PR #581, opened 2015-06-17
Allow any Rails version in the 4.2.x series

JoelQ:
Locking down the rails version to the patch level means we need to edit
the Gemfile every time a security update is released. By locking to the
4.2.x series instead, we can just run `bundle update rails` to get the
latest security patches.

Because this is locked to an explicit minor version, only security
patches should get pulled in automatically.  Assuming that Rails follows
SemVer properly, unexpected changes should only come in minor or major
versions.

I ran `bundle install` and pulled in the security patches released for
Rails 4.2.2 (see
http://weblog.rubyonrails.org/2015/6/16/Rails-3-2-22-4-1-11-and-4-2-2-have-been-released-and-more/
for details and CVEs)


tute:
:shipit:

tute:
Thank you Joël!

PR #583, opened 2015-06-25
Add a PageObject class to generated app

gabebw:
It can use Capybara methods like `visit` or `find` directly; it can use `t` instead of `I18n.t`; it has all the Formulaic helpers like `fill_form`; it has all the FactoryGirl helpers.

Example:

```ruby
class UserPage < PageObject
  def initialize
    @user = create(:user)
  end

  def visit_profile
    visit user_path(@user)
  end
end
```

### DIFF
--- a/templates/page_object.rb
+++ b/templates/page_object.rb
@@ -1,0 +1,7 @@
+class PageObject
+  include FactoryGirl::Syntax::Methods
+  include Capybara::DSL
+  include Rails.application.routes.url_helpers
+  include AbstractController::Translation
+  include Formulaic::Dsl
+end
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ Suspenders also comes with:
 * Configuration for [Hound][hound] Continuous Integration (style)
 * The analytics adapter [Segment][segment] (and therefore config for Google
   Analytics, Intercom, Facebook Ads, Twitter Ads, etc.)
+* A `PageObject` class for [easier testing with page objects][page objects]
 
 [setup]: http://robots.thoughtbot.com/bin-setup
 [compress]: http://robots.thoughtbot.com/content-compression-with-rack-deflater/
@@ -110,6 +111,7 @@ Suspenders also comes with:
 [travis]: http://docs.travis-ci.com/user/travis-pro/
 [hound]: https://houndci.com
 [segment]: https://segment.com
+[page objects]: https://robots.thoughtbot.com/better-acceptance-tests-with-page-objects
 
 ## Heroku
 
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -217,6 +217,10 @@ end
       copy_file "i18n.rb", "spec/support/i18n.rb"
     end
 
+    def add_page_objects_for_tests
+      copy_file "page_object.rb", "spec/support/page_object.rb"
+    end
+
     def configure_i18n_for_missing_translations
       raise_on_missing_translations_in("development")
       raise_on_missing_translations_in("test")
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -98,6 +98,7 @@ module Suspenders
       build :configure_travis
       build :configure_i18n_for_test_environment
       build :configure_i18n_tasks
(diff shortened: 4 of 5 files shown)


templates/page_object.rb
```diff
@@ -0,0 +1,7 @@
+class PageObject
```
gabebw: We've been using this on my client project with great success.

jferris: This seems cool, but I strongly believe that Suspenders should contain configuration as much as possible, and we should limit the amount of code we copy into generated apps.

We don't use the page object pattern on every application, and I think it would make sense to verify that we want this to be our standard approach before including it.

Could this code form the base of a library for building tests from page objects?

gabebw: That's an interesting idea.

gabebw: Here's a previous Trello card on page objects, from 2013; we marked it as a Success: https://trello.com/c/CPilXWup/93-page-objects-in-test-suites

A lot of them look very similar to this.

jferris: The experiment wasn't as well-defined in that card as it generally is for our newer ones. For example, there isn't a success criteria or an expected outcome. The feelings on that card also weren't unanimous, so while I'd agree that page objects have proven to be useful, I don't think that we've shown as a team that we want to use page objects on all new projects going forward.

For projects that do want page objects, I think it would be great to define a library with a base class that provides them, rather than copying and pasting them into all new projects. To me, the threshold for Suspenders is, "will I want to copy this code into every project I make?" I think we can find a better way to share this particular piece of code than copying and pasting it.

gabebw: Yes, I agree. I feel like we're arguing, which wasn't my intention - I
think a page_object library with a punny name is a good idea.

I have actually copied this exact code into a lot of projects, though.
Perhaps this deserves its own Trello card?

On Thu, Jun 25, 2015 at 4:07 PM, Joe Ferris <notifications@github.com>
wrote:

> In templates/page_object.rb
> <https://github.com/thoughtbot/suspenders/pull/583#discussion_r33298705>:
>
> > @@ -0,0 +1,7 @@
> > +class PageObject
>
> The experiment wasn't as well-defined in that card as it generally is for
> our newer ones. For example, there isn't a success criteria or an expected
> outcome. The feelings on that card also weren't unanimous, so while I'd
> agree that page objects have proven to be useful, I don't think that we've
> shown as a team that we want to use page objects on all new projects going
> forward.
>
> For projects that do want page objects, I think it would be great to
> define a library with a base class that provides them, rather than copying
> and pasting them into all new projects. To me, the threshold for Suspenders
> is, "will I want to copy this code into every project I make?" I think we
> can find a better way to share this particular piece of code than copying
> and pasting it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/thoughtbot/suspenders/pull/583/files#r33298705>.
>



-- 
Gabe Berke-Williams
http://gabebw.com/

jferris: > I feel like we're arguing, which wasn't my intention

I disagree!



> I have actually copied this exact code into a lot of projects, though. Perhaps this deserves its own Trello card?

We've proven that some people are interested in page objects, so I think it would be worthwhile to try and see what that looks like as a library. Extracting the real code you've already used on several projects seems like a good place to start.

I think a Trello card to track that pilot project would be a good idea as well.

gabebw: Trello card: https://trello.com/c/JeCCTPkf/512-page-object-library

gylaz: I like Martin Fowler's take on [Page Objects]](http://martinfowler.com/bliki/PageObject.html)).

I've followed his advice:
> Despite the term "page" object, these objects shouldn't usually be built for each page, but rather for the significant elements on a page [1].
and it has allowed me to re-use page objects across feature specs.

gabebw: Greg, that sounds great, but I'm not quite sure what you mean or what you'd change. Perhaps some code would clear it up, or maybe you can write a little more?

gylaz: I was simply contributing to the discussion.

mcmire:
Should we close this PR now and focus on developing a library instead?

gabebw:
Yep.

PR #584, opened 2015-06-25
Update to RSpec 3.3 in suspenders apps

derekprior:
RSpec 3.3 added some very handy commands for filtering examples. With
the addition of the status persistence file configuration we can now:

* run `rspec --only-failures` to re-run only the examples that failed.
* run `rspec --next-failure` to re-run only the first failing test from
  your previous run. If that passes it will run the successive failures
  until one of them fails again

Additionally, rspec 3.3 includes stable random ordering (even when you
add new examples), the ability to bisect failures when debugging
ordering-dependent failures, better failure output...

http://rspec.info/blog/2015/06/rspec-3-3-has-been-released/

### DIFF
--- a/templates/spec_helper.rb
+++ b/templates/spec_helper.rb
@@ -16,7 +16,7 @@ RSpec.configure do |config|
     mocks.verify_partial_doubles = true
   end
 
-  config.example_status_persistence_file_path = "tmp/examples.txt"
+  config.example_status_persistence_file_path = "tmp/rspec_examples.txt"
   config.order = :random
 end
 


teoljungberg:
I'm into this

calleerlandsson:
Cool! I'm in favor.

templates/spec_helper.rb
```diff
@@ -16,6 +16,7 @@
     mocks.verify_partial_doubles = true
   end
 
+  config.example_status_persistence_file_path = "tmp/examples.txt"
```
tute: Let's call this file `tmp/rspec_examples.txt`?

tute:
:+1:

mcmire:
Oh man! I'm glad they're adding support for re-running failed tests. This is an idea I proposed a [long time ago](https://github.com/rspec/rspec-core/issues/456) but didn't get around to making a PR.

This looks good to me (I agree with the comment Tute left above)

templates/Gemfile.erb
```diff
@@ -39,7 +39,7 @@ group :development, :test do
   gem "dotenv-rails"
   gem "factory_girl_rails"
   gem "pry-rails"
-  gem "rspec-rails", "~> 3.1.0"
+  gem "rspec-rails", "~> 3.3.0"
```
JoelQ: Thoughts on using `"~> 3.3" so as to automatically pick up minor versions?

tute: Both ways result in the same behavior.

JoelQ: Won't `"~> 3.3.0"` update to `3.3.1` but not `3.4.0`

tute: I didn't get your "minor" mention. I was not aware that it works that way, I'm in favor of the change if that is the case.

derekprior: I'd be in favor of changing it to `>= 3.3.0` myself, but I think that's a change for a different PR.

delphaber:
Hi, I think there aren't any tests for `spec_helper`and `rails_helper`. Maybe these lines could be useful:

```ruby
  it "adds rspec's helper files" do
    expect(File).to exist("#{project_path}/spec/spec_helper.rb")
    expect(File).to exist("#{project_path}/spec/rails_helper.rb")
  end
````

PR #588, opened 2015-07-05
Fix memoization for rails's server port value

delphaber:
I think multiple instance of `AppBuilder` are currently being initialized during `suspenders APP_NAME` process because `bin/setup` and `development.rb`have different port values.

So I think we can use a class variable to memoize the value. Is it dirty?

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -485,7 +485,7 @@ end
     end
 
     def port
-      @port ||= [3000, 4000, 5000, 7000, 8000, 9000].sample
+      @@port ||= [3000, 4000, 5000, 7000, 8000, 9000].sample
     end
 
     def serve_static_files_line


tute:
> I think multiple instance of AppBuilder are currently being initialized during `suspenders APP_NAME`

I add two questions: Why would be initialize more than one `AppBuilder`? Also surprising to me is the `sample`; why a sample, instead of the default or a given parameter?

delphaber:
> Why would be initialize more than one `AppBuilder`?
I don't know why :(
> why a sample, instead of the default or a given parameter?
It was introduced here: https://github.com/thoughtbot/suspenders/commit/4caa6123c83e726d6fcd23956ce9883ca36b1d42

tute:
In https://github.com/thoughtbot/suspenders/issues/602#issuecomment-130671649 we discussed the bug we are working around here. @jferris what do you think of simplifying suspenders, making it use default port 3000 always?

jferris:
> What do you think of simplifying suspenders, making it use default port 3000 always?

I have no objection to this, as I generally only run one project at a time. I'm not sure if most people run one project at once or several.

That change was introduced by @croaky, because I think he's in the habit of keeping several projects running simultaneously.

@croaky - any thoughts on this?

croaky:
Yup, I run multiple projects at once for long periods of time.

I don't feel strongly either way. Do what you feel.

tute:
@delphaber do you want to remove the code around random ports from suspenders?

delphaber:
Ok, I'm gonna leave it `3000` as default

delphaber:
Done in https://github.com/thoughtbot/suspenders/pull/611

I'm closing this.

PR #590, opened 2015-07-15
Move recipient_interceptor to :staging gem group

delphaber:
Since recipient interceptor is used only in staging, could we move the gem in the appropriate group?

### DIFF
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -18,7 +18,6 @@ gem "normalize-rails", "~> 3.0.0"
 gem "pg"
 gem "rack-canonical-host"
 gem "rails", "<%= Suspenders::RAILS_VERSION %>"
-gem "recipient_interceptor"
 gem "refills"
 gem "sass-rails", "~> 5.0"
 gem "simple_form"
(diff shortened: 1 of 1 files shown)


templates/Gemfile.erb
```diff
@@ -56,3 +55,7 @@ end
 group :staging, :production do
   gem "rack-timeout"
 end
+
+group :staging do
```
edwardloveall: I think it should also be in development and test. You don't want to be sending emails there either.

tute:
Thanks! I'm ok with having it globally. That way we can register interceptors in environments without needing to think if code is loaded or not, the library does nothing if no interceptors are registered (like in production), and the library is tiny.

edwardloveall:
Even if it's tiny, I still feel like it's unnecessary to have it in production at all. Even tiny code can cause problems, and if you can avoid those problems, I think it's better.

mike-burns:
I do agree with that reasoning, but I also worry about staging and production being different. If even tiny code can cause problems, I want to catch those on staging first.

edwardloveall:
That's a good reason I hadn't considered. It's awkward to have a gem you're not intending to use, but you're right, bugs are worse.

tute:
Leaving it globally, thanks for your work!

PR #592, opened 2015-07-24
Replace Travis with CircleCI and add auto deploy to staging

drapergeek:
This is the result of two discussions on our research board. The first was to replace Travis with CircleCI:
https://trello.com/c/Se4mELMh/199-hosted-ci-for-continuous-deployment

A few reasons we've enjoyed using Circle:
* Faster builds due to parallel builds
* Built in bundle caching
* Easy deployment
* Really easy setup
* The ability to ssh into the CI server during a test run. This has been great when debugging a test failure. It's not something you use a lot but it is so great when you need it.
* A free plan for private repos that supports small MVPS

This also adds automatic deployment to staging when the build passes. This is the result of this discussion:
https://trello.com/c/Se4mELMh/199-hosted-ci-for-continuous-deployment

A few reasons for those who don't have access to the trello card:
* Part of the CI now is whether or not you can actually deploy successfully. If you merge to master, it will automatically try and deploy to staging. If that fails, your build will actually fail. This should be treated with the same urgency that any other failing build on master would be treated.
* Staging is always up to date. No options. This makes it really nice for verifying stories for clients.
* Automate more things!


lib/suspenders/app_builder.rb
```diff
@@ -209,8 +209,8 @@ def configure_rspec
       copy_file "spec_helper.rb", "spec/spec_helper.rb"
     end
 
-    def configure_travis
-      template 'travis.yml.erb', '.travis.yml'
+    def configure_ci
+      template 'circle.yml.erb', 'circle.yml'
```
gabebw: Does this need to be an erb file? It looks like it's pure YAML.

gabebw: Actually, it looks like we lost specifying the Ruby version in Circle. I think it needs this, per https://circleci.com/docs/configuration#ruby-version:

```ruby
machine:
  ruby:
    version: <%= Suspenders::RUBY_VERSION %>
```

drapergeek: It automatically uses your `.ruby-version`, is there any reason we should specify it twice? That seems like it would just make it harder to upgrade.


I left it as an erb just in case we ended up adding dynamic information, happy to remove it though, I don't feel strongly either way.

gabebw: > It automatically uses your .ruby-version

Neat! That all sounds reasonable.

tute:
LGTM.

drapergeek:
Thanks @tute and @gabebw

croaky:
We don't want to set up the project with `bin/setup` on CI?

jferris:
> We don't want to set up the project with bin/setup on CI?

I also liked that we used the same `bin/setup` locally and on CI.

PR #593, opened 2015-07-28
Rename development data concept to avoid db/seeds confusion

mjankowski:
Using db/seeds for data which must be loaded in a production environment
is a standard Rails convention.  We cause confusion by using the word
"seed" elsewhere in a Rails app, where that is not the context.

The `dev:prime` task is actually providing local fake/sample data for
developers to see the app in use, but not production-required data to
actually run the application.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -33,7 +33,7 @@ module Suspenders
     end
 
     def provide_dev_prime_task
-      copy_file 'development_seeds.rb', 'lib/tasks/development_seeds.rake'
+      copy_file 'dev.rake', 'lib/tasks/dev.rake'
     end
 
     def configure_generators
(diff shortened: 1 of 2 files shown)


lib/suspenders/app_builder.rb
```diff
@@ -33,7 +33,7 @@ def provide_setup_script
     end
 
     def provide_dev_prime_task
-      copy_file 'development_seeds.rb', 'lib/tasks/development_seeds.rake'
+      copy_file 'development_data.rb', 'lib/tasks/development_data.rake'
```
bernerdschaefer: What about `dev.rake`, so it matches its namespace?

bernerdschaefer:
Thanks for this! :+1:

tute:
:+1:

lib/suspenders/app_builder.rb
jferris: > What about `dev.rake`, so it matches its namespace?

Agreed! It's easier to find rake tasks if they're defined according to convention.

bf4:
Ref: https://github.com/rails/rails/pull/20881

PR #595, opened 2015-08-01
Add ctags configuration dotfile per project basis

gylaz:
Why:

* Use a config dotfile for ctag options.
* Can add more options here on per-project basis,
  like excluding directories, and support for language specific DSLs.
* Start breaking up `templates/` directory into smaller subdirectories.

Related to https://github.com/thoughtbot/dotfiles/pull/390

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -320,6 +320,10 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
       end
     end
 
+    def copy_dotfiles
+      directory("dotfiles", ".")
+    end
+
     def init_git
       run 'git init'
     end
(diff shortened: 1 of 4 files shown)


tute:
LGTM. You mention "Start breaking up `templates/` directory into smaller subdirectories" in your commit message but I don't see related changes.

gylaz:
> ... but I don't see related changes.

@tute The new file is in `templates/dotfiles/`, and that directory gets copied by `directory("dotfiles", ".")` line. The idea is to follow that direction, and eventually pull out more files into `templates/dotfiles/` as well as start new sub-directories. Does that make sense?

tute:
It makes sense, thanks for your clarification.

tute:
Thank you @gylaz! I merged it into master.

Issue #594, opened 2015-07-29
Circle auto-deploys not working


iwz:
It seems that we may be missing a piece or two to get new apps up and running with auto-deploys from circle.

We may need to add the git remote called `staging`, pointed to the correct heroku app:

```yaml
test:
  pre:
    - cp .sample.env .env
deployment:
  staging:
    branch: master
    commands:
      - git remote add staging git@heroku.com:APP_NAME.git
      - ./bin/deploy staging
```

We also need to configure Circle's settings with the Heroku API and SSH key. This is a manual step, AFAICT, and it's not documented in this repo, I don't think.

tute:
Thank you @iwz! I was on vacations and then forgot to checkout this repo, sorry about the delay!

There is a PR to get our circle config up to speed: https://github.com/thoughtbot/suspenders/pull/596.

Will ping the contributor there to incorporate the code in this issue. Thanks!

PR #596, opened 2015-08-01
Update CircleCI config

pedrosmmoreira:
The current circle.yml template does not set the env vars before running the build, which can lead to build failures. It also doesn't set a db setup command, which causes Circle to run `rake db:create
db:schema:load` by inference.

This PR sets the command to copy env vars from the .sample.env file post checkout, following CircleCI's recommendation for this type of operation: https://circleci.com/docs/manually#checkout
This allows for the env vars to be available for all steps of the build,making it safe to run commands that load an environment, particularly`rake`.

It also sets the database setup command to `bundle exec rake db:setup` to keep in line with the Suspenders setup script, even though I believe it is technically equivalent to the inferred command.

### DIFF
--- a/templates/circle.yml.erb
+++ b/templates/circle.yml.erb
@@ -1,8 +1,9 @@
+checkout:
+  post:
+    - cp .sample.env .env
+database:
+  override:
+    - bin/setup
 test:
   override:
     - bundle exec rake
-deployment:
-  staging:
-    branch: master
-    commands:
-      - bin/deploy staging


templates/circle.yml.erb
```diff
@@ -1,3 +1,9 @@
+checkout:
+  post:
+    - cp .sample.env .env
+database:
+  override:
+    - bundle exec rake db:setup
```
tute: `setup` script also runs seeds. Do we ever need to run seeds before test runs? I am ok with leaving DB setup to Circle CI's default.

tute:
One question. Thanks for your work!

templates/circle.yml.erb
pedrosmmoreira: I considered that, but was under the impression that this template deliberately didn't use/fully reproduce the `setup` script. The reason I added the db setup override was to keep parity between the commands used in development and CI. If you think this doesn't add value, I'm happy to remove it ;)

derekprior: In the travis version of the configuration we actually ran `bin/setup` which does more than required for running tests but is a useful test of the new developer/designer onboarding workflow. See discussion here: https://github.com/thoughtbot/suspenders/pull/441

tute: > In the travis version of the configuration we actually ran bin/setup which does more than required for running tests but is a useful test of the new developer/designer onboarding workflow. See discussion here: #441

@pedrosmmoreira what do you think of running `bin/setup` in the section `database.override`?

tute:
@pedrosmmoreira if you'd like to incorporate the staging auto deploy mentioned in https://github.com/thoughtbot/suspenders/issues/594 that would be great. We can do it as part of a different PR too.

pedrosmmoreira:
@tute Happy to include those changes in as soon as I can ;)

pedrosmmoreira:
@tute I've ran some experiments with the auto deploy mentioned in #594 and that raises some questions, namely that the `bin/deploy` script is only present for suspenders apps generated with the `--heroku` flag.

Since that is an optional flag, we can either investigate appending the auto deploy as part of the heroku generation or we can assume it shouldn't be set "out of the box". Happy to go with either direction :)

tute:
Let's do the auto deploy in a different PR. After we run `bin/setup` in Circle CI, I'll merge. Thank you!

pedrosmmoreira:
Sounds good. Shall we eliminate the deployment section in `circle.yml` in this PR and reimplement later as it is broken currently?

tute:
> Shall we eliminate the deployment section in circle.yml in this PR and reimplement later as it is broken currently?

Sounds good to me.

pedrosmmoreira:
All done :)

tute:
Thank you!

pedrosmmoreira:
Glad to help :)

PR #597, opened 2015-08-03
Remove Gemfile.lock

tute:
suspenders as a gem relies on version ranges for its dependencies
defined in its gemspec. Precise patch versions declared in
`Gemfile.lock` don't need to be tracked.

We expect suspenders to always work well with latest patch versions of
all its dependencies.

More information:
http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/


tute:
Will avoid steps like https://github.com/thoughtbot/suspenders/pull/571/files during development.

mcmire:
:100:

tute:
Added it to the gitignore. Thanks @mcmire!

PR #604, opened 2015-08-25
Generate Heroku deployable apps by default, and have a different script to set up heroku services

gabebw:
* I consistently forget to pass `--heroku true`, even though 100% of apps I work on are deployed to Heroku
* Without the `--heroku true` option, the apps don't have key gems like `rails_stdout_logging`, so errors don't have backtraces
* thoughtbot uses Heroku for all new projects unless there's a compelling reason not to. Since Suspenders is our application template, it should reflect our real-world usage.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 1.31.0 (unreleased)
 
 * Update to Ruby 2.2.3
+* Assume new apps will be deployed to Heroku (like `--heroku true` in previous versions)
 
 1.30.0 (July 30, 2015)
 
--- a/README.md
+++ b/README.md
@@ -113,11 +113,12 @@ Suspenders also comes with:
 
 ## Heroku
 
-You can optionally create Heroku staging and production apps:
+By default, suspenders will create Heroku staging and production apps. To turn
+off this behavior:
 
-    suspenders app --heroku true
+    suspenders app --heroku false
 
-This:
+The Heroku integration:
 
 * Creates a staging and production Heroku app
 * Sets them as `staging` and `production` Git remotes
@@ -133,7 +134,6 @@ This:
 You can optionally specify alternate Heroku flags:
 
     suspenders app \
-      --heroku true \
       --heroku-flags "--region eu --addons newrelic,sendgrid,ssl"
 
 See all possible Heroku flags:
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -6,7 +6,7 @@ module Suspenders
     class_option :database, type: :string, aliases: "-d", default: "postgresql",
       desc: "Configure for selected database (options: #{DATABASES.join("/")})"
 
-    class_option :heroku, type: :boolean, aliases: "-H", default: false,
+    class_option :heroku, type: :boolean, aliases: "-H", default: true,
       desc: "Create staging and production Heroku apps"
 
     class_option :heroku_flags, type: :string, default: "",


derekprior:
:cosign:

gabebw:
It is a little worrisome that none of the tests fail now.

mike-burns:
Let it be know: I dislike automatic things, especially with side effects.

tute:
> Let it be know: I dislike automatic things, especially with side effects.

I feel the same way. I just got bit by this, and I'm fine with that.

derekprior:
This entire project is an automatic side effect.

Getting bit by this early in development would be annoying. Having multiple
coworkers get bit by it many times a month is worse. Getting but by it when
you really need to see some log data would be awful.

Seems like a worthwhile protection to me.

mike-burns:
To be clear: these PR is for the opposite of what I, @tute , and @derekprior want. Unless I misunderstand what this PR does?

mcmire:
I think this makes total sense. I'd rather this happen automatically than have to do this manually.

iwz:
@mike-burns 

> this PR is for the opposite of what [...] @derekprior wants. 

From the PR:

> Without the --heroku true option, the apps don't have key gems like rails_stdout_logging, so errors don't have backtraces

@derekprior said:

> Getting bit by it when you really need to see some log data would be awful. Seems like a worthwhile protection to me.

I believe Derek is _for_ this change, as it adds extra protection by enabling logging on heroku by default.

iwz:
I often use suspenders to create a sandbox app to play around with, and I'm 99% sure this if this is merged, I'll be creating all kinds of heroku apps I didn't intend to create. So, I'm not a fan of auto-creating heroku apps by default.

However, I _am_ a fan of getting the `rails_stdout_logging` gem by default. I'd be cool with just having the heroku gems auto-inserted, and `--heroku=true` meaning to create apps. Or, maybe splitting these into two, like `--create-apps=true` or something.

tute:
> However, I am a fan of getting the rails_stdout_logging gem by default. I'd be cool with just having the heroku gems auto-inserted, and --heroku=true meaning to create apps.

This idea sounds perfect to me.

graysonwright:
What if Suspenders asked you whether you want to use Heroku or not? Have we thought about making Suspenders interactive?

Another option to satisfy both parties would be having Heroku read an environment variable if it's present. People could choose whether new apps use Heroku or not, on a per-system basis.

mike-burns:
I've no interest in an interactive app. Feel free to add a `-i` flag if you're really into not passing `--herolu true`, though.

mcmire:
Okay, so just to understand, we're talking about adding the Heroku gems inserted into the Gemfile always, and then if you want to make a Heroku app, you still need to use `--heroku true`? Even though it doesn't fully solve the issue here, I guess I could accept this for now.

For the record, this is yet another perfect case for generators. I think the reason why creating apps on Heroku automatically is desirable is that it's a pain to do so manually, and there's not a simple command for it. It'd be great if you could add it on to an existing Suspenders app. Something like:

```
suspenders new my-app
cd my-app
# ... later ...
suspenders generate heroku-apps
```

At the moment, is it so bad to do this automatically? If you didn't want the apps, then I feel like it'd be easier to get rid of them than add them to begin with.

mike-burns:
> For the record, this is yet another perfect case for generators.

No one is opposed to this. It's just a lot of work.

> I think the reason why creating apps on Heroku automatically is desirable is that it's a pain to do so manually, and there's not a simple command for it.

`heroku apps:create --ssh-git`

mcmire:
@mike-burns Okay, and you just taught me about that option. But to add both staging and production, it's not one command. I don't really feel like copying and pasting this anywhere:

``` shell
heroku create my-app-staging --remote staging --ssh-git
heroku config:add RAILS_ENV=staging RACK_ENV=staging --remote staging
heroku create my-app-production --remote-production --ssh-git
```

bernerdschaefer:
As a possible middle-ground, I toyed around with the idea on another project of having a git-ignored `bin/once` script which would do the stateful, run-once part of setting up a project (creating apps and tokens, configuring CI, etc.).

derekprior:
The idea of creating apps is separate. 

This change makes it so suspenders assumes you will be deploying to heroku eventually and switches the command line option for this from opt-in to opt-out. As far as I know, there's no real danger is configuring for heroku and deploying elsewhere, but there is danger in configuring for non-heroku and deploying to heroku (no logs).

I think we're straying a bit from what's proposed here into stuff that could be follow-on changes or discussions.

mcmire:
> The idea of creating apps is separate.

But that's one of the things that `--heroku true` does... isn't it?

tute:
@gabebw @derekprior how would you guys feel with making suspenders set up the new Rails app as "deployable to Heroku", keep the flag off by default, and changing its behavior to create the applications in Heroku?

tute:
@gabebw @derekprior how would you guys feel with making suspenders set up the new Rails app as "deployable to Heroku", keep the flag off by default, and changing its behavior to create the applications in Heroku only?

derekprior:
sounds good to me.

tute:
> > how would you guys feel with making suspenders set up the new Rails app as "deployable to Heroku", keep the flag off by default, and changing its behavior to create the applications in Heroku only?
> sounds good to me.

@gabebw would you work on this?

graysonwright:
@gabebw that sounds great.

mike-burns:
Of this method: https://github.com/thoughtbot/suspenders/blob/77312c9629535e3ccb279bc7316e77ace4ed0fb4/lib/suspenders/generators/app_generator.rb#L168-L182

I'd be pretty excited to have these be the default:

```
provide_review_apps_setup_script
create_heroku_pipelines_config_file
provide_deploy_script
configure_automatic_deployment
```

And pretty disappointed to have this on by default:

```
set_heroku_serve_static_files
set_heroku_remotes
set_heroku_rails_secrets
set_heroku_application_host
create_heroku_pipeline
```

In general I want to be able to create and destroy apps with suspenders without side effects -- especially things that hit external APIs.

tute:
@sebarmano started work on https://github.com/thoughtbot/suspenders/pull/604#issuecomment-215756219. Seba, keep in mind that Rails 5 in Heroku will need less setup: https://blog.heroku.com/archives/2016/5/2/container_ready_rails_5. We might be able to simplify code, particularly around the Heorku specific gems.

sebarmano:
Great @tute . I'm finishing the last details of the changes and tests based on @mike-burns comment. As soon as I have them I'll submit the PR to start the code review and start discussing what can be simplified for Rails 5.

tute:
Closing in favor of https://github.com/thoughtbot/suspenders/pull/756, where we implement https://github.com/thoughtbot/suspenders/pull/604#issuecomment-143247939.

tute:
Thank you all for your input.

PR #605, opened 2015-08-28
Add foreman as development dependency

tute:
Avoids errors when trying to start foreman in development environment.

### DIFF
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -27,6 +27,7 @@ gem "title"
 gem "uglifier"
 
 group :development do
+  gem "foreman"
   gem "spring"
   gem "spring-commands-rspec"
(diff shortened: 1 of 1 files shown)


dgalarza:
You may want to update the README that's generated for the project as well as it explicitly states that foreman is not a development dependency intentionally right now.

https://github.com/thoughtbot/suspenders/blob/master/templates/README.md.erb#L19-L20

That being said, I wonder if we should be keeping that intentional decision. Not everyone likes using foreman to manage their development enviornment.

tute:
> I wonder if we should be keeping that intentional decision. Not everyone likes using foreman to manage their development enviornment.

I actually agree with this, and given it's described in the README I'm fine with leaving it out. Thank you!

PR #606, opened 2015-08-28
Enable file-type inference by default

teoljungberg:
This is disabled by default in RSpec 3, which encourages this behaviour:

```ruby
RSpec.describe ThingsController, type: :controller do
end
```

But, with this enabled we can simply write:

```ruby
describe ThingsController do
end
```


tute:
We already have this in https://github.com/thoughtbot/suspenders/blob/master/templates/rails_helper.rb#L18.

teoljungberg:
Oh, hah. In that case - closing

PR #607, opened 2015-08-28
Update to Bitters 1.1

tysongach:
We released Bitters 1.1 a bit ago…so this updates to that!

cc: @whmii Maybe we add updating Suspenders as a checklist for Bitters releases?


tute:
LGTM. I didn't know suspenders included bitters by default.

PR #608, opened 2015-08-31
Remove disable xml params.

DamirSvrtan:
I believe this is no longer an issue - [Vulnerability notes](http://www.kb.cert.org/vuls/id/380039) state that the fixed versions are: 3.2.11, 3.1.10, 3.0.19, 2.3.15, so it should be fixed in the Rails 4.2.xx.

### DIFF
--- a/templates/disable_xml_params.rb
+++ b/templates/disable_xml_params.rb
@@ -1,3 +1,2 @@
-# Protect against injection attacks
-# http://www.kb.cert.org/vuls/id/380039
+# Don't accept XML as input params
 ActionDispatch::ParamsParser::DEFAULT_PARSERS.delete(Mime::XML)


jferris:
We very rarely accept XML as input params. I'm guessing that, when we do, it's easier to just parse the XML ourselves using Nokogiri.

I don't believe automatically parsing every input format is a good default behavior, so I think we should continue overriding this.

DamirSvrtan:
Oh, i thought this was just a security measure

DamirSvrtan:
Ok, no problem, I thought that this was just a relic of the past - a security patch from two years ago that's not needed anymore. Probably should remove the comment from the initializer:

```ruby
-# Protect against injection attacks		
-# http://www.kb.cert.org/vuls/id/380039
```

tute:
> Probably should remove the comment from the initializer:

Sounds good to me, want to update your PR with only that change and an explanatory commit message? Thank you!

DamirSvrtan:
I've removed the previous commits and made a new one.

tute:
Thank you! Merged in.

PR #609, opened 2015-09-02
Use variables for development database host/port

kevintraver:
Allow postgres development database host/port to be set.

### DIFF
--- a/templates/postgresql_database.yml.erb
+++ b/templates/postgresql_database.yml.erb
@@ -2,7 +2,8 @@ development: &default
   adapter: postgresql
   database: <%= app_name %>_development
   encoding: utf8
-  host: localhost
+  host: <%%= ENV.fetch("POSTGRES_DEV_URL", "localhost") %>
+  port: <%%= ENV.fetch("POSTGRES_DEV_PORT", "5432") %>
   min_messages: warning
   pool: <%%= ENV.fetch("DB_POOL", 5) %>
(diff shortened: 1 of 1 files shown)


kevintraver:
Is there a way to entirely ignore the port setup if it is not specified? I assume it should probably stay a socket path if unchanged?

tute:
Why do you need this? Does setting `DATABASE_URL` work in development?

kevintraver:
A use case would be using Docker. Currently you must have a local instance of Postgres running when generating a new app.

pedrosmmoreira:
@kevintraver @tute Just adding my two cents here: Suspenders does make some assumptions, namely of being used in a machine setup with the thoughtbot laptop script. I think the concern regarding Docker is valid, but it would have to be addressed in a wider scope.

I've been going back and forth on the idea of translating Laptop, Dotfiles and Suspenders to a Docker environment for some time, would be happy to collaborate on that :)

tute:
> I've been going back and forth on the idea of translating Laptop, Dotfiles and Suspenders to a Docker environment for some time, would be happy to collaborate on that :)

I believe @jferris has been thinking about this too. I am interested in this project as well.

> @kevintraver @tute Just adding my two cents here: Suspenders does make some assumptions, namely of being used in a machine setup with the thoughtbot laptop script. I think the concern regarding Docker is valid, but it would have to be addressed in a wider scope.

I agree with this. `DATABASE_URL` should be enough, it's what Heroku uses, and so I will close this PR for now.

Thanks for your help!

PR #610, opened 2015-09-03
Set auto-deploy for Heroku suspended apps

pedrosmmoreira:
Previously, the Circle CI config for suspended applications enabled
automatic deployments for all apps. Given that the `bin/deploy` script
is only generated for apps suspended with the `--heroku true` flag, this
led to builds failing on the deployment step due to the script not being
found.

This PR adds the deployment command to `circle.yml` only when an app is
suspended with the Heroku options. There are still manual setup steps
required, such as setting the Heroku API and SSH key but those are
clearly indicated by the error messages on Circle CI and makes
Suspenders defaults more "out-of-the-box" friendly.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -400,6 +400,20 @@ you can deploy to staging and production with:
       run "chmod a+x bin/deploy"
     end
 
+    def configure_automatic_deployment
+      staging_remote_name = heroku_app_name_for("staging")
+      deploy_command = <<-YML.strip_heredoc
+      deployment:
+        staging:
+          branch: master
+          commands:
+            - git remote add staging git@heroku.com:#{staging_remote_name}.git
+            - bin/deploy staging
+      YML
+
+      append_file "circle.yml", deploy_command
+    end
+
     def create_github_repo(repo_name)
       path_addition = override_path_for_tests
       run "#{path_addition} hub create #{repo_name}"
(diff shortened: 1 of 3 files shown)


lib/suspenders/app_builder.rb
```diff
@@ -400,6 +400,18 @@ def provide_deploy_script
       run "chmod a+x bin/deploy"
     end
 
+    def configure_automatic_deployment
+      deploy_command = <<-YML.strip_heredoc
+      deployment:
+        staging:
+          branch: master
+          commands:
```
tute: Do we have access to the staging URL, so we can add here:

```
- git remote add staging git@heroku.com:appname-staging.git
```

pedrosmmoreira: I considered that but since the staging remote is set when the app is generated with the Heroku flag, I thought it was unnecessary.

tute: > I considered that but since the staging remote is set when the app is generated with the Heroku flag, I thought it was unnecessary.

That's locally though, not on the CI host. If the suspended app is generated with the Heroku flag, the deploy script would fail in Circle.

pedrosmmoreira: Ah, yes. That does make sense, I'll get this rebased and corrected

tute:
Thank you! :) :clap:

Issue #602, opened 2015-08-12
Which port should foreman be running on?


LimeBlast:
I'm sorry if this is a silly question, or if I've misunderstood something, but there seems to be some conflicting defaults in the port configuration for foreman/running the dev server.

According to the `.foreman` file, the port should be `7000`, but looking in the `config/environments/developments.rb` file, it suggests a port of 5000, and further still, the default port setting for `HOST` and `ASSET_HOST` in `.sample.env` is 3000.

Should these not all be the same?

tute:
It looks like suspenders build with random ports, but `.sample.env` fixes the Rails default of 3000.

```
./lib/suspenders/app_builder.rb:      @port ||= [3000, 4000, 5000, 7000, 8000, 9000].sample
./templates/sample.env:ASSET_HOST=localhost:3000
```

Random port was introduced in https://github.com/thoughtbot/suspenders/commit/4caa6123 by @croaky, to be able to run multiple suspenders apps with foreman without conflicts. Now that array has 6 elements (which are random anyway), so this doesn't seem like a reliable solution. Also, I personally don't encounter that problem because I typically load two Rails apps at most simultaneously.

I wouldn't mind suspenders defaulting to 3000, but let's see what other users think. I wouldn't mind using the 3000-3100 range ([Chrome considers 6000 unsafe](https://github.com/thoughtbot/suspenders/commit/208a6f09)), and making `.sample.env` smart about the random selection.

LimeBlast:
For now, I've changed all of mine to 3000 (just because that's what I'm used to). If you did stick with a random port, would it be worth making sure the `HOST` and `ASSET_HOST` get set to the same, and maybe update `config/environments/development.rb` to use:

      config.action_mailer.default_url_options = { host: ENV.fetch("HOST") }

:)

tute:
Absolutely. I expected `action_mailer` to already read the proper host, but there might be a bug,

LimeBlast:
:+1:

tute:
Thanks! Closing in favor of PR: https://github.com/thoughtbot/suspenders/pull/588

PR #611, opened 2015-09-05
Use 3000 as default rails server port

delphaber:


tute:
This rollbacks the random port for new suspenders up, while leaving the code configurable (as introduced in https://github.com/thoughtbot/suspenders/commit/4caa6123c83e726d6fcd23956ce9883ca36b1d42). That existing code is still buggy though, as it doesn't take `sample.env` file into account (https://github.com/thoughtbot/suspenders/issues/602#issuecomment-130671649). Should we rollback all the changes from that commit?

delphaber:
Ok I'll complete remove the changes from that commit!

tute:
ping @delphaber. Thank you!

lib/suspenders/app_builder.rb
```diff
@@ -253,7 +253,7 @@ def configure_simple_form
     end
 
     def configure_action_mailer
-      action_mailer_host "development", %{"localhost:#{port}"}
+      action_mailer_host "development", %{ENV.fetch("APPLICATION_HOST")}
```
delphaber: I think we can use APPLICATION_HOST because it's already set to `localhost:3000` in `.sample.env`

templates/bin_setup.erb
```diff
@@ -23,7 +23,7 @@ mkdir -p .git/safe
 
 # Pick a port for Foreman
 if ! grep --quiet --no-messages --fixed-strings 'port' .foreman; then
-  printf 'port: <%= config[:port_number] %>\n' >> .foreman
+  printf 'port: 3000\n' >> .foreman
```
delphaber: We need to overwrite default forman man (5000)

delphaber:
@tute sorry for the lateness. I removed the random port logic completely. I hope it's fine!

lib/suspenders/app_builder.rb
```diff
@@ -253,7 +253,7 @@ def configure_simple_form
     end
 
     def configure_action_mailer
-      action_mailer_host "development", %{"localhost:#{port}"}
+      action_mailer_host "development", %{"localhost:3000"}
```
delphaber: I cannot use `%{ENV.fetch("APPLICATION_HOST")}` because when we run `suspenders app_name` for the first time `.env` does not exists yet (it will be created only executing `bin/setup.sh`).

I don't known if `localhost:3000` is already the Rails default for development environment.

tute:
Thank you! :clap:

Issue #570, opened 2015-05-21
web-console && workers


delphaber:
What do you think about this web-console FAQ?

https://github.com/rails/web-console#why-i-constantly-get-unavailable-session-errors

Should `sample.env`include `UNICORN_WORKERS=1`?

tute:
I didn't hear of this problem with suspenders. I assume it does happen to you?

delphaber:
Yep :(

delphaber:
It happens on pristine suspenders app.

![screenshot 2015-06-05 01 17 55](https://cloud.githubusercontent.com/assets/139284/7996487/fcf01fb8-0b20-11e5-92b7-f5f802c424e6.png)

bf4:
web-console is based on (ripped off of?) [better_errors](https://github.com/charliesome/better_errors) which https://github.com/charliesome/better_errors#unicorn-puma-and-other-multi-worker-servers which also recommends one web server...

personally, if I'm ever trying to use this tool, I switch from foreman to rails s so that it 'just works'

jessieay:
I didn't even realize this lib was using `web-console` - good to know. Seeing `UNICORN_WORKERS=1` in `sample.env` seems like a good idea to me.

tute:
I'm :+1: to `UNICORN_WORKERS=1` in `.sample.env`. Pinging @jferris in case he has seen this or thought of other solutions?

jferris:
We stopped using Better Errors on most projects because of issues like this. If this is officially part of Rails now, I guess we'd better get a single-threaded, single-process development server going.

tute:
As of https://github.com/thoughtbot/suspenders/pull/501, we need to configure Puma to use one worker in development.

tute:
> @jferris: We stopped using Better Errors on most projects because of issues like this. If this is officially part of Rails now, I guess we'd better get a single-threaded, single-process development server going.

It's not officially part of rails, we include `web-console` in suspender's Gemfile, if that's what you are referring to.

tute:
Closing in favor of discussion in PR. Thanks!

PR #612, opened 2015-09-06
Set puma workers to 1 in .sample.env to play nice with web-console

delphaber:
Should close https://github.com/thoughtbot/suspenders/issues/570

### DIFF
--- a/templates/sample.env
+++ b/templates/sample.env
@@ -9,3 +9,4 @@ SMTP_ADDRESS=smtp.example.com
 SMTP_DOMAIN=example.com
 SMTP_PASSWORD=password
 SMTP_USERNAME=username
+WEB_CONCURRENCY=1


tute:
Pinging @seanpdoyle and @jferris for their comments on this.

jferris:
I worry about running a multiprocess server in production (which has the possibility of race conditions and state confusion) but only running a single server locally when developing.

seanpdoyle:
I agree with Joe, but I'm torn.

I usually end up setting this to 1 (and used to do the same with unicorn) while profiling with the local version of https://github.com/newrelic/rpm.

Is there a middle ground? 

Could web-console / newrelic be modified to work with multiple processes (that's obviously out of scope for this PR, but is it worth considering?)?

tute:
> I worry about running a multiprocess server in production (which has the possibility of race conditions and state confusion) but only running a single server locally when developing.

Would we ever reach a point in development or test when we hit a race condition, without actively seeking it? It sounds very unlikely to me, in which case this PR doesn't introduce that potential issue.

seanpdoyle:
@tute @jferris also, `WEB_CONCURRENCY=1` and `MAX_WORKERS=1` resolve some performance / usability issues for `ember-cli-rails` projects:

* https://github.com/thoughtbot/ember-cli-rails/issues/94
* https://github.com/thoughtbot/ember-cli-rails/pull/222
* https://github.com/thoughtbot/dome/pull/19

tute:
Will merge this in unless @seanpdoyle or @jferris object.

@delphaber, can you please rebase your commit on top of latest master, and add context of this discussion into the commit message? Thank you! :)

delphaber:
@tute sure thing :)

delphaber:
Done!

tute:
Thank you!

Issue #418, opened 2014-10-16
Environment variables seem incomplete


lpender:
```
suspenders app --heroku true
cd app
git push staging master
```

ends in

```
KeyError: key not found: "SMTP_ADDRESS"
```

Adding the following config vars makes the heroku deploy build pass:

```
SMTP_ADDRESS=
SMTP_DOMAIN=
SMTP_PASSWORD=
SMTP_USERNAME=
EMAIL_RECIPIENTS=
ASSETS_VERSION=
ASSET_HOST=
```

However, I'm still getting an application error from Heroku when visiting the page.

Thoughts?

jkrmr:
This was helpful: https://github.com/thoughtbot/suspenders/pull/532#issuecomment-86976291

shashankshandilyav:
Could someone guide me what should we do when we have this issue?

PR #613, opened 2015-09-14
Add mandatory environment variables

mib32:
If you not set up any environment variables that are accessed using `fetch`, then the deploy to heroku will throw an error. 
That could by quite frustrating because, say, you added `HOST` variable, it thorns an error, then deploy again, and get error about `ASSET_HOST` and so on. 
I think that would be great if list of mandatory environment variables was present in README.

### DIFF
--- a/templates/sample.env
+++ b/templates/sample.env
@@ -4,3 +4,7 @@ APPLICATION_HOST=localhost:3000
 RACK_ENV=development
 SECRET_KEY_BASE=development_secret
 EXECJS_RUNTIME=Node
+SMTP_ADDRESS=smtp.example.com
+SMTP_DOMAIN=example.com
+SMTP_PASSWORD=password
+SMTP_USERNAME=username


README.md
```diff
@@ -140,6 +140,15 @@ See all possible Heroku flags:
 
     heroku help create
 
+In order to successfully push app to heroku, following environment variables must be set:
+
+* HOST
```
tute: This one is not necessary anymore, after https://github.com/thoughtbot/suspenders/commit/f6f4869f9dbad9104597bc4380330e8311f75d08.

README.md
```diff
@@ -140,6 +140,15 @@ See all possible Heroku flags:
...
+* ASSET_HOST or APPLICATION_HOST
```
tute: `APPLICATION_HOST` is required, and that renders `ASSET_HOST` optional. I'd remove it from this list.

tute:
It's easy to forget updating the docs when the code is changed. What do you think of adding dummy values to https://github.com/thoughtbot/suspenders/blob/master/templates/sample.env, instead of changing this?

mcmire:
I agree with @tute. We have a `.sample.env` that should be responsible for setting any default values for required variables.

mib32:
Amended according to feedback, thanks guys! What do you think now?

mcmire:
`.sample.env` looks good -- what about removing the changes from the README?

mib32:
@mcmire Right, done.

tute:
I'm sorry I made you work more than necessary. It's on purpose that these have no values. See https://github.com/thoughtbot/suspenders/issues/418#issuecomment-64154294:

> These ENV variables are required so they fail fast rather than down the road in a less obvious way.

When you try to deploy, `fetch` will fail with a helpful error message, forcing you into setting up SMTP. Do you think there is a better way of handling this?

mcmire:
(Ah I was thinking that someone had made a comment on that in the past, I didn't look very far though. Sorry as well)

mib32:
Well, i think that this branch does not interfere with this purpose, because the deploy will still fail, right? 
`sample.env` filled with that values just makes you sure that "ok, after i will set SMTP up, there would be no other things that'll hold me back". Imagine if somewhere in the future you will be `fetch`ing more variables.

tute:
> Well, i think that this branch does not interfere with this purpose, because the deploy will still fail

Right, merged in. Thank you!

PR #614, opened 2015-09-16
Cast ENV variables in database.yml to Integer

delphaber:
If you define e.g. `MAX_THREADS=1` in `.env` you get

```bash
(erb):17:in `each': Cannot load `Rails.application.database_configuration`:
comparison of Fixnum with String failed (ArgumentError)
```

That's because in `database.yml` this line:

```yml
  pool: <%= [ENV.fetch("MAX_THREADS", 5), ENV.fetch("DB_POOL", 5)].max %>
```
will be evaluated as

```yml
pool: <%= ["1", 5)].max %>
```

Casting ENV variables to Integer(), as in `puma.rb`, should fix the problem.

### DIFF
--- a/templates/postgresql_database.yml.erb
+++ b/templates/postgresql_database.yml.erb
@@ -4,8 +4,8 @@ development: &default
   encoding: utf8
   host: localhost
   min_messages: warning
-  pool: <%%= ENV.fetch("DB_POOL", 5) %>
-  reaping_frequency: <%%= ENV.fetch("DB_REAPING_FREQUENCY", 10) %>
+  pool: <%%= Integer(ENV.fetch("DB_POOL", 5)) %>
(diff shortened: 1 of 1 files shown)


delphaber:
Oh, I tested it locally with `foreman start` and `rails s`

tute:
Nitpick, what about calling `.to_i` instead of `Integer`?
Looks good to me, thank you!

delphaber:
I saw `Integer()` in `puma.rb` which come from heroku.

```ruby
"".to_i # => 0
nil.to_i # =>  0
```
while `Integer("")` and `Integer(nil)` both raise an error.

I think that the guys at heroku wanted a less magical casting

mcmire:
I think `Integer` makes sense here -- it'll raise an error if given incorrect input, which is probably what you want most of the time.

tute:
Indeed; merged in. Thank you!

PR #615, opened 2015-09-16
Fix HOST/APPLICATION_HOST env error

adarsh:
Reason for change:
* In 9d27905884540e44109cc66630062101307f150a, `HOST` was replaced for  `APPLICATION_HOST`
* Some instances were missed, leading to errors on deploy from a fresh build.

What the change was:
* Replace remaining `HOST` instances with `APPLICATION_HOST`
* Add regression spec

Original commit: https://github.com/thoughtbot/suspenders/commit/9d27905884540e44109cc66630062101307f150a

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -86,7 +86,7 @@ module Suspenders
       config = <<-RUBY
 
   # Ensure requests are only served from one, canonical host name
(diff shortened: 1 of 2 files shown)


mike-burns:
All of this seems right to me. Thanks for catching this.

tute:
Thank you!

PR #618, opened 2015-09-25
Configure capybara-webkit to block unknown URLs

### LINKED ISSUE #550: Configure Capybara Webkit to block unknown URLS
If `--capybara-webkit`, add to `spec/spec_helper.rb`:

```ruby
RSpec.configure do |config|
  config.before(:each, js: true) do
    page.driver.block_unknown_urls
  end
end
```

tute:
Removes capybara-webkit minimum version specification in Gemfile,
relying on bundler to install latest available release.

See: http://blog.arkency.com/2015/02/upgrade-capybara-webkit-to-1-dot-4-and-save-your-time/

[fixes #550]

cc @jferris @croaky for reviews. Thank you!

### DIFF
--- a/templates/rails_helper.rb
+++ b/templates/rails_helper.rb
@@ -14,6 +14,9 @@ module Features
 end
 
 RSpec.configure do |config|
+  config.before(:each, js: true) do
+    page.driver.block_unknown_urls
+  end
   config.include Features, type: :feature
   config.infer_base_class_for_anonymous_controllers = false
   config.infer_spec_type_from_file_location!
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -43,7 +43,7 @@ group :development, :test do
 end
 
 group :test do
-  gem "capybara-webkit", ">= 1.2.0"
+  gem "capybara-webkit"
   gem "database_cleaner"
   gem "formulaic"
   gem "launchy"


templates/rails_helper.rb
```diff
@@ -18,6 +18,10 @@ module Features
   config.infer_base_class_for_anonymous_controllers = false
   config.infer_spec_type_from_file_location!
   config.use_transactional_fixtures = false
+
+  config.before(:each, js: true) do
+    page.driver.block_unknown_urls
+  end
```
croaky: I think the configuration block is alphabetized.

croaky:
This is the more modern version of https://robots.thoughtbot.com/speed-up-javascript-capybara-specs-by-blacklisting-urls ? Should we update that blog post?

croaky:
is there still a capybara_webkit flag that can be disabled for Suspenders runs? Will these lines not be added if that flag is set?

croaky:
If we need to take into account the flag, maybe we use a similar "separate file in spec/support" pattern such as https://github.com/thoughtbot/suspenders/blob/master/templates/database_cleaner_rspec.rb ?

templates/rails_helper.rb
tute: > I think the configuration block is alphabetized.

Done.

tute:
> This is the more modern version of https://robots.thoughtbot.com/speed-up-javascript-capybara-specs-by-blacklisting-urls ? Should we update that blog post?

Should we have an "Oudated" lane for this in https://trello.com/b/sWEy2tVb/blog-editorial-calendar? cc @gabebw 

> is there still a capybara_webkit flag that can be disabled for Suspenders runs? Will these lines not be added if that flag is set?

There is not such flag AFAIK.

Thanks for your review!

gabebw:
> Should we have an "Oudated" lane for this in https://trello.com/b/sWEy2tVb/blog-editorial-calendar? cc @gabebw


I think we can present it as a regular blog post, like a card that says "Update 'Speed up Capybara JS specs' post", since it'll go through almost exactly the same process.

I am very in favor of this PR.

tute:
> I think we can present it as a regular blog post

https://trello.com/c/krK6TXDx/745-update-speed-up-capybara-js-specs-post

tute:
Thank you!

PR #620, opened 2015-10-07
Provide Shoulda Matchers config

pedrosmmoreira:
The recent release of Shoulda Matchers 3.0 introduces some changes,
namely requiring configuration in order to operate.
The absence of the file leads to hard to decipher errors: they indicate
methods are missing and can lead to some frustration trying to
understand what is necessary to get Shoulda to behave correctly.

This PR introduces a build step to provide a config file for in order
for it to be enabled "out-of-the-box". It opts in to enable interaction
with all of Rails, as this seems the principle of least surprise.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -197,6 +197,13 @@ end
       copy_file 'database_cleaner_rspec.rb', 'spec/support/database_cleaner.rb'
     end
 
+    def provide_shoulda_matchers_config
+      copy_file(
+        "shoulda_matchers_config_rspec.rb",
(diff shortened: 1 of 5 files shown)


mcmire:
Thanks for this! This looks good to me.

delphaber:
I think `require: false` in Gemfile and `require "shoulda_machers` in `rails_helper.rb` could be removed.

Source: https://robots.thoughtbot.com/shoulda-matchers-3-0

mcmire:
You're right, it totally can.

pedrosmmoreira:
Yep, I'll remove the require and push. Thanks!

tute:
Thank you! :)

PR #624, opened 2015-10-15
Use pry-byebug instead of just byebug

mcmire:
Why use just Pry or byebug when you can have the power of both?

This lets us use `next`, `continue`, `up`, etc. instead of a Pry
session; conversely, it gives us a full IRB environment inside of a
byebug session.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+HEAD
+
+* Add ability to use byebug navigation commands inside of Pry using the
+  `pry-byebug` gem
+
 1.34.0 (November 15, 2015)
 
 * Fix `block_unknown_urls` deprecation warning with capybara_webkit when running
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -35,10 +35,10 @@ group :development, :test do
   gem "awesome_print"
   gem "bullet"
   gem "bundler-audit", require: false
-  gem "byebug"
   gem "dotenv-rails"
   gem "factory_girl_rails"
   gem "i18n-tasks"
+  gem "pry-byebug"
   gem "pry-rails"
   gem "rspec-rails", "~> 3.3.0"
 end


jferris:
Does this make it harder to use regular byebug commands like `n` and `c`?

mcmire:
No it doesn't -- those commands are still available. It's just that you get a full IRB environment as well, instead of a half-one.

deivid-rodriguez:
Actually, those aliases are not enabled by default anymore. See https://github.com/deivid-rodriguez/pry-byebug#matching-byebug-behaviour

mcmire:
Okay -- is it worth it to add a `~/.pryrc` to Suspenders?

tute:
@mcmire I don't have an opinion on this (never used byebug yet). Do you think we should merge this, work more on it, or discard it?

mcmire:
As it is, this would inconvenience people who are used to using byebug on its own. However, it looks like there isn't anything we can do in this PR to fix that problem. Actually there isn't anything we can do in Suspenders at all. We can certainly add a `~/.pryrc` to our dotfiles, but it'd mean that people who aren't using the dotfiles will have to add a `~/.pryrc` themselves.

I don't know. I think it's still worth it to merge this in, seeing that we haven't gotten much feedback on this.

gylaz:
What about adding `.pryrc` to Suspenders?

mcmire:
Is that something that belongs in Suspenders? Should `dotfiles` be the one adding... oh, you mean add it to the project directory. Oh, that makes more sense :)

Okay, ping me again on this tomorrow, I'll work on finishing this up.

gylaz:
I've just tried out `pry-byebug` on my recent project and `c`, `s`, and `n` work fine with `debugger` breakpoint.

gylaz:
I haven't noticed anything different using `pry-byebug` vs `byebug`. What does `it gives us a full IRB environment inside of a byebug session.` mean?

mcmire:
If you use `debugger` then you'll be dropped into a byebug session, not Pry. (Use `binding.pry` instead.)

`byebug` parses what you type on the prompt just like IRB, but the parser is limited and doesn't allow full Ruby expressions. I can't remember which expressions I had trouble with, but I feel like I was trying to do something pretty basic and byebug wouldn't let me.

I can certainly try switching to byebug for a while and seeing how that goes.

gylaz:
> If you use debugger then you'll be dropped into a byebug session, not Pry. (Use binding.pry instead.)

Thanks for the suggestion, but I prefer Byebug (and its `debugger` alias). I haven't enjoyed Pry as much mostly due to it not supporting navigation out of the box. I was under the impression this PR would benefit both Byebug and Pry users, which caused my comment that I haven't noticed any benefit for Byebug.

jferris:
> If you use `debugger` then you'll be dropped into a byebug session, not Pry.

If that's the case, it won't bother me if the debugger shortcuts aren't available from `binding.pry`, because I can just continue to use `debugger` and won't be affected.

For what it's worth, I consider `debugger` and `binding.pry` very different tools, and I'm not sure when you'd want to use them together. I use byebug when something is behaving differently than I expected (debugging hat) and I use pry when I'm exploring an object I don't understand (explorer hat). However, if people like to use the tools together, it won't bother me if it doesn't break my workflow for byebug.

deivid-rodriguez:
>> If you use debugger then you'll be dropped into a byebug session, not Pry.
>
> If that's the case, it won't bother me if the debugger shortcuts aren't available from binding.pry, because I can just continue to use debugger and won't be affected

Yep, that's the case.

Maybe this goes a bit off-topic but I've actually been [thinking of providing the opposite](https://github.com/deivid-rodriguez/pry-byebug/issues/45#issuecomment-136197572) too. Provide a different entrypoint to `pry-byebug` (say, the `pry-byebug` method), so it doesn't break the workflow of people who want to "use the explorer hat" only (by using plain `binding.pry`).

mcmire:
> Thanks for the suggestion, but I prefer Byebug (and its debugger alias). I haven't enjoyed Pry as much mostly due to it not supporting navigation out of the box.
> ...
> I was under the impression this PR would benefit both Byebug and Pry users

Ah... I understand... I misworded the commit message.

Let's try this again :) This PR is intended for Pry users who like Pry but want the navigation shortcuts that Byebug provides. Since `pry-byebug` depends on `byebug`, it doesn't remove that dependency, so you are still free to use Byebug if you like.

tute:
This is great and I'd like it merged! :)

mcmire:
Done.

PR #625, opened 2015-10-16
Don't load i18n tasks in production

derekprior:
This gem is used in development and test and is not required in
production. At minimum, loading it in production is a memory use issue
(however small), but it likely also adds overhead to every translation
to track used keys.

This could explain some complaints we've seen of slow view rendering on
apps that I know to have been recently generated by suspenders.


derekprior:
Having investigated i18n tasks some more, it appears to be static, not dynamic analysis so it's not likely to be a runtime performance impact, but it's still code we don't need to load in production.

gabebw:
Good catch. This makes sense to me. :+1:

tute:
:+1:

mcmire:
:+1:

tute:
@derekprior I don't see this in master. Did you merge it?

derekprior:
Yeah, think I cleaned up my branches before pushing master.

PR #627, opened 2015-10-16
Remove email_validator

derekprior:
If we need to validate email addresses we can add this gem manually.
Having it in suspenders saves us no setup, costs us a small amount of
runtime memory, and increases the amount of code we're running for no
real reason.


gabebw:
> If we need to validate email addresses we can add this gem manually.

I think that argument applies to every gem in suspenders: "If we need X, we can add gem Y manually".

Do we generally not use this at all?

tute:
> Do we generally not use this at all?

I rely on clearance's validator.

derekprior:
> I think that argument applies to every gem in suspenders: "If we need X, we can add gem Y manually".

I'm just getting started :grinning:. 

> Do we generally not use this at all?

I've not once used it directly. The intent of suspenders shouldn't be "let's put gems in here that you might want to use" but rather, "let's configure an app with out best practices". If you need to validate email addresses, email_validator is a fine choice, but that knowledge can live somewhere else (software matrix?)

jferris:
I think the highest value in Suspenders is from bundling gems we use on basically every project combined with our recommended configuration. The configuration part is key to me; if you just need to add a gem to the `Gemfile`, that's not a big deal. However, getting our production configuration setup on every project is a waste of time.

In this case, it seems as though the validator is rarely used and easy to add when necessary.

gabebw:
>  The configuration part is key to me; if you just need to add a gem to the Gemfile, that's not a big deal. However, getting our production configuration setup on every project is a waste of time.

That's a great point. :+1:

mike-burns:
See also: #511.

mcmire:
I forgot this gem existed, although I know that I've used it. However, given that I don't use it that often, I think it'd be fine if we removed it.

Should we comment it out in the Gemfile instead of just outright removing it?

gabebw:
I think we should remove it entirely, or not. Comments are a slipper slope.

mcmire:
Why are they a slippery slope? They wouldn't really act as comments in this case -- they'd be more like suggestions on what to add later. Is there a better way to offer these suggestions?

tute:
> Why are they a slippery slope? They wouldn't really act as comments in this case -- they'd be more like suggestions on what to add later. Is there a better way to offer these suggestions?

We don't suggest other gems for common tasks like user authentication or file upload. To me, choosing to use `email_validator` is analogous to choosing to use `clearance`: we add it to the Gemfile after the application has been generated.

bernerdschaefer:
Perhaps @croaky can weigh in on the [original suspenders feature][feature], or the [commit] which added the gem here.

[feature]: https://github.com/thoughtbot/suspenders/commit/b91ea65c1885e3198fd05ae4622ef09fe00b6143
[commit]: https://github.com/thoughtbot/suspenders/commit/4cf30e62ff1789395b8c932a94c3e7a9bcb0b81f

mcmire:
Okay, so how do we know that's a possible option to use? Is this where our internal spreadsheet of solutions comes into play?

jferris:
> Why are they a slippery slope? They wouldn't really act as comments in this case -- they'd be more like suggestions on what to add later. Is there a better way to offer these suggestions?

I don't think it's realistic to make a suggestion for every useful gem we've found in the `Gemfile`. Why would we pick email validation as one of the special cases to call out?

mcmire:
I would say that it's a problem that we've run into often enough that we thought it would be useful to put in the Gemfile. Now I agree with you that the Gemfile is probably not the right place for a list of gems that we've used to solve various problems but I think we should have *some* place for that, so that's why I was asking.

croaky:
> I forgot this gem existed, although I know that I've used it. However, given that I don't use it that often, I think it'd be fine if we removed it.

When you have a Rails app, does it typically have an email address on one of the models? In cases where it does, do you want to validate it? When you do, do you use this gem or something custom?

I can see this not coming up very often in terms of time spent on it, but I don't think it's necessarily meant to save setup/implementation time.

I often see custom regexes for this validation in apps. My thinking is the more we use this gem, the less likely we are to encounter strange differences in those regexes, or accidentally introduce a security issue. Can focus our efforts on supporting one library.

jferris:
> When you have a Rails app, does it typically have an email address on one of the models? In cases where it does, do you want to validate it? When you do, do you use this gem or something custom?

The only model in most of my applications with an email has been `User`, which is provided and validated by Clearance or another library, so I generally haven't had to write email validations.

When I do have an email, I worry about over-validating the email address. I find it very annoying as a user when I can't put a `+` or a newer TLD in an email field. My confidence in any Ruby library to allow all valid emails and disallow any invalid emails is low given the difficulty of the problem.

In cases where it's important to have a valid email, I'd rather validate addresses by sending an email to them, as that's necessary regardless of the format: the validator can't tell you if it's a live address and the user controls it.

derekprior:
> I often see custom regexes for this validation in apps. My thinking is the more we use this gem, the less likely we are to encounter strange differences in those regexes, or accidentally introduce a security issue. Can focus our efforts on supporting one library.

I don't think it follows that simply having an unused gem in the gemfile is enough to prevent people who would be susceptible to such a mistake from making it. It requires that they remember a gem exists in the gemfile, then go look up how to use it. I'm not sure how much that differs from remembering a gem exists on rubygems and looking up to how to use it other than the implicit thoughtbot suspenders endorsement.

I think those endorsements are valuable, but our template is not the right place for them.

mcmire:
> I think those endorsements are valuable, but our template is not the right place for them.

Cool, then I'm fine with this :+1:

derekprior:
It sounds like people are okay with not having this enabled by default and that we've identified a desire to have library preferences that aren't appropriate for suspenders identified. I think the [software matrix](http://tbot.io/matrix) (sorry, internal only) is the start of that.

Is this good to merge?

tute:
> It sounds like people are okay with not having this enabled by default and that we've identified a desire to have library preferences that aren't appropriate for suspenders identified. I think the software matrix (sorry, internal only) is the start of that.
> Is this good to merge?

Ready for merge! :ship:

PR #628, opened 2015-10-16
Move refills to the development group

derekprior:
Refills is not a runtime dependency. The gem is used strictly as a
generator in development mode. It's memory footprint is not
insignificant, so not loading in production will give us more headroom
on Heroku.


gabebw:
:+1:

tute:
:+1:

derekprior:
According to derailed, this saves 18MB. That's a good deal of memory across 3 processes.

mcmire:
Nice!

PR #632, opened 2015-10-21
Remove i18n-tasks

jferris:
Although this catches some actual mistakes, it creates too much noise to
be worthwhile. It looks for unused and missing translations using static
analysis, so it gets confused by dynamic keys and frequently needs to be
configured to ignore entire namespaces.


BlakeWilliams:
:+1:

tute:
I've been using and liking it, and I see continuing using it in future projects.
But if people don't, I'm not against adding it myself.

iancanderson:
It also depends on `term-ansicolor`, which is GPL-licensed, potentially making clients' legal teams unhappy.

sikachu:
You might want to fix the test case.

thorncp:
:+1: I think this should be opt-in.

gylaz:
I've enjoyed using this gem. It has caught missing, or orphan translations for me (and likely for others on the project). I think it does a good job at keeping the translation files clean and up-to-date.

tute:
> I think this should be opt-in.

Do we use I18n by default for Rails apps at thoughbot, or do we have a portion of non i18n'd applications. If the former, I think it should be opt out.

mcmire:
I agree that i18n-tasks is a little finicky -- half the time I even end up having to futz with it. However, I don't think that's a reason for removing it outright -- honestly, I think if there are issues with the gem, we should spend time to properly fix it.

derekprior:
There's a non-trivial support cost to this gem in #ruby. It comes up many times a month and I have to assume that before people ask about it in Slack they are spending a good deal of time trying to figure things out on their own.

tute:
I'm not against removing it, and adding it as we see fit. I have been seeing more false positives than I'd like to, and I have seen its value drop for me.

jferris:
I still feel as though we should remove this. It's caught one or two issues for me, but the issues it catches are not bugs or significant maintenance burdens, and I occasionally have to spend 10-15 minutes arguing with it.

If somebody really wants this on their project, I think it would be worth introducing a generator like the one @mike-burns is adding in #676 so that people can opt in.

tute:
:+1:

drapergeek:
:+1:

joshuaclayton:
:+1: lots of frustrations/configuration that goes along with it, and even then, it often doesn't play nicely.

PR #633, opened 2015-10-22
Fail the build if there are any violations found

teoljungberg:
We've been using this in Hound and enjoying the benefit it gives.

A PR should not be merged consciously with a failed build. And
inconsistent code should fall under that umbrella.

### DIFF
--- a/templates/hound.yml
+++ b/templates/hound.yml
@@ -1,4 +1,6 @@
 # See https://houndci.com/configuration for help.
+fail_on_violations: true
+
 coffeescript:
   # config_file: .coffeescript-style.json
   enabled: true


derekprior:
How does this work with incremental pushes to a branch? For instance, I've seen comments from hound that can be remedied by changes that don't touch the line that was commented on. Sometimes the original comment sticks around in that case. Would Hound still consider that a failure?

Beyond that question, I feel like there's a reason hound comments inline beyond convenience. Like any other review comment, it's a discussion. I merge code with hound violations sometimes, but I'd almost never merge a failing build.

I'd lean towards not defaulting to this, but I don't feel super strongly.

jferris:
Does it make sense for us to try this setting out on a Rails application before configuring it in Suspenders?

teoljungberg:
@jferris - As mentioned in the PR description, we have been using it in Hound for ~3 months.

teoljungberg:
@derekprior - Hound updates the status on each new push to the PR. So the build will be marked as a success if the violations are removed.

mjankowski:
> As mentioned in the PR description, we have been using it in Hound for ~3 months.

When you wind up in situations where you don't want to fix some of the issues, do you just merge anyway with a "failing build", or do you stop and do some work in another branch to first remove the rule that you don't care about?

teoljungberg:
I've fixed the violation, or made a PR to remove the rule if I strongly disagree with it.

mcmire:
I don't think this is a good idea. Why? Because I think what this is saying is "Hound knows better than you do, always", and that isn't always true. Sometimes there are cases in which I choose to disagree. Our guidelines aren't rules, and they shouldn't be. I think that code comments on PR's are enough of a check, to be honest (and perhaps more noticeable to your coworkers than a failing build would be).

Also, I don't actually think that inconsistent code deserves a failing build. The code may not be written the "right way", but that doesn't mean it doesn't work.

tute:
I also disagree with this change. When I write or read style comments they are typically paired with a "not important", or "stylistic", or "nitpick", and that's how I feel about them. I disagree with every person at thoughtbot in at least one trivial opinion:

* `inject` or `reduce`
* allow for an 81 character line in a particular case
* [multi line macros](https://github.com/thoughtbot/guides/pull/361/files)
* unnecessary spacing for `schema.rb` or [similarly laid out files that we write](https://github.com/thoughtbot/crowdwise/blob/f6d3665ae89c65d6110578281ef51ea12b34967f/app/models/life_insurance_calculator_form.rb)
* etc.

We all want consistent code bases, so we pay attention to hound and address most of its comments. But I don't think Hound should say "There! Your build is red now! You better fix it!". It should be a "why don't you make this line of code more in line with thoughtbot guidelines?".

And when we disagree with it, we can configure it not to complain on that again, but not *right now* so my build is green.

Red PRs because of Hound would feel a bit like a hassle to me, and I don't think we need them to take care of our style.

teoljungberg:
This didn't seem like a popular change, I'll close it

PR #635, opened 2015-10-22
Add quiet_assets as development dependency

tute:
I typically search the logs to find parameters, response codes and
stacktraces. Although we don't need to see what assets have been loaded,
those logs take up more space than the aforementioned ones.

`quiet_assets` hide the assets logs in development, making the log for
the landing page on my current suspenders project go from 64 lines
detailing assets, to 13 relevant lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ And development gems like:
   pre-loading
 * [Web Console](https://github.com/rails/web-console) for better debugging via
   in-browser IRB consoles.
+* [Quiet Assets](https://github.com/evrone/quiet_assets) for muting assets
+  pipeline log messages
 
 And testing gems like:
 
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -27,6 +27,14 @@ module Suspenders
       inject_into_class "config/application.rb", "Application", config
     end
 
+    def configure_quiet_assets
+      config = <<-RUBY
+    config.quiet_assets = true
+      RUBY
+
+      inject_into_class "config/application.rb", "Application", config
+    end
+
     def provide_setup_script
       template "bin_setup.erb", "bin/setup", force: true
       run "chmod a+x bin/setup"
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -86,6 +86,7 @@ module Suspenders
       build :provide_dev_prime_task
       build :configure_generators
       build :configure_i18n_for_missing_translations
+      build :configure_quiet_assets
     end
 
     def setup_test_environment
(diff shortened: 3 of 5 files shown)


mike-burns:
I'm surprisingly OK with this.

mike-burns:
For clarity and ease of debugging, how about adding this to `config/application.rb`:

```ruby
config.quiet_assets = true
```

That way we can quickly jump in there when we think something is fishy about our assets, and we can find it with a grep because it's named `quiet_assets`.

--

This gem is dual licensed, MIT and GNU GPL. That's cool, right?

tute:
> For clarity and ease of debugging, how about adding this to config/application.rb (...)

:+1:, on it

> This gem is dual licensed, MIT and GNU GPL. That's cool, right?

It looks like not cool? Their definition seems to state clearly that suspenders and applications on top of it count as derivatives: https://www.gnu.org/licenses/gpl-faq.html#GPLStaticVsDynamic

Other gems that might conflict:

## GPLv2

* diff-lcs 1.2.5 (http://diff-lcs.rubyforge.org/) - Diff::LCS computes the difference between two Enumerable sequences using the McIlroy-Hunt longest common subsequence (LCS) algorithm
* mime-types 2.6.1 (https://github.com/mime-types/ruby-mime-types/) - The mime-types library provides a library and registry for information about MIME content type definitions
* term-ansicolor 1.3.2 (http://flori.github.com/term-ansicolor) - Ruby library that colors strings using ANSI escape sequences

## GPLv3

* bundler-audit 0.4.0 (https://github.com/rubysec/bundler-audit#readme) - Patch-level verification for Bundler

## gpl

* pg 0.18.2 (https://bitbucket.org/ged/ruby-pg) - Pg is the Ruby interface to the {PostgreSQL RDBMS}[http://www.postgresql.org/]
* quiet_assets 1.1.0 (http://github.com/evrone/quiet_assets) - Turns off Rails asset pipeline log.

geoffharcourt:
@tute, I could have sworn I suggested this last year, but I clearly haven't. If the licensing isn't prohibitive, this is one of the first changes I make to any new app. :+1:

mcmire:
Yes :100:

jferris:
@tute Quiet Assets is provided under both licenses, which means you can choose which license to agree to.

http://programmers.stackexchange.com/questions/211096/what-does-released-under-mit-and-gpl-license-mean

geoffharcourt:
I didn't realize you can do that.

I've had a number of license questions recently. Some of the reading on the topic only served to cloud the issue further. Would there be benefit in doing a license clinic or workshop on a Friday so that folks who know these well can spread that knowledge.

tute:
> Quiet Assets is provided under both licenses, which means you can choose which license to agree to.

aha, TIL. Thanks!

tute:
Will add quiet_assets in the list in the README, and merge it in soon. Thank you all.

tute:
@gabebw do you know how does this compare with https://github.com/roidrage/lograge?

PR #636, opened 2015-10-22
Fixes deprecation warning from capybara-webkit

aaronrenner:
When enabling javascript testing in a new app, the developer would receive the following message in their rspec output:

    [DEPRECATION] block_unknown_urls is deprecated. Please use Capybara::Webkit.configure instead:

      Capybara::Webkit.configure do |config|
        config.block_unknown_urls
      end

    This option is global and can be configured once (not in a `before` or `setup` block).

This change creates spec/support/capybara_webkit.rb that contains the configuration and removes code in spec_helper.rb triggering the deprecation warning.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -246,6 +246,10 @@ end
       copy_file 'action_mailer.rb', 'spec/support/action_mailer.rb'
     end
 
+    def configure_capybara_webkit
+      copy_file "capybara_webkit.rb", "spec/support/capybara_webkit.rb"
+    end
+
     def configure_time_formats
       remove_file "config/locales/en.yml"
       template "config_locales_en.yml.erb", "config/locales/en.yml"
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -103,6 +103,7 @@ module Suspenders
       build :configure_i18n_for_test_environment
       build :configure_i18n_tasks
       build :configure_action_mailer_in_specs
+      build :configure_capybara_webkit
     end
 
     def setup_production_environment
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -43,6 +43,10 @@ RSpec.describe "Suspend a new project with default configuration" do
     expect(File).to exist("#{project_path}/spec/support/action_mailer.rb")
   end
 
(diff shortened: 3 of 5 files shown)


tute:
Thanks!

Should we append `Capybara::Webkit.configure(&:block_unknown_urls)` to `spec/rails_helper.rb` instead of creating a new file?

aaronrenner:
I'm open to suggestions. Having a separate file and `Capybara::Webkit.configure` in the block form would make it easier for developers to add additional configuration options in their individual projects. However, I'm not sure how often people tweak the `Capybara::Webkit` config.

tute:
> I'm not sure how often people tweak the Capybara::Webkit config

I'll appreciate @jferris input on this.

mcmire:
I think we should start splitting out rails_helper into separate files. That's usually one of the first things I do on a new project. I like to keep that file pretty minimal. So I think putting this in a separate file makes sense.

If we do this, then I think it's also worth it to move the `Capybara.javascript_driver = :webkit` line to this file too.

jferris:
I think using a separate file is cool, but I don't really have any concrete reason why. I just like putting config for various gems in separate files.

I don't generally configure capybara-webkit very much.

tute:
I personally like everything in the rails_helper file, but that's just habit. I can easily switch to `spec/support/` files. I'm not against this changeset as it is.

tute:
Alternative solution: https://github.com/thoughtbot/suspenders/pull/641

geoffharcourt:
As I wrote on #641, I approve of moving this into its own file in `spec/support`. Can you move back to the block form? As soon as you make any changes in the non-block form you get a noisy diff, even for a small change.

tute:
> Can you move back to the block form?

It is using the block form. Will merge soon!

tute:
> If we do this, then I think it's also worth it to move the `Capybara.javascript_driver = :webkit` line to this file too.

Can you add that to your commit, @aaronagray? Thank you!

tute:
> If we do this, then I think it's also worth it to move the `Capybara.javascript_driver = :webkit` line to this file too.

Can you add that to your commit, @aaronrenner? Thank you!

aaronrenner:
@tute I just updated `capybara_webkit.rb` to include `Capybara.javascript_driver = :webkit`. Let me know if you'd like me to update anything else.

tute:
Thank you!

tute:
:clap:

Issue #637, opened 2015-10-23
Define -v --version options to suspenders command


tute:
I want to be able to know in the command line what version of suspenders am I running. Now I get:

```
% suspenders --version
No value provided for required arguments 'app_path'
```

I'd like to see the version number.

tute:
Closing in favor of PR: https://github.com/thoughtbot/suspenders/pull/638

PR #638, opened 2015-10-23
Added version command to CLI

Fourfingerz:
In the latest version 1.33.0, when you type 'suspenders --version' in the CLI, Suspenders responds with the message 'No value provided for required arguments 'app_path'.

I fixed this by adding '-v' and '--version' commands in bin/suspenders.rb.  Those commands will now return the proper suspenders version in the CLI.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+1.34.0 (unreleased)
+
+* Suspenders command line responds to `-v` and `--version` options
+
 1.33.0 (October 23, 2015)
 
 * Add `quiet_assets` as development dependency
--- a/bin/suspenders
+++ b/bin/suspenders
@@ -11,6 +11,11 @@ if ['create', '--create'].include? ARGV[0]
   puts "[WARNING] the suspenders create argument is deprecated. Just use `suspenders #{ARGV.join}` instead"
 end
 
+if ['-v', '--version'].include? ARGV[0]
+  puts Suspenders::VERSION
+  exit 0
+end
+
 templates_root = File.expand_path(File.join("..", "templates"), File.dirname(__FILE__))
 Suspenders::AppGenerator.source_root templates_root
 Suspenders::AppGenerator.source_paths << Rails::Generators::AppGenerator.source_root << templates_root


tute:
:+1: In response to https://github.com/thoughtbot/suspenders/issues/637

mike-burns:
I usually use my package manager. In this case, gem(1):

```
~% gem list suspenders             

*** LOCAL GEMS ***

suspenders (1.20.0, 1.16.0)
```

tute:
Mike, you should upgrade your `suspenders`.

That is totally true, though, being it a command line, I did expect it to respond to this convention.

mike-burns:
To each their own, I suppose. Unix doesn't support `--version` (or any `--foo` arguments, in getopt(3)), so I'm not used to it.

```
~% ls --version
colorls: unknown option -- -
usage: colorls [-1AaCcdFfGgHhikLlmnopqRrSsTtux] [file ...]
~% sed --version
sed: unknown option -- -
usage: sed [-aEnru] [-i[extension]] command [file ...]
       sed [-aEnru] [-e command] [-f command_file] [-i[extension]] [file ...]
~% rcup --version
/usr/local/bin/rcup[166]: --: unknown option
~% ed --version
ed: unknown option -- -
usage: ed [-] [-s] [-p string] [file]
```

tute:
Ruby-land commands follow this pattern: `rails`, `ruby`, `foreman`, `rake`, `rspec` all respond to it.

`flog` has a creative answer:

```
% flog --version
/Users/tute/.rbenv/versions/2.2.3/lib/ruby/2.2.0/optparse.rb:1023: warning: instance variable @version not initialized
flog: version unknown
```

mike-burns:
I'm not opposed to this additional code going in, I just don't have a use for it.

tute:
> I'm not opposed to this additional code going in

Checking this file I found we don't need the deprecation message either, and proposed its removal: https://github.com/thoughtbot/suspenders/pull/642.

I don't feel strongly about the `--version` flag either way. I can do with `gem`, and when you call it `additional code` it makes me feel afraid of the dark future if we merge it in.

mcmire:
I think having a `--version` flag doesn't make sense for tools that are a part of your operating system or shell such as `ls`, `find`, etc. as they're all a part of the same ecosystem. Suspenders is an individual tool produced by an individual vendor, so it makes sense. That's my two cents.

PR #639, opened 2015-10-23
Introduce Heroku Pipelines support

zamith:
Why:

* We have observed that running acceptance through Heroku Pipelines is a
  better flow for code reviews, by allowing to show work in progress
  earlier, and fixing issue before merging into master

* Creating a pipeline for the new app, and adding the staging and
  production environments to it
* Adding a required `app.json` configuration file
* Setting `Rack::CanonicalHost` only in staging and production, and not
  review apps (otherwise review apps redirect to staging)
* Adding a setup script for review apps to scale up workers to 1 and run migrations

## Done

- [x] Check if heroku pipelines plugin is installed
```bash
if heroku pipelines &> /dev/null; then
  do_it
else
  printf 'You need heroku pipelines plugin. Run: heroku plugins:install heroku-pipelines'
fi
```
- [x] Fix `APPLICATION_HOST`: https://github.com/thoughtbot/suspenders/pull/639#issuecomment-152546822. If we do, we don't need a conditional around `Rack::CanonicalHost`.
- [x] Emails are being sent with the staging URL, and so links don't work for review apps (related to previous point)
- [x] Work around the fact that there are no after hooks after a review app is created, and we need to set up the database at first manually (`bin/setup_review_app`).
- [x] Emails aren't send because review apps are not starting with a worker dyno (`bin/setup_review_app`)
- [x] Add instructions for hooking up the pipeline to GitHub autodeploys in README
  * needs to set up Circle CI with the new GitHub repo
  * needs to hook up GitHub autodeploys in Heroku
  * needs to hook up GitHub auto review apps in Heroku

### DIFF
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -7,10 +7,9 @@ module Suspenders
 
       def set_heroku_remotes
         remotes = <<-SHELL.strip_heredoc
-
-          # Set up the staging and production apps.
           #{command_to_join_heroku_app('staging')}
           #{command_to_join_heroku_app('production')}
+
           git config heroku.remote staging
         SHELL
 
@@ -25,10 +24,62 @@ module Suspenders
         )
       end
 
+      def create_staging_heroku_app(flags)
+        rack_env = "RACK_ENV=staging RAILS_ENV=staging"
+        app_name = heroku_app_name_for("staging")
+
+        run_toolbelt_command "create #{app_name} #{flags}", "staging"
+        run_toolbelt_command "config:add #{rack_env}", "staging"
+      end
+
+      def create_production_heroku_app(flags)
+        app_name = heroku_app_name_for("production")
+
+        run_toolbelt_command "create #{app_name} #{flags}", "production"
+      end
+
       def set_heroku_rails_secrets
         %w(staging production).each do |environment|
           run_toolbelt_command(
             "config:add SECRET_KEY_BASE=#{generate_secret}",
+            environment,
+          )
+        end
+      end
+
+      def provide_review_apps_setup_script
+        app_builder.template(
+          "bin_setup_review_app.erb",
+          "bin/setup_review_app",
+          force: true,
+        )
+        app_builder.run "chmod a+x bin/setup_review_app"
+      end
+
+      def create_heroku_pipelines_config_file
+        app_builder.template "app.json.erb", "app.json"
+      end
+
+      def create_heroku_pipeline
+        pipelines_plugin = `heroku plugins | grep pipelines`
+        if pipelines_plugin.empty?
+          puts "You need heroku pipelines plugin. Run: heroku plugins:install heroku-pipelines"
+          exit 1
+        end
+
+        heroku_app_name = app_builder.app_name.dasherize
+        %w(staging production).each do |environment|
+          run_toolbelt_command(
+            "pipelines:create #{heroku_app_name} -a #{heroku_app_name}-#{environment} --stage #{environment}",
+            environment,
+          )
+        end
+      end
+
+      def set_heroku_serve_static_files
+        %w(staging production).each do |environment|
+          run_toolbelt_command(
+            "config:add RAILS_SERVE_STATIC_FILES=true",
             environment,
           )
         end
@@ -41,6 +92,7 @@ module Suspenders
       def command_to_join_heroku_app(environment)
         heroku_app_name = heroku_app_name_for(environment)
         <<-SHELL
+
 if heroku join --app #{heroku_app_name} &> /dev/null; then
   git remote add #{environment} git@heroku.com:#{heroku_app_name}.git || true
   printf 'You are a collaborator on the "#{heroku_app_name}" Heroku app\n'
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -6,9 +6,15 @@ module Suspenders
     extend Forwardable
 
     def_delegators :heroku_adapter,
+                   :create_heroku_pipelines_config_file,
+                   :create_heroku_pipeline,
+                   :create_production_heroku_app,
+                   :create_staging_heroku_app,
+                   :provide_review_apps_setup_script,
+                   :set_heroku_rails_secrets,
                    :set_heroku_remotes,
-                   :set_up_heroku_specific_gems,
-                   :set_heroku_rails_secrets
+                   :set_heroku_serve_static_files,
+                   :set_up_heroku_specific_gems
 
     def readme
       template 'README.md.erb', 'README.md'
@@ -69,7 +75,7 @@ module Suspenders
     end
 
     def provide_setup_script
-      template "bin_setup.erb", "bin/setup", force: true
+      template "bin_setup", "bin/setup", force: true
       run "chmod a+x bin/setup"
     end
 
@@ -130,6 +136,10 @@ module Suspenders
     def enable_rack_canonical_host
       config = <<-RUBY
 
+  if ENV.fetch("HEROKU_APP_NAME", "").include?("staging-pr-")
+    ENV["APPLICATION_HOST"] = ENV["HEROKU_APP_NAME"] + ".herokuapp.com"
+  end
+
   # Ensure requests are only served from one, canonical host name
   config.middleware.use Rack::CanonicalHost, ENV.fetch("APPLICATION_HOST")
       RUBY
@@ -137,7 +147,7 @@ module Suspenders
       inject_into_file(
         "config/environments/production.rb",
         config,
-        after: serve_static_files_line
+        after: "Rails.application.configure do",
       )
     end
 
@@ -369,29 +379,9 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
       run 'git init'
     end
 
-    def create_staging_heroku_app(flags)
-      rack_env = "RACK_ENV=staging RAILS_ENV=staging"
-      app_name = heroku_app_name_for("staging")
-
-      run_heroku "create #{app_name} #{flags}", "staging"
-      run_heroku "config:add #{rack_env}", "staging"
-    end
-
-    def create_production_heroku_app(flags)
-      app_name = heroku_app_name_for("production")
-
-      run_heroku "create #{app_name} #{flags}", "production"
-    end
-
     def create_heroku_apps(flags)
       create_staging_heroku_app(flags)
       create_production_heroku_app(flags)
-    end
-
-    def set_heroku_serve_static_files
-      %w(staging production).each do |environment|
-        run_heroku "config:add RAILS_SERVE_STATIC_FILES=true", environment
-      end
     end
 
     def provide_deploy_script
@@ -413,7 +403,6 @@ you can deploy to staging and production with:
     end
 
     def configure_automatic_deployment
-      staging_remote_name = heroku_app_name_for("staging")
       deploy_command = <<-YML.strip_heredoc
       deployment:
         staging:
@@ -518,20 +507,12 @@ end
       uncomment_lines("config/environments/#{environment}.rb", config)
     end
 
-    def run_heroku(command, environment)
-      run "heroku #{command} --remote #{environment}"
-    end
-
     def heroku_adapter
       @heroku_adapter ||= Adapters::Heroku.new(self)
     end
 
     def serve_static_files_line
       "config.serve_static_files = ENV['RAILS_SERVE_STATIC_FILES'].present?\n"
-    end
-
-    def heroku_app_name_for(environment)
-      "#{app_name.dasherize}-#{environment}"
     end
   end
 end
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -176,9 +176,12 @@ module Suspenders
       if options[:heroku]
         say "Creating Heroku apps"
         build :create_heroku_apps, options[:heroku_flags]
+        build :provide_review_apps_setup_script
         build :set_heroku_serve_static_files
         build :set_heroku_remotes
         build :set_heroku_rails_secrets
+        build :create_heroku_pipelines_config_file
+        build :create_heroku_pipeline
         build :provide_deploy_script
         build :configure_automatic_deployment
       end
--- a/spec/support/fake_heroku.rb
+++ b/spec/support/fake_heroku.rb
@@ -6,6 +6,9 @@ class FakeHeroku
   end
 
   def run!
+    if @args.first == "plugins"
+      puts "heroku-pipelines@0.29.0"
+    end
     File.open(RECORDER, 'a') do |file|
       file.puts @args.join(' ')
     end
@@ -37,6 +40,11 @@ class FakeHeroku
 
   def self.has_configured_vars?(remote_name, var)
     commands_ran =~ /^config:add #{var}=.+ --remote #{remote_name}\n/
+  end
+
+  def self.has_setup_pipeline_for?(app_name)
+    commands_ran =~ /^pipelines:create #{app_name} -a #{app_name}-staging --stage staging/ &&
+      commands_ran =~ /^pipelines:create #{app_name} -a #{app_name}-production --stage production/
   end
 
   def self.commands_ran
--- a/templates/bin_setup_review_app.erb
+++ b/templates/bin_setup_review_app.erb
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# Run this script to set up a review app's database and worker dyno
+
+set -e
+
+if [ -z "$1" ]; then
+  printf "You must provide a review app (same as the pull request) id.\n"
+  exit 64
+fi
+
+heroku pg:backups restore \
+  `heroku pg:backups public-url -a <%= app_name.dasherize %>-staging` \
+  DATABASE_URL \
+  --confirm <%= app_name.dasherize %>-staging-pr-$1 \
+  --app <%= app_name.dasherize %>-staging-pr-$1
+heroku run rake db:migrate --app <%= app_name.dasherize %>-staging-pr-$1
+heroku ps:scale worker=1 --app <%= app_name.dasherize %>-staging-pr-$1
+heroku restart --app <%= app_name.dasherize %>-staging-pr-$1
--- a/README.md
+++ b/README.md
@@ -17,6 +17,12 @@ Then run:
 
 This will create a Rails app in `projectname` using the latest version of Rails.
 
+### Associated services
+
+* Enable [Circle CI](https://circleci.com/) Continuous Integration
+* Enable [GitHub auto deploys to Heroku staging and review
+    apps](https://dashboard.heroku.com/apps/app-name-staging/deploy/github).
+
 ## Gemfile
 
 To see the latest and greatest gems, look at Suspenders'
@@ -128,9 +134,11 @@ This:
 * Adds the [Rails Stdout Logging][logging-gem] gem
   to configure the app to log to standard out,
   which is how [Heroku's logging][heroku-logging] works.
+* Creates a [Heroku Pipeline] for review apps
 
 [logging-gem]: https://github.com/heroku/rails_stdout_logging
 [heroku-logging]: https://devcenter.heroku.com/articles/logging#writing-to-your-log
+[Heroku Pipeline]: https://devcenter.heroku.com/articles/pipelines
 
 You can optionally specify alternate Heroku flags:
 
--- a/bin/setup
+++ b/bin/setup
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/bin/sh
 
 # Run this script immediately after cloning the codebase.
 
--- a/spec/features/heroku_spec.rb
+++ b/spec/features/heroku_spec.rb
@@ -20,6 +20,7 @@ RSpec.describe "Heroku" do
         "production",
         "SECRET_KEY_BASE",
       )
+      expect(FakeHeroku).to have_setup_pipeline_for(app_name)
 
       bin_setup_path = "#{project_path}/bin/setup"
       bin_setup = IO.read(bin_setup_path)
@@ -27,6 +28,14 @@ RSpec.describe "Heroku" do
       expect(bin_setup).to include("heroku join --app #{app_name}-production")
       expect(bin_setup).to include("heroku join --app #{app_name}-staging")
       expect(bin_setup).to include("git config heroku.remote staging")
+      expect(File.stat(bin_setup_path)).to be_executable
+
+      bin_setup_path = "#{project_path}/bin/setup_review_app"
+      bin_setup = IO.read(bin_setup_path)
+
+      expect(bin_setup).to include("heroku run rake db:migrate --app #{app_name}-staging-pr-$1")
+      expect(bin_setup).to include("heroku ps:scale worker=1 --app #{app_name}-staging-pr-$1")
+      expect(bin_setup).to include("heroku restart --app #{app_name}-staging-pr-$1")
       expect(File.stat(bin_setup_path)).to be_executable
 
       bin_deploy_path = "#{project_path}/bin/deploy"
@@ -50,6 +59,17 @@ RSpec.describe "Heroku" do
           commands:
             - bin/deploy staging
       YML
+    end
+
+    it "adds app.json file" do
+      expect(File).to exist("#{project_path}/app.json")
+    end
+
+    it "includes application name in app.json file" do
+      app_json_file = IO.read("#{project_path}/app.json")
+      app_name = SuspendersTestHelpers::APP_NAME.dasherize
+
+      expect(app_json_file).to match(/"name":"#{app_name}"/)
     end
   end
 
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -41,6 +41,16 @@ RSpec.describe "Suspend a new project with default configuration" do
     expect(secrets_file).to match(/secret_key_base: <%= ENV\["SECRET_KEY_BASE"\] %>/)
   end
 
+  it "adds bin/setup file" do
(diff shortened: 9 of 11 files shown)


geoffharcourt:
@tute, this looks OK, but I haven't done any worth with Pipelines yet, so it might be worth having another set of eyes.

tute:
Thanks @geoffharcourt! If you are working in Rails you'll probably be happy to check it out sometime in the near future! :)

cc @iwz @elle for some reviews. We still have some work to do but this is the bulk of it.

tute:
@zamith adding commits to https://github.com/thoughtbot/suspenders/tree/heroku-pipelines

Thanks for pairing!

lib/suspenders/app_builder.rb
```diff
@@ -366,6 +372,14 @@ def create_heroku_apps(flags)
       create_production_heroku_app(flags)
     end
 
+    def create_heroku_pipeline
+      pipeline_name = app_name.dasherize
+
+      run_heroku "pipelines:create -a #{app_name.dasherize}-staging", "staging"
```
rikkipitt: Should we be calling the pipeline by the base app name i.e. ```#{app_name.dasherize}``` rather than  ```#{app_name.dasherize}-staging```?

zamith:
@tute Anytime. :) Do you want to close this PR and open a new one from that branch? Or how do you want to go about it?

lib/suspenders/app_builder.rb
tute: > Should we be calling the pipeline by the base app name i.e. #{app_name.dasherize} rather than #{app_name.dasherize}-staging?

yes :+1:

tute:
Let's continue on your branch so we don't split the conversation.

lib/suspenders/app_builder.rb
zamith: When writing this we were assuming it would not work without the `-staging` suffix. Was that assumption wrong?

rikkipitt: Hi @zamith, I've used it without the ```-staging``` suffix with no problem. If you check the Pipelines docs on Heroku, they don't either. Maybe this slipped in by mistake as they do use ```-staging``` in one of their commands when adding an app to a pipeline.

zamith: Cool, then we can just drop it, I guess.

lib/suspenders/app_builder.rb
```diff
@@ -366,6 +372,20 @@ def create_heroku_apps(flags)
...
+    def create_heroku_pipeline
+      pipeline_name = app_name.dasherize
+
+      run_heroku "pipelines:create -a #{pipeline_name}", "staging"
+      run_heroku(
+        "pipelines:add -a #{app_name.dasherize}-staging #{pipeline_name}",
+        "staging",
+      )
+      run_heroku(
+        "pipelines:add -a #{app_name.dasherize}-production #{pipeline_name}",
```
elle: Both `app_name.dasherize` here and one line 380 could use `pipeline_name` for the interpolation.
And if on that thought, maybe consider renaming `pipeline_name` var to something that suits all 3 occurrences?

lib/suspenders/app_builder.rb
```diff
@@ -98,7 +102,9 @@ def enable_rack_canonical_host
       config = <<-RUBY
 
   # Ensure requests are only served from one, canonical host name
-  config.middleware.use Rack::CanonicalHost, ENV.fetch("APPLICATION_HOST")
+  unless ENV["HEROKU_APP_NAME"].include?("staging-pr-")
```
elle: A question: why are we checking for `staging-pr-` instead of just `staging-`?

lib/suspenders/app_builder.rb
zamith: Right. The though process was that these are the same right now, but they are not necessarily always the same, and this might change in the future.

However, it would be great to have a name that would fit all, we just couldn't come up with one.

lib/suspenders/app_builder.rb
calebthompson: It's the naming convention for PR apps that Heroku uses. This isn't the staging remote.

zamith: Because we just don't want this to run on review apps, which will have the `pr` part. We want this on regular staging.

lib/suspenders/generators/app_generator.rb
```diff
@@ -114,6 +114,7 @@ def setup_production_environment
       build :enable_rack_canonical_host
       build :enable_rack_deflater
       build :setup_asset_host
+      build :create_heroku_pipelines_config_file
```
elle: The location where this is called seems a bit weird to me. Is it only needed in `production` or do we need it in the base app? It might make more sense to place it under `suspenders_customization` rather than `setup_production_environment` section.

lib/suspenders/app_builder.rb
elle: Is the `app_name` passed in to this method?
If not, I would keep it simple for now, rather than something that will be nice in the future.
Also, if we are calling all the apps `x-staging` and `x-production` then it makes sense to call the pipeline `x`
What do you think?

lib/suspenders/app_builder.rb
elle: Ah, cool, thanks for explaining

tute: Technically `staging-` would reach the same effect. The staging app doesn't have a dash suffix or anything after.

lib/suspenders/app_builder.rb
elle: Also looking at the specs, even the spec use `app_name` when checking that those commands have run

zamith: That is true. At this point I don't see any problem with doing that. /cc @tute

lib/suspenders/app_builder.rb
zamith: That is true, though I feel it is more explicit to have the `pr`, wouldn't you say?

lib/suspenders/generators/app_generator.rb
zamith: This will be run on the app, but will only affect staging and production environments. `suspenders_customization` seems fine to me.

tute:
A note to take into account: emails are being sent with the staging URL, and so links don't work for review apps.

@zamith let's work on this on Friday if you can! We'll see when we can both sync up.

calebthompson:
This also breaks the `ENV["HOST"]` required variable. That variable is used in rack canonical host and redirects you away from the PR app.

lib/suspenders/app_builder.rb
tute: > I feel it is more explicit to have the pr, wouldn't you say?

Agreed.

lib/suspenders/generators/app_generator.rb
```diff
@@ -48,6 +48,7 @@ def suspenders_customization
       invoke :setup_dotfiles
       invoke :setup_git
       invoke :setup_database
+      invoke :setup_app_json
```
tute: Should this happen only when the heroku flag is on? Should it happen before `create_heroku_pipeline`?

tute:
> This also breaks the `ENV["HOST"]` required variable. That variable is used in rack canonical host and redirects you away from the PR app.

@calebthompson I don't follow. What breaks that var, and how?

calebthompson:
HOST needs to point to the app's URL. Cloning it from staging, etc. means we redirect to staging, etc. and can't ever route to the PR app.

I haven't been able to find a way to set the HOST env to the current app, unfortunately.

HOST is, I believe, used in rack canonical host and in email links.

tute:
> HOST is, I believe, used in rack canonical host and in email links.

Right, yes. Renamed to `APPLICATION_HOST` after 9d27905 and f6f4869.

calebthompson:
I think you've traded one problem variable for another.

I noticed you've specified that all of the variables are required, but not set defaults. Most of these should either be defaulted or generated, I think. https://devcenter.heroku.com/articles/app-json-schema#env

Since you're inheriting the variables from the downstream app, APPLICATION_HOST will be set to, for example, staging.upcase.com. This means that when you visit upcase-pr-42.herokuapp.com, the canonical host middleware will redirect you to staging.upcase.com.

When rack canonical host is disabled, you'll still be using APPLICATION_HOST (staging.upcase.com) as the URL to base mailer links on, so your i.e., "Confirm email address" link will take you to the wrong application and you won't be able to confirm the email.

There may be other places it's used as well.

lib/suspenders/generators/app_generator.rb
zamith: Yes, this only makes sense with the Heroku flag on.

tute:
Thanks Caleb, have that in mind.

This is funny, I'm hand testing pipelines and this branch, and the name can't contain zeros. Wat!

zamith:
@tute I did that in #646, adding also a few more things that would be useful before adding more functionality to the Heroku path here.

tute:
> /home/travis/.rvm/gems/ruby-2.2.3@global/gems/bundler-1.10.6/lib/bundler/resolver.rb:347:in `block in verify_gemfile_dependencies_are_found!': Could not find gem 'airbrake (>= 0) ruby' in any of the gem sources listed in your Gemfile or available on this machine. (Bundler::GemNotFound)

That's a surprising CI error.

zamith:
I fixed that already on the other branch. I'll have to merge it in here.

tute:
> I fixed that already on the other branch. I'll have to merge it in here.

Let's get that one into suspender's master, and rebase this on top. Is there anything left on the other one?

tute:
Here is a patch to an existing application that fixes the HOST for both assets and emails (and any):

```diff
From 41e9d7b1b846689b7170be3bc26c32492a891dbc Mon Sep 17 00:00:00 2001
From: Tute Costa <tutecosta@gmail.com>
Date: Fri, 30 Oct 2015 16:32:34 -0400
Subject: [PATCH] Configure correct application host for review apps

---
 config/environments/production.rb | 7 ++++---
 config/environments/staging.rb    | 4 ----
 2 files changed, 4 insertions(+), 7 deletions(-)

diff --git a/config/environments/production.rb b/config/environments/production.rb
index 74deb74..6a7281c 100644
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,6 +1,9 @@
 require Rails.root.join("config/smtp")
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
+  if ENV["HEROKU_APP_NAME"].include?("staging-pr-")
+    ENV["HOST"] = "#{ENV['HEROKU_APP_NAME']}.herokuapp.com"
+  end

   # Code is not reloaded between requests.
   config.cache_classes = true
@@ -29,9 +32,7 @@ Rails.application.configure do
   config.middleware.use Rack::Deflater

   # Ensure requests are only served from one, canonical host name
-  unless ENV["HEROKU_APP_NAME"].include?("staging-pr-")
-    config.middleware.use Rack::CanonicalHost, ENV.fetch("HOST")
-  end
+  config.middleware.use Rack::CanonicalHost, ENV.fetch("HOST")

   # Compress JavaScripts and CSS.
   config.assets.js_compressor = :uglifier
diff --git a/config/environments/staging.rb b/config/environments/staging.rb
index 6f853fa..90c6651 100644
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -5,9 +5,5 @@ Mail.register_interceptor(
 )

 Rails.application.configure do
-  if ENV["HEROKU_APP_NAME"]
-    config.action_controller.asset_host = "#{ENV['HEROKU_APP_NAME']}.herokuapp.com"
-  end
-
   config.force_ssl = false
 end
--
2.6.2
```

spec/support/fake_heroku.rb
```diff
@@ -36,8 +36,16 @@ def self.has_created_app_for?(environment, flags = nil)
...
-    File.foreach(RECORDER).any? do |line|
-      line =~ /^config:add #{var}=.+ --remote #{remote_name}\n$/
-    end
+    commands_ran =~ /^config:add #{var}=.+ --remote #{remote_name}\n/
+  end
+
+  def self.has_setup_pipeline_for?(app_name)
+    commands_ran =~ /^pipelines:create -a #{app_name}/ &&
+      commands_ran =~ /^pipelines:add -a #{app_name}-staging #{app_name}/ &&
+      commands_ran =~ /^pipelines:add -a #{app_name}-production #{app_name}/
```
tute: My commits changed these commands, I didn't update them yet and might be a second reason why we broke CI.

tute: Will wait till the other branch gets merged before fixing this (not sure if this Fake changes again).

zamith: I have that fixed in my local branch as well. Yes, I think it's best to wait for the other and merge everything. The Fake should not impact this at all.

tute:
@zamith just rebased your branch on top of master, with the test improvements PR in. That was a good split! :+1: 

I squashed before the commits, as feedback was included a while ago and I didn't think we continue to need all that context, and so I'd fix the rebase merge conflicts only once.

zamith:
@tute Sure, let's pick this up where we left it. I'll be adding the check to see if pipelines is already installed, which now should be easier to test. :)

tute:
Added two commits for your review, please. This is getting ready! Will add some README documentation now.

spec/support/fake_heroku.rb
```diff
@@ -39,6 +39,11 @@ def self.has_configured_vars?(remote_name, var)
     commands_ran =~ /^config:add #{var}=.+ --remote #{remote_name}\n/
   end
 
+  def self.has_setup_pipeline_for?(app_name)
+    commands_ran =~ /^pipelines:create #{app_name} -a #{app_name}-staging --stage staging/ &&
```
tute: Would `--remote staging` work here?

elle: I think so

tute:
This is ready for a last review. Will appreciate @zamith and @calebthompson reviews. Thank you! :+1: :smiley:

lib/suspenders/adapters/heroku.rb
```diff
@@ -34,6 +33,46 @@ def set_heroku_rails_secrets
...
+        template "app.json.erb", "app.json"
+      end
+
+      def create_heroku_pipeline
+        heroku_app_name = app_name.dasherize
+
+        pipelines_plugin = `heroku plugins | grep pipelines`
+        if pipelines_plugin.empty?
+          puts 'You need heroku pipelines plugin. Run: heroku plugins:install heroku-pipelines'
+          exit 1
```
teoljungberg: This feels a bit too harsh. I think that If the pipelines plugin is not installed, we should just setup the applications as normal. And not completely stop.

Thoughts?

zamith: I agree. But if so, do we still generate the config file and the setup scripts for the pipelines, or do we skip all that?

lib/suspenders/adapters/heroku.rb
```diff
@@ -34,6 +33,43 @@ def set_heroku_rails_secrets
...
+        end
+        app_builder.run_heroku(
+          "pipelines:create #{heroku_app_name} -a #{heroku_app_name}-staging --stage staging",
```
teoljungberg: What do you think about re-writing this to loop over the available remotes and make this command dynamic?


```ruby
%w(staging production).each do |remote|
  app_builder.run_heroku(
    "pipelines:create #{heroku_app_name} -a #{heroku_app_name}-#{remote} --stage #{remote}",
  )
end
```

Thoughts?

lib/suspenders/adapters/heroku.rb
```diff
@@ -34,6 +33,46 @@ def set_heroku_rails_secrets
...
+        app_builder.run_heroku(
+          "pipelines:add #{heroku_app_name} -a #{heroku_app_name}-production --stage production",
+          "production",
+        )
+      end
+
+      def set_heroku_serve_static_files
+        %w(staging production).each do |environment|
+          app_builder.run_heroku(
+            "config:add RAILS_SERVE_STATIC_FILES=true",
```
teoljungberg: Could this be resolved by setting this in `config/application.rb` instead? Less environment variables to think of.

lib/suspenders/adapters/heroku.rb
teoljungberg: That sounds like a sane approach.

templates/bin_setup_review_app.erb
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env sh
+
+# Run this script to set up a review app's database and worker dyno
+
+set -e
+
+if [ -z "$1" ]; then
+  printf "You must provide a review app (same as the pull request) id.\n"
+  exit 64
```
teoljungberg: Nice use of proper error codes.

lib/suspenders/app_builder.rb
```diff
@@ -130,6 +134,10 @@ def configure_smtp
     def enable_rack_canonical_host
       config = <<-RUBY
 
+  if ENV["HEROKU_APP_NAME"].include?("staging-pr-")
```
dgalarza: In my testing of this on a side project I ran into an issue where the ENV var `HEROKU_APP_NAME` was not set on the main staging server and only set on the review apps. Not sure if that was just something I did wrong but I ended up changing this to:

```ruby
if ENV.fetch("HEROKU_APP_NAME", "").include?("staging-pr-")
# ...
end
```

templates/bin_setup_review_app.erb
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env sh
```
mike-burns: Unless we intend to support Solaris 8,  this should be

```sh
#!/bin/sh
```

templates/bin_setup_review_app.erb
```diff
@@ -0,0 +1,18 @@
...
+
+# Run this script to set up a review app's database and worker dyno
+
+set -e
+
+if [ -z "$1" ]; then
+  printf "You must provide a review app (same as the pull request) id.\n"
+  exit 64
+else
```
mike-burns: We can reduce indentation by taking advantage of the fact that `exit` terminates the program.

```sh
if [ -z "$1" ]; then
  echo "You must provide a review app (same as the pull request) id."
  exit 64
heroku # ...

File: lib/suspenders/adapters/heroku.rb (hunk 8 above)
Comment (tute) [inline lib/suspenders/adapters/heroku.rb]: > This feels a bit too harsh. I think that If the pipelines plugin is not installed, we should just setup the applications as normal. And not completely stop.

This is inline with our goal of making `suspenders` command fail when any sub-action fails (https://github.com/thoughtbot/suspenders/issues/493). I'd rather know that something went wrong by not having an application generated, rather than having to scroll up the logs and find out what might have failed.

File: lib/suspenders/adapters/heroku.rb (hunk 9 above)
Comment (tute) [inline lib/suspenders/adapters/heroku.rb]: > What do you think about re-writing this to loop over the available remotes and make this command dynamic?

:+1:

File: lib/suspenders/adapters/heroku.rb (hunk 10 above)
Comment (tute) [inline lib/suspenders/adapters/heroku.rb]: > Could this be resolved by setting this in `config/application.rb` instead? Less environment variables to think of.

I think that's a good idea. This is moved code though, I'd do it as part of another PR. For reference, it was introduced in
https://github.com/thoughtbot/suspenders/commit/3f937cdb3c4e945b85809e2e6be48f8af59a910f. Thanks!

File: lib/suspenders/app_builder.rb (hunk 12 above)
Comment (tute) [inline lib/suspenders/app_builder.rb]: > In my testing of this on a side project I ran into an issue where the ENV var HEROKU_APP_NAME was not set on the main staging server and only set on the review apps. Not sure if that was just something I did wrong but I ended up changing this to:

Thank you!

File: templates/bin_setup_review_app.erb (hunk 13 above)
Comment (tute) [inline templates/bin_setup_review_app.erb]: > Unless we intend to support Solaris 8,  this should be
>
> #!/bin/sh

All other scripts are using `env`:

```
./bin/rake:#!/usr/bin/env ruby
./bin/rspec:#!/usr/bin/env ruby
./bin/setup:#!/usr/bin/env sh
./bin/suspenders:#!/usr/bin/env ruby
./spec/fakes/bin/heroku:#!/usr/bin/env ruby
./spec/fakes/bin/hub:#!/usr/bin/env ruby
./templates/bin_setup:#!/usr/bin/env sh
./templates/bin_setup_review_app.erb:#!/usr/bin/env sh
```

Should we change them as part of a different PR?

File: templates/bin_setup_review_app.erb (hunk 14 above)
Comment (tute) [inline templates/bin_setup_review_app.erb]: > We can reduce indentation by taking advantage of the fact that exit terminates the program.

Right, thanks!

**Comment by tute** [comment]
@zamith I think this is ready for merge. I can't find why we get this, though:

```
Failures:

  1) Heroku --heroku suspends a project for Heroku
     Failure/Error: expect(FakeHeroku).to have_configured_vars("staging", "SECRET_KEY_BASE")
       expected #has_configured_vars?("staging", "SECRET_KEY_BASE") to return true, got false
     # ./spec/features/heroku_spec.rb:18:in `block (3 levels) in <top (required)>'

  2) Heroku --heroku includes application name in app.json file
     Failure/Error: app_json_file = IO.read("#{project_path}/app.json")
     Errno::ENOENT:
       No such file or directory @ rb_sysopen - /Users/tute/Sites/opensource/suspenders/suspenders/tmp/dummy_app/app.json
     # ./spec/features/heroku_spec.rb:69:in `read'
     # ./spec/features/heroku_spec.rb:69:in `block (3 levels) in <top (required)>'

  3) Heroku --heroku adds app.json file
     Failure/Error: expect(File).to exist("#{project_path}/app.json")
       expected File to exist
     # ./spec/features/heroku_spec.rb:65:in `block (3 levels) in <top (required)>'

Finished in 50.15 seconds (files took 0.12484 seconds to load)
4 examples, 3 failures

Failed examples:

rspec ./spec/features/heroku_spec.rb:10 # Heroku --heroku suspends a project for Heroku
rspec ./spec/features/heroku_spec.rb:68 # Heroku --heroku includes application name in app.json file
rspec ./spec/features/heroku_spec.rb:64 # Heroku --heroku adds app.json file
```

Do you know?

**Comment by mike-burns** [comment]
Higher up in the tests:

```
/home/travis/.rvm/gems/ruby-2.2.4/gems/railties-4.2.5/lib/rails/generators/rails/app/app_generator.rb:24:in `method_missing': undefined method `run_heroku' for #<Suspenders::AppGenerator:0x000000011bbb50> (NoMethodError)
```

File: templates/bin_setup_review_app.erb (hunk 13 above)
Comment (mike-burns) [inline templates/bin_setup_review_app.erb]: I don't understand -- what would we change `#!/usr/bin/env ruby` to?

Comment (tute) [inline templates/bin_setup_review_app.erb]: > I don't understand -- what would we change `#!/usr/bin/env ruby` to?

I misread your suggestion, fixing `#!/usr/bin/env sh`. Thanks!

**Comment by tute** [comment]
> Higher up in the tests:
>
> /home/travis/.rvm/gems/ruby-2.2.4/gems/railties-4.2.5/lib/rails/generators/rails/app/app_generator.rb:24:in `method_missing': undefined method `run_heroku' for #<Suspenders::AppGenerator:0x000000011bbb50> (NoMethodError)

Right. It's like `@heroku_adapter ||= Adapters::Heroku.new(self)` in `AppBuilder` initializes it with `AppGenerator` as a dependency. Confusing to me, going offline for now but it's the only thing that's left!

**Comment by zamith** [comment]
I'll try to give it a shot tomorrow.

File: lib/suspenders/adapters/heroku.rb (hunk 8 above)
Comment (teoljungberg) [inline lib/suspenders/adapters/heroku.rb]: @tute We should fail if a subcommand fails, but I don't consider the user missing a plugin that doesn't impact the use of suspenders a valid reason for failing the whole generation.

Comment (mike-burns) [inline lib/suspenders/adapters/heroku.rb]: The failure gets buried and is easy to miss. We should fail immediately.

There is a movement toward splitting Suspenders into tiny components; this will allow us to separate failures more easily. Until then, we need to fail fast.

**Comment by tute** [comment]
@zamith some of the errors were fixed by moving all heroku methods into the adapter.

An error persists when using `FakeHeroku`. See the check:

```ruby
        pipelines_plugin = `heroku plugins | grep pipelines`
        if pipelines_plugin.empty?
          puts 'You need heroku pipelines plugin. Run: heroku plugins:install heroku-pipelines'
          exit 1
        end
```

I can't make `FakeHeroku` `plugins` return something that will make that check go through, and not `exit` during tests. This makes the "created pipelines" expectation fail right now (CI is green because I commented out the `exit`, but that is not what we want).

When we solve this, I'll merge.

tute:
> When we solve this, I'll merge.

Solved! Can you please review the last two commits? Then, merge and release. Thanks!!

zamith:
Sorry, just saw this now. Do you have any pending issues with this?

tute:
This was released, if there are issues we'll handle in subsequent PRs. Thanks for your work! :clap: 

Happy New Year!

zamith:
Cool. Thanks for the pairing. :)

Happy New Year!

tute:
It looks like we now have the `scripts.postdeploy` option available in `app.json`:

```json
  "scripts": {
    "postdeploy": "bundle exec rake bootstrap"
  },
```

https://devcenter.heroku.com/articles/app-json-schema#scripts

rikkipitt:
@tute, they finally released this? Brilliant. Would be worth testing it works when promoting an app from staging to production with an assets compile rake command. Want me to give it a go?

tute:
I was thinking of getting rid of our custom https://github.com/thoughtbot/suspenders/blob/4f67c6f17245ccd077f7bedc641b3724f3c9dd8e/templates/bin_setup_review_app.erb. We can run `rake db:reset` for setting up a database with data, but I'm not sure how could we scale up workers so review apps can send emails.

I'm not sure what you mean about promoting apps and this script, @rikkipitt.

devindreszer:
@tute You can setup your dyno configuration in the `app.json` file as well:

```
"formation": {
    "web": {
      "quantity": 1
    },
    "worker": {
      "quantity": 1
    }
  }
```

PR #641, opened 2015-10-23
Remove Capybara-Webkit deprecation warning

geoffharcourt:
Capybara-Webkit's current release (1.7.1) issues a deprecation warning
on the use of `page.driver.block_unknown_urls`. This change updates to
use the preferred, non-deprecated syntax.

### DIFF
--- a/templates/rails_helper.rb
+++ b/templates/rails_helper.rb
@@ -14,8 +14,9 @@ end
 
 RSpec.configure do |config|
   config.before(:each, js: true) do
-    page.driver.block_unknown_urls
+    page.block_unknown_urls = true
   end
+
   config.include Features, type: :feature
   config.infer_base_class_for_anonymous_controllers = false
   config.infer_spec_type_from_file_location!


tute:
This looks like a simpler fix than https://github.com/thoughtbot/suspenders/pull/636. In that PR we were discussing if we should configure each gem in each file. I didn't hear anyone feeling strongly either way yet.

geoffharcourt:
@tute, I missed that PR. I am very pro-moving the Capybara config to its own file in `spec/support/capbara-webkit.rb`, but I am very anti-using the block format, because as soon as you do any further Capybara-Webkit config (which I think almost any involved app will do), the diff from moving the `&:` back to the block syntax creates a needlessly noisy diff.

tute:
> I am very pro-moving the Capybara config to its own file in spec/support/capbara-webkit.rb, but I am very anti-using the block format, because as soon as you do any further Capybara-Webkit config (which I think almost any involved app will do), the diff from moving the &: back to the block syntax creates a needlessly noisy diff.

Sounds good to me! Want to close this PR in favor of that one? I think we can merge it as is then. :+1:

geoffharcourt:
Closing in favor of #636

templates/rails_helper.rb
```diff
@@ -14,8 +14,9 @@ module Features
 
 RSpec.configure do |config|
   config.before(:each, js: true) do
-    page.driver.block_unknown_urls
+    page.block_unknown_urls = true
```
jferris: I'm surprised this works. The `page` object here is a `Capybara::Session`, and I don't know how this property would be defined on it.

geoffharcourt: You're right. Fixing the reference, but I am fairly sure I got this to pass when I ran `rake`, which is concerning.

PR #642, opened 2015-10-23
Remove old deprecation

tute:
This deprecation note was added almost exactly four years ago:
2011-10-21. This was before we released version 1, so it's not needed
anymore.


mike-burns:
Nice find!

PR #645, opened 2015-10-30
Add bullet as development dependency

tute:
Bullet notifies automatically if there is any request in the application
where we are performing N+1 queries, or unused eager loading.

With this check we make sure we eager load what is needed, and nothing
more, without having to check each endpoint.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -19,6 +19,23 @@ module Suspenders
       )
     end
 
+    def add_bullet_gem_configuration
+      config = <<-RUBY
+  config.after_initialize do
+    Bullet.enable = true
+    Bullet.bullet_logger = true
+    Bullet.rails_logger = true
+  end
+
+      RUBY
+
+      inject_into_file(
+        "config/environments/development.rb",
+        config,
+        after: "config.action_mailer.raise_delivery_errors = true\n",
+      )
+    end
+
     def raise_on_unpermitted_parameters
       config = <<-RUBY
     config.action_controller.action_on_unpermitted_parameters = :raise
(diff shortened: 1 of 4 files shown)


zamith:
What kind of notification are you expecting from bullet? I think it writes to a log by default, right?

thiagogabriel:
@zamith "Bullet won't do ANYTHING unless you tell it to explicitly."

I guess [this](https://github.com/Helabs/pah/blob/master/lib/pah/files/config/initializers/bullet.rb) configuration is enough, what do you think @tute ?

thiagogabriel:
Since it will also be used on specs, it is important to include `Bullet.raise = true` to make the tests fail.

tute:
I'm going to go with this for now:
```ruby
  config.after_initialize do
    Bullet.enable = true
    Bullet.bullet_logger = true
    Bullet.rails_logger = true
  end
```

I don't think it should fail the build, I remember having seen false positives. After testing for a while we can discuss that again.

zamith:
That config will log everything, but not warn you about anything right? Do you forget to check the logs, or is that not an issue?

tute:
> That config will log everything, but not warn you about anything right? Do you forget to check the logs, or is that not an issue?

Without bullet I have to see N+1 patterns in `development.log` for specific actions, or much more verbose `test.log`. We might not see them as we might be working with less data, in which case we have to consciously check for potential N+1 due to lack of eager loading (or unused eager loads).

With bullet, you get a `bullet.log` formatted file, much easier to read.

zamith:
I agree with that, and I think bullet is a nice add. Just questioning how you would use it. Especially if it comes with suspenders and you're not aware it is there, because it doesn't announce itself in any way. On the other hand, I too have experienced false positives, so failing the build doesn't seem like a good option as a default. 

It's a nice to have though.

tute:
Thanks!

PR #646, opened 2015-10-30
Improve testability

zamith:
Improve testability for Heroku actions

Why:

* It was very coupled and difficult to test in isolation, there was pretty much one single integration test.

* Making it easier to test and reason about Heroku functionality we want to use, by allows for smaller, faster and more focused tests.
* Using mocks instead of depending on side effects because it was tricky to build a fake of the an AppBuilder, that has a lot of functionality and it is already being tested in our integration tests.
* Using a mockist approach, that ensures the correct methods are called with the right arguments (this last part is wip)

Interested in opinions before I go much further with this. /cc @tute

### DIFF
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -1,0 +1,67 @@
+module Suspenders
+  module Adapters
+    class Heroku
+      def initialize(app_builder)
+        @app_builder = app_builder
+      end
+
+      def set_heroku_remotes
+        remotes = <<-SHELL.strip_heredoc
+
+          # Set up the staging and production apps.
+          #{command_to_join_heroku_app('staging')}
+          #{command_to_join_heroku_app('production')}
+        SHELL
+
+        app_builder.append_file "bin/setup", remotes
+      end
+
+      def set_up_heroku_specific_gems
+        app_builder.inject_into_file(
+          "Gemfile",
+          %{\n\s\sgem "rails_stdout_logging"},
+          after: /group :staging, :production do/,
+        )
+      end
+
+      def set_heroku_rails_secrets
+        %w(staging production).each do |environment|
+          run_toolbelt_command(
+            "config:add SECRET_KEY_BASE=#{generate_secret}",
+            environment,
+          )
+        end
+      end
+
+      private
+
+      attr_reader :app_builder
+
+      def command_to_join_heroku_app(environment)
+        heroku_app_name = heroku_app_name_for(environment)
+        <<-SHELL
+if heroku join --app #{heroku_app_name} &> /dev/null; then
+  git remote add #{environment} git@heroku.com:#{heroku_app_name}.git || true
+  printf 'You are a collaborator on the "#{heroku_app_name}" Heroku app\n'
+else
+  printf 'Ask for access to the "#{heroku_app_name}" Heroku app\n'
+fi
+        SHELL
+      end
+
+      def heroku_app_name_for(environment)
+        "#{app_builder.app_name.dasherize}-#{environment}"
+      end
+
+      def generate_secret
+        SecureRandom.hex(64)
+      end
+
+      def run_toolbelt_command(command, environment)
+        app_builder.run(
+          "heroku #{command} --remote #{environment}",
+        )
+      end
+    end
+  end
+end
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -1,6 +1,14 @@
+require "forwardable"
+
 module Suspenders
   class AppBuilder < Rails::AppBuilder
     include Suspenders::Actions
+    extend Forwardable
+
+    def_delegators :heroku_adapter,
+                   :set_heroku_remotes,
+                   :set_up_heroku_specific_gems,
+                   :set_heroku_rails_secrets
 
     def readme
       template 'README.md.erb', 'README.md'
@@ -214,14 +222,6 @@ end
       create_file '.ruby-version', "#{Suspenders::RUBY_VERSION}\n"
     end
 
-    def setup_heroku_specific_gems
-      inject_into_file(
-        "Gemfile",
-        %{\n\s\sgem "rails_stdout_logging"},
-        after: /group :staging, :production do/
-      )
-    end
-
     def enable_database_cleaner
       copy_file 'database_cleaner_rspec.rb', 'spec/support/database_cleaner.rb'
     end
@@ -386,35 +386,6 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
       create_production_heroku_app(flags)
     end
 
-    def set_heroku_remotes
-      remotes = <<-SHELL
-
-# Set up the staging and production apps.
-#{join_heroku_app('staging')}
-#{join_heroku_app('production')}
-      SHELL
-
-      append_file 'bin/setup', remotes
-    end
-
-    def join_heroku_app(environment)
-      heroku_app_name = heroku_app_name_for(environment)
-      <<-SHELL
-if heroku join --app #{heroku_app_name} &> /dev/null; then
-  git remote add #{environment} git@heroku.com:#{heroku_app_name}.git || true
-  printf 'You are a collaborator on the "#{heroku_app_name}" Heroku app\n'
-else
-  printf 'Ask for access to the "#{heroku_app_name}" Heroku app\n'
-fi
-      SHELL
-    end
-
-    def set_heroku_rails_secrets
-      %w(staging production).each do |environment|
-        run_heroku "config:add SECRET_KEY_BASE=#{generate_secret}", environment
-      end
-    end
-
     def set_heroku_serve_static_files
       %w(staging production).each do |environment|
         run_heroku "config:add RAILS_SERVE_STATIC_FILES=true", environment
@@ -453,8 +424,7 @@ you can deploy to staging and production with:
     end
 
     def create_github_repo(repo_name)
-      path_addition = override_path_for_tests
-      run "#{path_addition} hub create #{repo_name}"
+      run "hub create #{repo_name}"
     end
 
     def setup_segment
@@ -546,20 +516,12 @@ end
       uncomment_lines("config/environments/#{environment}.rb", config)
     end
 
-    def override_path_for_tests
-      if ENV['TESTING']
-        support_bin = File.expand_path(File.join('..', '..', 'spec', 'fakes', 'bin'))
-        "PATH=#{support_bin}:$PATH"
-      end
-    end
-
     def run_heroku(command, environment)
-      path_addition = override_path_for_tests
-      run "#{path_addition} heroku #{command} --remote #{environment}"
+      run "heroku #{command} --remote #{environment}"
     end
 
-    def generate_secret
-      SecureRandom.hex(64)
+    def heroku_adapter
+      @heroku_adapter ||= Adapters::Heroku.new(self)
     end
 
     def serve_static_files_line
--- a/spec/adapters/heroku_spec.rb
+++ b/spec/adapters/heroku_spec.rb
@@ -1,0 +1,52 @@
+require "spec_helper"
+
+module Suspenders
+  module Adapters
+    RSpec.describe Heroku do
+      it "sets the heroku remotes" do
+        setup_file = "bin/setup"
+        app_builder = double(app_name: app_name)
+        allow(app_builder).to receive(:append_file)
+
+        Heroku.new(app_builder).set_heroku_remotes
+
+        expect(app_builder).to have_received(:append_file).
+          with(setup_file, /heroku join --app #{app_name.dasherize}-production/)
+        expect(app_builder).to have_received(:append_file).
+          with(setup_file, /heroku join --app #{app_name.dasherize}-staging/)
+      end
+
+      it "sets up the heroku specific gems" do
+        app_builder = double(app_name: app_name)
+        allow(app_builder).to receive(:inject_into_file)
+
+        Heroku.new(app_builder).set_up_heroku_specific_gems
+
+        expect(app_builder).to have_received(:inject_into_file).
+          with("Gemfile", /rails_stdout_logging/, anything)
+      end
+
+      it "sets the heroku rails secrets" do
+        app_builder = double(app_name: app_name)
+        allow(app_builder).to receive(:run)
+
+        Heroku.new(app_builder).set_heroku_rails_secrets
+
+        expect(app_builder).to(
+          have_configured_var("staging", "SECRET_KEY_BASE"),
+        )
+        expect(app_builder).to(
+          have_configured_var("production", "SECRET_KEY_BASE"),
+        )
+      end
+
+      def app_name
+        SuspendersTestHelpers::APP_NAME
+      end
+
+      def have_configured_var(remote_name, var)
+        have_received(:run).with(/config:add #{var}=.+ --remote #{remote_name}/)
+      end
+    end
+  end
+end
--- a/spec/features/heroku_spec.rb
+++ b/spec/features/heroku_spec.rb
@@ -1,57 +1,72 @@
 require "spec_helper"
 
 RSpec.describe "Heroku" do
-  before do
+  context "--heroku" do
+    before(:all) do
+      clean_up
+      run_suspenders("--heroku=true")
+    end
+
+    it "suspends a project for Heroku" do
+      app_name = SuspendersTestHelpers::APP_NAME.dasherize
+
+      expect(FakeHeroku).to(
+        have_gem_included(project_path, "rails_stdout_logging"),
+      )
+      expect(FakeHeroku).to have_created_app_for("staging")
+      expect(FakeHeroku).to have_created_app_for("production")
+      expect(FakeHeroku).to have_configured_vars("staging", "SECRET_KEY_BASE")
+      expect(FakeHeroku).to have_configured_vars(
+        "production",
+        "SECRET_KEY_BASE",
+      )
+
+      bin_setup_path = "#{project_path}/bin/setup"
+      bin_setup = IO.read(bin_setup_path)
+
+      expect(bin_setup).to include("heroku join --app #{app_name}-production")
+      expect(bin_setup).to include("heroku join --app #{app_name}-staging")
+      expect(File.stat(bin_setup_path)).to be_executable
+
+      bin_deploy_path = "#{project_path}/bin/deploy"
+      bin_deploy = IO.read(bin_deploy_path)
+
+      expect(bin_deploy).to include("heroku run rake db:migrate")
+      expect(File.stat(bin_deploy_path)).to be_executable
+
+      readme = IO.read("#{project_path}/README.md")
+
+      expect(readme).to include("./bin/deploy staging")
+      expect(readme).to include("./bin/deploy production")
+
+      circle_yml_path = "#{project_path}/circle.yml"
+      circle_yml = IO.read(circle_yml_path)
+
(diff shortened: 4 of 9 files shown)


lib/suspenders/adapters/heroku.rb
```diff
@@ -0,0 +1,52 @@
+module Suspenders
+  module Adapters
+    class Heroku
+      def initialize(app_builder)
+        @app_builder = app_builder
```
tute: Should we define an `attr_reader` for `app_builder`, and access it through the getter?

lib/suspenders/app_builder.rb
```diff
@@ -1,6 +1,11 @@
+require "forwardable"
+
 module Suspenders
   class AppBuilder < Rails::AppBuilder
     include Suspenders::Actions
+    extend Forwardable
```
tute: Do we need `Forwardable`? Does `SimpleDelegator` achieve the same result? How do they differ?

lib/suspenders/adapters/heroku.rb
```diff
@@ -0,0 +1,52 @@
...
+
+# Set up the staging and production apps.
+#{join_heroku_app('staging')}
+#{join_heroku_app('production')}
+      SHELL
+
+        @app_builder.append_file "bin/setup", remotes
+      end
+
+      def join_heroku_app(environment)
```
tute: Does this need to be a public method?

tute:
Looks good to me! Left some comments and questions, will wait for your answers and we merge. Thank you! :smiley: :clap:

lib/suspenders/app_builder.rb
zamith: Simple Delegator only allows delegation to one object, which you define in
the initialization. Right now we only have one adapter, but we can have
more (at least GitHub), so Forwardable gives us that flexibility.

lib/suspenders/adapters/heroku.rb
zamith: You mean a private reader? I usually do that, but as it is not a well establish convention (there is actually a warning against it in ruby), I didn't do that. Shall I move to that?

lib/suspenders/adapters/heroku.rb
zamith: Not at all, but it was a public method on app builder so I didn't change the visibility. However, we can take this opportunity to do that.

zamith:
@tute I made the expectations a bit more specific, what do you think?

lib/suspenders/adapters/heroku.rb
mcmire: Ruby will print a warning if it's private, but not protected, so you'd need to make it that.

zamith: Is that your preferred way of dealing with this situation? Is that the solution you want, or the solution you can have, because of this warning?

lib/suspenders/adapters/heroku.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      private
+
+      def join_heroku_app(environment)
+        heroku_app_name = heroku_app_name_for(environment)
+        <<-SHELL.strip_heredoc
+          if heroku join --app #{heroku_app_name} &> /dev/null; then
+            git remote add #{environment} git@heroku.com:#{heroku_app_name}.git || true
+            printf 'You are a collaborator on the "#{heroku_app_name}" Heroku app\n'
```
zamith: Should we care about the line length in a heredoc? Any changes here will affect the generated file.

lib/suspenders/adapters/heroku.rb
mcmire: Yes, that's how we tend to solve that problem (and the warning will actually be removed in some future version of Ruby). It's a well established convention within thoughtbot to do this, at least, and so I think that's what we'd prefer to do here.

tute: I did not know about the warning. I believe the reasoning behind them being protected is for subclassing, but I never found myself in the need to make a reader protected yet (and always defaulted to private).

The reason behind using getters instead of instance variables is that they make refactoring, renaming and extending objects easier.

lib/suspenders/adapters/heroku.rb
tute: > Should we care about the line length in a heredoc? Any changes here will affect the generated file.

I'm personally ok with this longer line.

tute: > Should we care about the line length in a heredoc? Any changes here will affect the generated file.

We can also leave the previous, unindented version (if we get a hound complaint we ignore it as we are currently ignoring the long line warning).

spec/spec_helper.rb
```diff
@@ -6,6 +6,8 @@
 
 Dir['./spec/support/**/*.rb'].each { |file| require file }
 
+ENV["TESTING"] = "1"
```
tute: It looks like we use this env flag twice in suspenders' code, to do the same `PATH` trick. Do you think we should instead forget about that variable, and modify the path directly in here? We have only two bin scripts there, and we don't want to hit the real commands while testing, so I'd probably do that.

tute:
Left a question around `ENV["TESTING"]`, all else looks good to me! :clap:

lib/suspenders/adapters/heroku.rb
zamith: I think it's much uglier looking, but I have no really good reason not to do that. Will do.

lib/suspenders/adapters/heroku.rb
zamith: I usually go with private accessors myself. So, :+1:.

zamith:
@tute What do think of this solution?

tute:
Looking great! A super minor nitpick, and a more real question. I think then we merge.

tute:
:clap: :smiley: :heart:

lib/suspenders/app_builder.rb
```diff
@@ -531,16 +512,12 @@ def raise_on_missing_translations_in(environment)
       uncomment_lines("config/environments/#{environment}.rb", config)
     end
 
-    def override_path_for_tests
-      if ENV['TESTING']
-        support_bin = File.expand_path(File.join('..', '..', 'spec', 'fakes', 'bin'))
-        "PATH=#{support_bin}:$PATH"
-      end
+    def run_heroku(command, environment)
+      run "heroku #{command} --remote #{environment}"
```
tute: Should we delegate this method too, and have it defined only once? Does `set_heroku_rails_secrets` also belong in the adapter?

tute: There are many heroku specific snippets in the `AppBuilder`. What is the boundary, you think? Should we continue extracting functionality into the new object?

zamith: Yes, my idea is to eventually get it all removed. This is just a start.

tute:
> Yes, my idea is to eventually get it all removed. This is just a start. What are your thoughts?

Sounds good to me. Will loop in another pair of eyes before merging. @mcmire @jferris can you please let me know what you think?

mcmire:
Ping me tomorrow on this again, I'll take a closer look then.

tute:
> Ping me tomorrow on this again, I'll take a closer look then.

Pinging @mcmire again!

zamith:
I extracted some more examples, which I think makes the discussion even easier.

jferris:
I don't know how hard it is, but I'd be interested in seeing the adapters restructured as other generators. We call some generators (like `rspec:install`), and I think it would be awesome if Suspenders could be broken down into largely calling a number of generators that you could opt in or out of.

That would also have the interesting effect of making it easier to pull in some Suspenders configurations after the fact, so you could opt out of mailer configs and then apply them later. This is something we've had a lot of requests for.

If that's particularly difficult, I could see the delegator-based adapter approach as a solid first step. Either way, I think it's great to see the app builder class getting broken up a bit; that thing has become massive and hard to navigate.

zamith:
@jferris The idea me and @tute were tossing around was that of being able to call a generator for heroku whenever you want, not only when creating the app, because it turns out a lot of people forget about it (see #604). This is a first step towards that.

mcmire:
> I don't know how hard it is, but I'd be interested in seeing the adapters restructured as other generators.

That's what I've called for too, and I think we're slowly getting there.

lib/suspenders/adapters/heroku.rb
```diff
@@ -0,0 +1,67 @@
...
+    class Heroku
+      def initialize(app_builder)
+        @app_builder = app_builder
+      end
+
+      def set_heroku_remotes
+        remotes = <<-SHELL.strip_heredoc
+
+          # Set up the staging and production apps.
+          #{join_heroku_app('staging')}
```
mcmire: Maybe `command_to_join_heroku_app`?

lib/suspenders/adapters/heroku.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      def heroku_app_name_for(environment)
+        "#{app_builder.app_name.dasherize}-#{environment}"
+      end
+
+      def generate_secret
+        SecureRandom.hex(64)
+      end
+
+      def run_heroku(command, environment)
```
mcmire: `run_heroku_command`? `run_heroku_subcommand`?

lib/suspenders/app_builder.rb
```diff
@@ -1,6 +1,14 @@
+require "forwardable"
+
 module Suspenders
   class AppBuilder < Rails::AppBuilder
     include Suspenders::Actions
+    extend Forwardable
+
+    def_delegators :heroku_adapter,
```
mcmire: I think we can also use the `delegate` helper that Rails provides here, for better readability.

spec/adapters/heroku_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        Heroku.new(app_builder).set_heroku_remotes
+
+        expect(app_builder).to have_received(:append_file).
+          with(anything, /heroku join --app #{app_name.dasherize}-production/)
+        expect(app_builder).to have_received(:append_file).
+          with(anything, /heroku join --app #{app_name.dasherize}-staging/)
+      end
+
+      it "setups the heroku specific gems" do
```
mcmire: "sets up" :)

lib/suspenders/adapters/heroku.rb
```diff
@@ -0,0 +1,67 @@
...
+
+          # Set up the staging and production apps.
+          #{join_heroku_app('staging')}
+          #{join_heroku_app('production')}
+        SHELL
+
+        app_builder.append_file "bin/setup", remotes
+      end
+
+      def setup_heroku_specific_gems
```
mcmire: This should really be named `set_up_heroku_specific_gems`. I know, grammar nazi :)

spec/adapters/heroku_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+
+      it "sets the heroku rails secrets" do
+        app_builder = double(app_name: app_name)
+        allow(app_builder).to receive(:run)
+
+        Heroku.new(app_builder).set_heroku_rails_secrets
+
+        expect(app_builder).to have_received(:run).
+          with(/SECRET_KEY_BASE.+staging/)
```
mcmire: Is it possible to use the full text of the command here?

spec/adapters/heroku_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  module Adapters
+    RSpec.describe Heroku do
+      it "sets the heroku remotes" do
+        app_builder = double(app_name: app_name)
+        allow(app_builder).to receive(:append_file)
+
+        Heroku.new(app_builder).set_heroku_remotes
+
+        expect(app_builder).to have_received(:append_file).
+          with(anything, /heroku join --app #{app_name.dasherize}-production/)
```
mcmire: If we care to write these tests, why don't we care about the first argument here?

spec/features/heroku_spec.rb
```diff
@@ -1,58 +1,73 @@
 require "spec_helper"
 
 RSpec.describe "Heroku" do
-  before do
-    drop_dummy_database
-    remove_project_directory
-  end
+  context "--heroku" do
+    before(:all) do
```
mcmire: Why move this from a `before :each` to a `before :all`? I feel like that would cause problems.

spec/support/suspenders.rb
```diff
@@ -12,9 +12,10 @@ def create_tmp_directory
   def run_suspenders(arguments = nil)
     Dir.chdir(tmp_path) do
       Bundler.with_clean_env do
-        ENV['TESTING'] = '1'
-
-        %x(#{suspenders_bin} #{APP_NAME} #{arguments})
+        `
```
mcmire: Perhaps use `system` here instead of backticks?

mcmire:
@zamith Sorry for the late review on this. I think this looks good overall, I just had a few suggestions/comments above.

lib/suspenders/adapters/heroku.rb
zamith: This was taken directly from the `app_builder`, I didn't change any of the method names, but I'm more than willing to change them if this is something you feel should all be done at this time.

lib/suspenders/adapters/heroku.rb
zamith: Again, this came directly from the `app_builder`, but if we're changing it maybe `run_toolbelt_command` could work, since we already know it related to Heroku.

lib/suspenders/app_builder.rb
zamith: The thing is that `rails/delegation` (and ActiveSupport in general I guess) is not loaded at this point, at least from what I could gather from the tests. Do you feel it is worth loading it?

spec/adapters/heroku_spec.rb
zamith: :+1:

lib/suspenders/adapters/heroku.rb
zamith: That's fine. :)

spec/adapters/heroku_spec.rb
zamith: It is, but I tried to test as much as it was being tested in the integration test. However, even with this I have my doubts, because aren't we coupling the test to much to the implementation, by having the entire command written down twice?

spec/adapters/heroku_spec.rb
zamith: Again, I was unsure of how much coupling the tests and the implementation should have. This is being tested in the integration tests, however. I can add it here as well.

spec/features/heroku_spec.rb
zamith: The reason is the same of why this was done for the `new_project_spec`, and that is that running suspenders is slow, and thus we want to run it as few times as possibles. The rationale is that for each different configuration of the command we run it once and only once.

Since all the tests are doing is read only, it should never be too much of an issue, I think.

spec/support/suspenders.rb
zamith: Hum... I believe this was suggested by hound.

tute:
Rebased and squashed commits together. Mind double checking the commit message and we merge, @zamith? Thanks so much! :clap:

lib/suspenders/adapters/heroku.rb
tute: Given this is a refactor, it makes sense to me to rename variables or methods that can get better names. :+1:

lib/suspenders/adapters/heroku.rb
tute: Likewise, :+1: to renames if we find better ones!

lib/suspenders/app_builder.rb
mcmire: Oh, if that's the case, then no, this is fine.

spec/adapters/heroku_spec.rb
mcmire: You wouldn't really have the same command written down twice, because there's two variants of that command. You're unspooling the logic in these tests, basically.

I tend to make tests fairly readable on their own and try not to leave the reader guessing as to what's going on, so I guess that'd be my rationale for listing out the entire command. That's my personal opinion though.

@tute, do you have any concerns about the readability of this test?

spec/features/heroku_spec.rb
mcmire: Gotcha.

spec/support/suspenders.rb
mcmire: Weird... I wonder why.

lib/suspenders/adapters/heroku.rb
mcmire: Yeah, I'd be fine with `run_toolbelt_command`.

zamith:
@tute Does it not refer to the adapter organisation we're following on purpose? Is it clear enough from the code?

I'm still doing a commit with some small changes, but we can then squash it back with this one.

spec/adapters/heroku_spec.rb
zamith: Actually, I would need to know (or stub) the secret, in order to be able to write the full command down. I can, however, extract the regex somewhere else, to make the expectation more readable.

spec/support/suspenders.rb
zamith: Not really sure, but here's [the comment](https://github.com/thoughtbot/suspenders/pull/646#discussion_r43736087).

spec/adapters/heroku_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      it "sets the heroku rails secrets" do
+        app_builder = double(app_name: app_name)
+        allow(app_builder).to receive(:run)
+
+        Heroku.new(app_builder).set_heroku_rails_secrets
+
+        expect(app_builder).to(
+          have_configured_var("staging", "SECRET_KEY_BASE"),
```
mcmire: Nice, good idea :+1:

spec/support/suspenders.rb
mcmire: Okay.

Something I just noticed -- is `add_fakes_to_path` not working to set the `PATH`? Is this why you had to add `export PATH` here?

zamith: They both work. As far as I could tell, when the suspenders binary is running, it will use the `PATH` set here (and disregards the other, I think because it's a different thread or bash session, not sure). But for tests that don't actually run suspenders, like the one's in the adapter, then `add_fakes_to_path` comes into play.

If there's a better way to do this, I'll take it.

mcmire: Oh okay. I think that instead of having to set `PATH` inside of the command itself, you could set it outside. I think that if you set `ENV` outside of running a command, those environment variables will be set within the shell that the command runs inside of. I'm not 100% sure. But a cleaner way might be to simply call `add_fakes_to_path` before running the command.

zamith: Right. I just now noticed the `with_clean_env` call to `Bundler`. Calling `add_fakes_to_path` after that seems to work.

zamith:
@tute @mcmire I've rebased with master and squashed everything, do you have any further comments?

mcmire:
@zamith Nope, this looks good to me.

tute:
Thank you, @zamith! Now back to our good old Pipelines PR! :)

zamith:
Yay. :tada: 

Let's do that then. Also, we still need to move the rest of the heroku related functionality from the `app_builder` to the adapter. Do you want to do this now, or is this something that can be done piece by piece?

tute:
> we still need to move the rest of the heroku related functionality from the app_builder to the adapter. Do you want to do this now, or is this something that can be done piece by piece?

If you think it will make Heroku Pipelines easier, now sounds better. Otherwise, after that feature sounds best.

tute:
Thank you!! :)

PR #647, opened 2015-10-30
No need to configure action_mailer_host in staging

tute:
It inherits production's, which reads the same environmental variable.


geoffharcourt:
LGTM.

tute:
Thanks for checking Geoff!

PR #648, opened 2015-11-02
Config Circle to use rake binstub instead

sikachu:
There is no need to use `bundle exec` given we have binstub in `bin/`.


tute:
:+1:

PR #649, opened 2015-11-04
Changed order in customize_gemfile in app_generator

rentziass:
I changed the order of build inside customize_gemfile in app_generator.rb

Why:
I noticed that with a fresh gemset, apart from the dependencies that come with Suspenders, it was crashing during app generation while trying to configure simple_form, due to a missing "bundle install", which was meant to happen after that. I actually feel pretty dumb to open a pull request like this on such a great repo

### DIFF
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -58,7 +58,6 @@ module Suspenders
 
     def customize_gemfile
       build :replace_gemfile
-      build :configure_simple_form
       build :set_ruby_to_version_being_used
 
       if options[:heroku]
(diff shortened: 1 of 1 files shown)


tute:
Aha, my bad. I introduced this in https://github.com/thoughtbot/suspenders/commit/4b517a11. This brings back the warning messages, but solves the problem you noticed.

> I actually feel pretty dumb to open a pull request like this on such a great repo

If that were true what would be left for me, given I introduced this! Thanks for your patch, this is great. :heart:

rentziass:
I absolutely meant no offense, it's just that I consider myself quite a newbie when facing repos like this, I was quite a bit timorous when creating the pull request :smile:

zamith:
@rentziass All contributions are greatly appreciated. Thanks a lot for taking the time. :)

Issue #650, opened 2015-11-05
bundler audit scan reports successful with update failure of vulnerability db


eliotsykes:
`bin/rake bundler:audit` updates the vulnerability db then runs the audit scan. 

The update stage can fail (say due to restrictive firewall), but this will not stop the audit scan from reporting that it completed successfully.

One potential issue with this is that the audit will be reported as successful with an out-of-date vulnerability db.

More info at https://github.com/rubysec/bundler-audit/issues/114

Would a temporary fix would be welcome in suspenders to catch this issue until/if the issue is fixed in bundler-audit?

mcmire:
Definitely! Go right ahead.

tute:
> Would a temporary fix would be welcome in suspenders to catch this issue until/if the issue is fixed in bundler-audit?

I'd welcome it if it were simple. I personally am used to seeing bundler audit's successful output, and I see it many times a day. If I see it fail once I'll notice, and if it persists I will worry enough to take a look at what and why is happening.

Solving the issue for real (merging a newer version dependency when it's available with the expected exit code) is a certain :+1: from me.

tute:
Thanks for your input!

tute:
Closing in favor of PR, which will be simpler thanks to bundler-audit's merge of the fix! Thank you for your input.

PR #651, opened 2015-11-07
Fix bundler-audit to exit on update failure

eliotsykes:
Prevent bundle audit check from passing when the vulnerability db is
out-of-date due to a failed update. This fix is intended to be temporary
until a fix is released for:

https://github.com/rubysec/bundler-audit/issues/114

### DIFF
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -34,7 +34,7 @@ end
 group :development, :test do
   gem "awesome_print"
   gem "bullet"
-  gem "bundler-audit", require: false
+  gem "bundler-audit", ">= 0.5.0", require: false
   gem "dotenv-rails"
   gem "factory_girl_rails"
   gem "pry-byebug"


eliotsykes:
Candidate for fixing #650

templates/bundler_audit.rake
```diff
@@ -4,9 +4,26 @@ if Rails.env.development? || Rails.env.test?
   namespace :bundler do
     desc "Updates the ruby-advisory-db and runs audit"
     task :audit do
+      fix_bundler_audit_issue_114_to_exit_on_update_failure
       %w(update check).each do |command|
         Bundler::Audit::CLI.start [command]
       end
```
tute: We don't need this iterator here I think. Could it be only `Bundler::Audit::CLI.start "check"`?

templates/bundler_audit.rake
```diff
@@ -4,9 +4,26 @@ if Rails.env.development? || Rails.env.test?
...
     end
   end
+
+  # Temporary fix: https://github.com/rubysec/bundler-audit/issues/114
+  def fix_bundler_audit_issue_114_to_exit_on_update_failure
```
tute: I wouldn't be so specific in this method name. `exit_with_error_on_bundler_audit_inability_to_update` is a tiny step forward, I'm sure we can find a better one.

templates/bundler_audit.rake
```diff
@@ -4,9 +4,26 @@ if Rails.env.development? || Rails.env.test?
...
     desc "Updates the ruby-advisory-db and runs audit"
     task :audit do
+      fix_bundler_audit_issue_114_to_exit_on_update_failure
       %w(update check).each do |command|
         Bundler::Audit::CLI.start [command]
       end
     end
   end
+
+  # Temporary fix: https://github.com/rubysec/bundler-audit/issues/114
```
tute: We don't need this code comment. Let's leave all the context in the commit message.

templates/bundler_audit.rake
eliotsykes: Thanks for the feedback! 

It could be: `Bundler::Audit::CLI.start ["check"]`. Where's a good spot to put the call to `update`?

tute:
> It could be: `Bundler::Audit::CLI.start ["check"]`. Where's a good spot to put the call to `update`?

You already do it in the conditional: `if Bundler::Audit::Database.update!`.

templates/bundler_audit.rake
```diff
@@ -4,9 +4,25 @@ if Rails.env.development? || Rails.env.test?
...
     desc "Updates the ruby-advisory-db and runs audit"
     task :audit do
+      prepare_bundler_audit_to_exit_with_error_if_update_fails
       %w(update check).each do |command|
         Bundler::Audit::CLI.start [command]
       end
     end
   end
+
+  def prepare_bundler_audit_to_exit_with_error_if_update_fails
```
tute: What about `update_bundler_audit_or_exit_with_error`?

eliotsykes:
`prepare_bundler_audit_to_exit_with_error_if_update_fails` applies a patch to the existing `Bundler::Audit::CLI#update` method (currently) without invoking it.

eliotsykes:
In case its of use, here's a sister PR for bundler-audit showing the before/after `update` source https://github.com/rubysec/bundler-audit/pull/121/files

templates/bundler_audit.rake
```diff
@@ -4,9 +4,25 @@ if Rails.env.development? || Rails.env.test?
...
+    Bundler::Audit::CLI.class_eval do
+      def update
+        say "Updating ruby-advisory-db ..."
+        if Bundler::Audit::Database.update!
```
tute: Do we need the `Bundler::Audit::` prefix, given we are in the context of that class?

tute:
Looking good, thank you!

templates/bundler_audit.rake
eliotsykes: `update_bundler_audit_or_exit_with_error?` is a clearer name if the method is changed from patching `update` to running `Database.update!` etc. With patching, the code can take advantage of `say "...", :pretty_colors`.

templates/bundler_audit.rake
eliotsykes: Dropping the prefix sadly fails with `NameError: uninitialized constant Database`.

eliotsykes:
@tute thanks for all the feedback, happy to incorporate further changes if/when needed.

templates/bundler_audit.rake
```diff
@@ -4,9 +4,25 @@ if Rails.env.development? || Rails.env.test?
   namespace :bundler do
     desc "Updates the ruby-advisory-db and runs audit"
     task :audit do
+      prepare_bundler_audit_to_exit_with_error_if_update_fails
       %w(update check).each do |command|
         Bundler::Audit::CLI.start [command]
       end
```
tute: > It could be: `Bundler::Audit::CLI.start ["check"]`. Where's a good spot to put the call to `update`?

Don't we already do it in `if Bundler::Audit::Database.update!`?

tute:
`bundler-audit` has just merged in the commit: https://github.com/rubysec/bundler-audit/commit/96cc110974e2c26702fa99e22bc430a425696f4e. Should we point our dependency to that SHA, or later?

tute:
Ping @eliotsykes

eliotsykes:
Apologies for being a bit slow responding @tute

> bundler-audit has just merged in the commit: rubysec/bundler-audit@96cc110. Should we point our dependency to that SHA, or later?

I'd expect @postmodern would have a better idea answering this.

geoffharcourt:
@tute @eliotsykes this commit has been merged to master on bundler-audit, I think we can just pick a published version and optimistically pin to that.

tute:
> I think we can just pick a published version and optimistically pin to that.

Last version is 0.4.0, published in June 30, 2015 (doens't contain that commit).

geoffharcourt:
Ack, @tute, my mistake. I'm somewhat against pinning to a commit if we have reason to believe a new version might be released in the short-term.

tute:
Yay, bundler-audit just released v0.5! https://github.com/rubysec/bundler-audit/releases

@eliotsykes, want to update our dependency, so suspenders runs with the fixed version? Then rebase the commit, and we merge. Thank you!

tute:
Actually, we don't need it, as it will be installed by default. Thanks for your work and input! :tada: :smiley:

derekprior:
@tute if we want to ensure new projects have at least 0.5 then I think we need to specify that in the Gemfile. 

If I run suspenders on a machine with 0.4 then it may not update to 0.5.

tute:
> @tute if we want to ensure new projects have at least 0.5 then I think we need to specify that in the Gemfile.

That's right, thank you @derekprior. @eliotsykes, want to send the patch?

eliotsykes:
Thanks @tute - earlier commits removed, rebased, bundler-audit >= 0.5.0 now required

tute:
Thank you!

PR #653, opened 2015-11-10
Fix spring binstub test

graysonwright:
## Problem:

It appears that Spring has switched from using double-quotes in their
binstub to using single-quotes. The change has broken one of our tests,
where we are asserting that the binstub contains a line matching a
regular expression.

## Solution:

Update the regular expression to use single-quotes, to match those
created by the binstub generator.


geoffharcourt:
:ship:

tute:
:sheep:

PR #654, opened 2015-11-12
Raise on missing Sprockets assets in test

seanpdoyle:
When testing features, we'd prefer to be notified of asset pipeline
issues as early as possible.


mcmire:
Seems logical :+1:

tute:
Makes sense to me, though CI is red for this PR. After we fix it, :shipit:!

tute:
Restarted CI, didn't seem like a related issue. Merge away on green!

PR #655, opened 2015-11-13
Use Heroku Local (Forego) instead of Foreman

croaky:
Heroku Local has replaced Foreman in the Heroku Toolbelt.

https://devcenter.heroku.com/changelog-items/692

The command `heroku local` has replaced `foreman`.
Heroku Local makes use of [Forego] to accomplish its tasks
and it is faster and has better cross-platform support.

[Forego]: https://github.com/ddollar/forego

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -321,7 +321,7 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
       copy_file "puma.rb", "config/puma.rb"
     end
 
-    def setup_foreman
+    def set_up_forego
       copy_file 'sample.env', '.sample.env'
       copy_file 'Procfile', 'Procfile'
     end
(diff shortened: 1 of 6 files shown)


templates/sample.env
```diff
@@ -1,4 +1,4 @@
-# http://ddollar.github.com/foreman/
+# http://ddollar.github.com/forego/
```
mehlah: I guess this line can be :scissors:

lib/suspenders/app_builder.rb
```diff
@@ -304,7 +304,7 @@ def configure_puma
       copy_file "puma.rb", "config/puma.rb"
     end
 
-    def setup_foreman
+    def setup_forego
```
zamith: Should this be renamed to `set_up_forego`? /cc @mcmire

mcmire: Yes, please :)

tute:
After the grammar fix, LGTM!

templates/sample.env
croaky: Good catch, I just changed to link to the correct Forego GitHub page.

lib/suspenders/app_builder.rb
croaky: :paperclip: :100: :pencil2: :memo: I like it.

PR #657, opened 2015-11-14
Add foreman installer to bin/setup

toobulkeh:
Since it's explicitly excluded, it would be nice to include this in the bin/setup pattern.

### DIFF
--- a/templates/bin_setup.erb
+++ b/templates/bin_setup.erb
@@ -8,6 +8,7 @@ set -e
 
 # Set up Ruby dependencies via Bundler
 gem install bundler --conservative
+gem install foreman
 bundle check || bundle install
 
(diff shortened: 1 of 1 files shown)


toobulkeh:
This may not be ideal, since foreman recommends using foreman.pkg for Mac OS X environments, and suspenders is probably environment-agnostic.

https://github.com/ddollar/foreman#foreman

tute:
Hey, thanks for your contribution! We recently merged a change favoring `heroku local` over `foreman` though: https://github.com/thoughtbot/suspenders/pull/655, so we won't this anymore.

Thanks again for your contribution! :)

toobulkeh:
Doh my mistake.

PR #658, opened 2015-11-14
Use a gemset based on the app_name

toobulkeh:
I believe this requires rvm, but it might be a worthwhile addition for users with many projects in the same environment. I'm interested in your thoughts.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -227,7 +227,7 @@ end
     end
 
     def set_ruby_to_version_being_used
-      create_file '.ruby-version', "#{Suspenders::RUBY_VERSION}\n"
+      create_file '.ruby-version', "#{Suspenders::RUBY_VERSION}@#{app_name}\n"
     end
 
     def enable_database_cleaner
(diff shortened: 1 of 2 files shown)


tute:
Hi @toobulkeh, thanks for your contributions!

I never used myself gemsets. Do they protect us against potential problems? Why are they better than sharing libraries across applications?

Thanks!

toobulkeh:
https://rvm.io/gemsets/basics

Gemsets and Bundler are relatively redundant, but I've seen a few cases where we use gems that aren't managed by the Gemfile (like Foreman was) and have a need to manage specific versions separate between projects.

gylaz:
With `rbenv` I haven't had a need for gemsets in years.

tute:
> With `rbenv` I haven't had a need for gemsets in years.

Is this because of something specific to `rbenv`? If so, what?

zamith:
I don't use gemsets anymore as well, other than having to use `bundle exec` more often I haven't had any issues. But I'm interested in knowing what kind of issues could occur.

gylaz:
> Is this because of something specific to rbenv? If so, what?

Yes -- it doesn't support gemsets. I haven't missed them.

tute:
@toobulkeh given suspenders doesn't use foreman anymore, do you still see any benefit on adding this?

toobulkeh:
Foreman was just an example. I believe I've had it happen on other gems before. If it comes up again, I'll add an example here.

PR #661, opened 2015-11-16
Set default Heroku app to staging, when using Heroku

toobulkeh:
Per [Heroku Docs](https://devcenter.heroku.com/articles/multiple-environments) using the `--remote` flag is much preferred over the `--app` flag. This default sets all Heroku commands to use the staging app by default, and you're already able to use the `heroku --remote production` shortcut


tute:
This is great, thank you! Can you please add a test to cover that line?
Thank you! :0

tute:
This is great, thank you! Can you please add a test to cover that line?
Thank you! :smiley:

toobulkeh:
I'll see what I can do :)

tute:
Thank you! :clap: :smiley:

PR #664, opened 2015-11-17
Fix spring binstub test

toobulkeh:
It appears that spring has yet again changed their binstub loading technique.

References [a previous commit] (https://github.com/thoughtbot/suspenders/commit/cbb95b648173ff42fe0480f4bc273e9bd4ae7599) and  [the spring change](https://github.com/rails/spring/commit/b583da754ed5abad3f7da8b9b46de673bc10eada)

I am not positive about this multi-line pattern. I attempted a freespace regex mode, but the regex was just more confusing due to the amount of spaces in the lines.

I would probably recommend not matching so exactly if they continue to change this.

### DIFF
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -160,10 +160,18 @@ RSpec.describe "Suspend a new project with default configuration" do
   it "adds spring to binstubs" do
     expect(File).to exist("#{project_path}/bin/spring")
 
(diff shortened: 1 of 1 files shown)


jferris:
Are we be overly specific in these tests? Could we check the binstubs for the string "spring" to make sure they've changed? I'm not sure what regression we're protecting against besides making sure we don't accidentally stop initializing Spring.

toobulkeh:
That makes sense to me

tute:
Thank you! :)

PR #666, opened 2015-11-18
Make Forego port match APPLICATION_HOST port

### LINKED ISSUE #665: Forego uses different port than APPLICATION_HOST
6b7614a8dc7ff2a7dd32d087039e49fd5984a35d switched from Foreman to Forego, and removed the `.foreman` file that told Foreman to use port 3000. Forego uses port 5000 by default, but the Suspenders sample `.env` file still has `ASSET_HOST` and `APPLICATION_HOST` set to `localhost:3000`.

According to a42ec7e773ad79ee8a6d994012c9bd94810c43c2 using port 5000 is undesirable, so probably the solution would be to have Forego use port 3000 as Foreman previously did. It seems there is no Forego equivalent to the `.foreman` file, though there is an open PR at https://github.com/ddollar/forego/pull/56. For me, adding `PORT=3000` to the `.env` file worked, but I have a vague feeling this might conflict with other tools.

Any thoughts on the correct way to resolve this? I'd be willing to put together a PR.

grantovich:
Suspenders previously included a `.foreman` file which specified that Foreman should use port 3000, and not the default of 5000 (rationale in a42ec7e). The switch from Foreman to Forego in 6b7614a removed this file since Forego doesn't support it, causing a reversion to port 5000 and a mismatch with the `APPLICATION_HOST` and `ASSET_HOST` variables in the sample `.env` file.

Forego does support specifying a port via the PORT environment variable, so this change adds that variable to the sample `.env` file.

(closes #665)

### DIFF
--- a/templates/sample.env
+++ b/templates/sample.env
@@ -1,6 +1,7 @@
 # https://github.com/ddollar/forego
 ASSET_HOST=localhost:3000
 APPLICATION_HOST=localhost:3000
+PORT=3000
 RACK_ENV=development
 SECRET_KEY_BASE=development_secret
 EXECJS_RUNTIME=Node


tute:
@mike-burns, I remember you spotted conflicts when using the `HOST` env variable. This change introducing `PORT` env variable (after discussion in https://github.com/thoughtbot/suspenders/issues/665) looks good to me, and would like your review before merging. Thanks!

mike-burns:
LGTM.

tute:
Thank you!

PR #669, opened 2015-11-21
Add lang attribute to html element in layout

tute:
Identifying the language of the document is important for a number of
reasons:

* It allows braille translation software to substitute control codes
for accented characters, and insert control codes necessary to prevent
erroneous creation of Grade 2 braille contractions.
* Speech synthesizers that support multiple languages will be able to
orient and adapt to the pronunciation and syntax that are specific to
the language of the page, speaking the text in the appropriate accent
with proper pronunciation.
* Marking the language can benefit future developments in technology,
for example users who are unable to translate between languages
themselves will be able to use machines to translate unfamiliar
languages.
* Marking the language can also assist user agents in providing
definitions using a dictionary.

See http://www.w3.org/TR/WCAG20-TECHS/H57.html

### DIFF
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -129,6 +129,12 @@ RSpec.describe "Suspend a new project with default configuration" do
     expect(prod_env_file).not_to match(/"HOST"/)
   end
 
+  it "configures language in html element" do
+    layout_path = "/app/views/layouts/application.html.erb"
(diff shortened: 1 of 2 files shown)


tute:
cc @tysongach for reviews. Will continue to add commits in this branch as we find places where we can improve during the workshop.

tysongach:
:tada: :tada: :tada:

tute:
cc @ckundo for review of his suggestion to use `I18n.locale`

PR #670, opened 2015-11-25
Copy sample.env to .env on generation

brandoncordell:
I know issue #668 said we should generate .env *instead* of .sample. but this would break the behavior of bin/setup. Since .env is ignored by git, all other collaborators wouldn't have a base env file to copy. Instead I added an extra step to copy .env into the generated suspenders project as well as .sample.env.


lib/suspenders/app_builder.rb
```diff
@@ -330,8 +330,9 @@ def configure_puma
     end
 
     def set_up_forego
-      copy_file 'sample.env', '.sample.env'
-      copy_file 'Procfile', 'Procfile'
+      copy_file "sample.env", ".sample.env"
```
tute: It looks like we don't need this line. Want to remove and squash your commits together?

tute:
> this would break the behavior of bin/setup. Since .env is ignored by git, all other collaborators wouldn't have a base env file to copy

We can remove that step, as well. Soon @geoffharcourt will merge https://github.com/thoughtbot/dotfiles/pull/442, which will make our dotfiles not ignore `.env`.

Thank you!

brandoncordell:
@tute Cool, thanks! I'll get that taken care of and pushed up.

brandoncordell:
@tute with `.ctags` located in `templates/dotfiles/` I feel we should move `templates/sample.env` to `templates/dotfiles/.env`. What do you think?

tute:
> @tute with .ctags located in templates/dotfiles/ I feel we should move templates/sample.env to templates/dotfiles/.env. What do you think? This will allow us to use the existing copy_dotfiles method, which looks to copy the entire templates/dotfiles directory.

Makes sense to me :+1:

tute:
Looks good! Can you please rebase so I merge? Thank you!

brandoncordell:
rebasing upstream/master is diverging master from origin/master. I can't pull from there because it can't fast-forward. Am I doing something wrong?

tute:
It was all right, it had merge conflicts. I solved them, and merged into master. Thank you!

brandoncordell:
Oh cool, thanks @tute!

lpender:
:+1: thanks guys!

PR #671, opened 2015-11-27
Switch Airbrake to Honeybadger

### LINKED ISSUE #667: Switch from Airbrake to Honeybadger
Consensus at thoughtbot seems to be that Honeybadger provides a better experience than Airbrake.

This should be encoded into suspenders, by making it default to Honeybadger instead of Airbrake.

brandoncordell:
[closes #667]


tute:
Will hold off on merge until we get to install it in an app and extract and compare the solution, or @calebthompson tells us that this is the best implementation for suspenders.

Thank you both!

calebthompson:
Yeah that looks about right, but I might still wait to see it in an app.

johnpaulashenfelter:
As a client of Thoughtbot, I can only :+1: this so many times. Airbrake is my least favorite choice. I prefer BugSnag but HoneyBadger might also let you make the change from NewRelic to Skylight with more confidence (https://github.com/thoughtbot/suspenders/pull/692) since bigger HoneyBadger plans also include some tools for performance.

calebthompson:
:shipit:

lib/suspenders/generators/app_generator.rb
```diff
@@ -239,7 +239,12 @@ def remove_routes_comment_lines
 
     def outro
       say 'Congratulations! You just pulled our suspenders.'
-      say "Remember to run 'rails generate airbrake' with your API key."
+      say "Remember to run 'bundle exec honeybadger install' with your API key."
```
mcmire: With the `--heroku` option specified, this would say:

```
Remember to run 'bundle exec honeybadger install' with your API key.
Run 'bundle exec honeybadger heroku install' with your API key unless you're using the Heroku Honeybadger add-on.
```

Is this necessary?

lib/suspenders/generators/app_generator.rb
```diff
@@ -239,7 +239,12 @@ def remove_routes_comment_lines
...
+
+      if options[:heroku]
+        say "Run 'bundle exec honeybadger heroku install' with your API key
+        unless you're using the Heroku Honeybadger add-on"
```
mcmire: Assuming this message is necessary (see above), can you add a period at the end of this?

lib/suspenders/generators/app_generator.rb
brandoncordell: You're right that doesn't read well. `Remember to run 'bundle exec honeybadger install' with your API key.` should probably be an else condition to `if options[:heroku]` instead or maybe `if options[:heroku]` could print something like "This isn't necessary if you're using the Heroku Honeybadger add-on"?

Which do you think is better?

lib/suspenders/generators/app_generator.rb
```diff
@@ -239,7 +239,13 @@ def remove_routes_comment_lines
 
     def outro
       say 'Congratulations! You just pulled our suspenders.'
-      say "Remember to run 'rails generate airbrake' with your API key."
+
+      if options[:heroku]
+        say "Run 'bundle exec honeybadger heroku install' with your API key
+        unless you're using the Heroku Honeybadger add-on."
```
mcmire: I don't know how this message will appear, but judging by the indentation here, it may look like:

```
Run 'bundle exec honeybadger heroku install' with your API key
        unless you're using the Heroku Honeybadger add-on.
```

What do you think about using a heredoc to ensure that this message is indented correctly?

brandoncordell: @mcmire Good catch, let me take a look!

lib/suspenders/generators/app_generator.rb
```diff
@@ -239,7 +239,7 @@ def remove_routes_comment_lines
 
     def outro
       say 'Congratulations! You just pulled our suspenders.'
-      say "Remember to run 'rails generate airbrake' with your API key."
+      honeybadger_outro
```
brandoncordell: @mcmire What do you think of this approach?

geoffharcourt: I'd rather that be called `say_honeybadger_outro` or else rephased as `say honeybadger_outro` to communicate what it's doing.

brandoncordell: Thanks for the suggestion @geoffharcourt, I'll get it cleaned up!

lib/suspenders/generators/app_generator.rb
```diff
@@ -239,7 +239,7 @@ def remove_routes_comment_lines
 
     def outro
       say 'Congratulations! You just pulled our suspenders.'
-      say "Remember to run 'rails generate airbrake' with your API key."
+      say_honeybadger_outro
```
tute: Looking good! For consistency, let's leave the `say` call up here, and return a string from `honeybadger_outro`?

brandoncordell:
@tute I went to run my specs and I'm getting the follow warnings and error.

```
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.4-compliant syntax, but you are running 2.2.3.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```

And then

```
/Users/brandoncordell/Code/suspenders/tmp/dummy_app/config/initializers/callback_terminator.rb:5:in `<top (required)>': undefined method `halt_callback_chains_on_return_false=' for ActiveSupport:Module (NoMethodError)
```

I've tried install ruby `2.2.4` and `2.3.0` but then I get a Ruby version mismatch between .ruby-version and the Gemfile, but what's weird is the version referenced in the "Gemfile" of the error seems to change to whatever version I'm *not* using.

I'm at a loss right now. The only thing I've changed recently is that I did a `brew update && brew upgrade` recently. I didn't run into this issue when I worked on #670

lib/suspenders/generators/app_generator.rb
```diff
@@ -253,5 +253,17 @@ def get_builder_class
...
+    private
+
+    def honeybadger_outro
+      msg = "Run 'bundle exec honeybadger heroku install' with your API key"
+
+      if options[:heroku]
+        msg << " unless you're using the Heroku Honeybadger add-on"
+      end
+
+      "#{msg}."
```
brandoncordell: I'm not sure I like how I'm building this outro message here. I want it to end in a period like the rest of the outro no matter which version of the message is used. Is there a cleaner way to do this?

geoffharcourt: You could make the second part of the message a private method that returns the `unless you're using...` string or a blank string, and then `#honeybadger_outro` could look like this:

```ruby
    def honeybadger_outro
      "Run 'bundle exec honeybadger heroku install' with your API key#{honeybadger_message_suffix}."
    end

    def honeybadger_message_suffix
      if options[:heroku]
        " unless you're using the Heroku Honeybadger add-on"
      end
    end
```

tute:
This looks good to me. @mcmire or @calebthompson, are you positive that this is all we need suspenders' side? I'd like to see a project generated on this branch and successfully talking with Honeybadger (and I won't have the time in the short term).

If it is ready, I'm good for merge!

calebthompson:
According to this, yes.

http://docs.honeybadger.io/lib/ruby.html

tute:
Thanks! Merging.

calebthompson:
Ping @starrhorne :tada:

toobulkeh:
Why require the command to be run? They support ENV variables appropriately:
> Don't like config files? You can place your API key in the `$HONEYBADGER_API_KEY` environment variable.

Or if the command is the preferred option (maybe for the Heroku integration), might I suggest including a .gitignore line for the yml file that setup command generates?

PR #676, opened 2015-12-03
A static page generator

mike-burns:
As discussed in #511, switch to using generators instead of an always-enabled command-line tool. This patch is the start to that.

It:

- Adds the `suspenders` gem to the `Gemfile`.
- Moves `high_voltage` into a `static` generator.
- Invokes the `static` generator as part of the `app` generator.

This makes it backward compatible with the existing executable, but also moves us in the direction of having separate generators for each component.

In theory, a user could add the suspenders gem to their Gemfile manually and then run the `suspenders:static` generator to get `high_voltage`. This is the final goal, afterall.


mcmire:
Neat!

So okay, as I understand, you aren't able to run the generator on its own *yet*. Is that right?

mike-burns:
Not _yet_ ,but as soon as I get these tests passing: yes, it will work on its own.

The tests are failing because the `gem "suspenders"` in the Gemfile is pointing to a gem that does not exist; it needs to point to the path. I am adding a `--path` flag to the generator to help with this.

mike-burns:
Next up: make sure Travis's bundler knows about all the gems it needs in order to run these tests.

tute:
I love the idea of moving to this pattern, a feature at a time, and add them with generators to new apps but also existing apps.

zamith:
I really like this idea in general. One thing I don't think I've seen discussed in this or the referenced PR, is something like a `suspenders add default` or something like that, that would pretty much run the generators for all the things that suspenders installs right now. 

Is this a feature that is interesting, or the reasoning is that there is no "default" app, just a different combinations of these smaller opinions (what authentication gem to use, what background jobs, etc...)?

mike-burns:
@zamith I intend to add a `suspenders:all` generator. So like this:

```sh
bin/rails g suspenders:all
```

zamith:
@mike-burns That's cool with me. Definitely agree with @tute that moving this a feature at a time is a good idea, to start driving adoption and getting feedback, before completing such a major change.

jferris:
The `static` generator is run by doing `bin/rails generate suspenders:static`, right?

This is great.

mike-burns:
@jferris yes exactly.

mike-burns:
Help wanted: get this to pass on CI, please.

(I can't even get it to pass on my work laptop, but it does pass on my home laptop. The differences between the machines are many.)

mike-burns:
I do not understand why this is freezing on CI. I need a Travis CI expert.

tute:
> I do not understand why this is freezing on CI. I need a Travis CI expert.

This is also happening in my rails 5 branch. I retriggered the builds with no cache and we got the same behavior. Will debug this.

mike-burns:
> Will debug this.

You are my hero. Thank you thank you.

tute:
In my branch CI stalls when `copy_file` asks for user input to replace a preexisting file. Removing it before will unstuck it.

Maybe this branch asks for user input as well?

mike-burns:
Which file? How did you discover this?

Should we set option `:force`? Is that how we override this?

```ruby
class Foo < Thor
  option :force
end
```

I think `#copy_file` comes from Thor...

tute:
It does come from Thor. I discovered this cancelling a hanging test run, which gave me the stacktrace:

```
suspenders ribbon % be rake
/Users/tutec/.rbenv/versions/2.3.1/bin/ruby -I/Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/lib:/Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-support-3.4.1/lib /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/rspec-core-3.4.4/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb

Randomized with seed 25049
....[Simple Form] Simple Form is not configured in the application and will use the default values. Use `rails generate simple_form:install` to generate the Simple Form configuration.
[Simple Form] Simple Form is not configured in the application and will use the default values. Use `rails generate simple_form:install` to generate the Simple Form configuration.

RSpec is shutting down and will print the summary report... Interrupt again to force quit.
/Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/line_editor/readline.rb:20:in `readline': Interrupt
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/line_editor/readline.rb:20:in `readline'
rake aborted!
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/line_editor.rb:7:in `readline'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/shell/basic.rb:395:in `ask_simply'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/shell/basic.rb:68:in `ask'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/shell/basic.rb:238:in `block in file_collision'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/shell/basic.rb:237:in `loop'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/shell/basic.rb:237:in `file_collision'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/actions/create_file.rb:99:in `force_on_collision?'
Interrupt:
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/actions/create_file.rb:92:in `force_or_skip_or_conflict'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/actions/create_file.rb:76:in `on_conflict_behavior'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/actions/empty_directory.rb:113:in `invoke_with_conflict_check'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/actions/create_file.rb:60:in `invoke!'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/actions.rb:94:in `action'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/actions/create_file.rb:25:in `create_file'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/actions/file_manipulation.rb:27:in `copy_file'
        from /Users/tutec/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-5.0.0.beta4/lib/rails/generators/rails/app/app_generator.rb:17:in `copy_file'

        from /Users/tutec/Sites/opensource/suspenders/suspenders/lib/suspenders/app_builder.rb:328:in `configure_puma'
        ...
 ```

`puma.rb` was in my case existing. You can also find out running the following script in a parent directory:

```
./suspenders/bin/suspenders test-`date +"%Y%m%d%H%M%S"`
cd test-*
./bin/setup
rails g scaffold User email password
rake db:migrate
git add .
git commit -m "Initial commit"
```

If it asks for input, it will do so in the tests as well.

**Comment by mike-burns** [comment]
I cannot reproduce it locally from the tests, but on your advice I inserted more print debugging and found that it is hanging in `spec/features/heroku_spec.rb:4` on CI.

I tried running it locally -- `./suspenders/bin/suspenders test-`date +"%Y%m%d%H%M%S"` --heroku=true` -- but it does not hang locally.

Now I know that I need more print-debugging in the Heroku adapter.

**Comment by mike-burns** [comment]
After much more debugging, this is what I get on CI. Locally it gets here but keeps going.
[hanging-heroku-static.log.txt](https://github.com/thoughtbot/suspenders/files/253390/hanging-heroku-static.log.txt)

**Comment by mike-burns** [comment]
Thanks for the inspiration, @tute. It was Spring, in this case.

(I likely did not encounter this locally because I have `DISABLE_SPRING` set. I'm not willing to ever unset that so I can't confirm. Put it on my tombstone: `env DISABLE_SPRING=1 echo Here lies Mike. RIP.`)

File: lib/suspenders/generators/app_generator.rb [hunk 1]
```diff
@@ -52,10 +55,12 @@ def suspenders_customization
       invoke :setup_segment
       invoke :setup_bundler_audit
       invoke :setup_spring
+      invoke :generate_all
```
mike-burns: It's more accurate for the future if we call this `#generate_default` instead of `#generate_all`.

tute: > It's more accurate for the future if we call this #generate_default instead of #generate_all.

:+1:

lib/suspenders/app_builder.rb
```diff
@@ -240,6 +240,17 @@ def create_database
       bundle_command 'exec rake db:create db:migrate'
     end
 
+    def replace_gemfile(path)
+      remove_file 'Gemfile'
+      template 'Gemfile.erb', 'Gemfile' do |content|
```
tute: We can use `template force: true` here, and then we don't need `remove_file`.

lib/suspenders/app_builder.rb
```diff
@@ -240,6 +240,17 @@ def create_database
       bundle_command 'exec rake db:create db:migrate'
     end
 
+    def replace_gemfile(path)
+      remove_file 'Gemfile'
```
tute: Should we follow our double-quote standard in new code?

lib/suspenders/app_builder.rb
```diff
@@ -342,6 +353,7 @@ def setup_stylesheets
 
     def install_refills
       generate "refills:import", "flashes"
+      run "rm app/views/refills/_flashes.html.erb"
```
tute: Do we need to call this here? It loos like the subsequent `remove_dir` would take care of this already.

lib/suspenders/generators/static_generator.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails/generators'
```
tute: > Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

Let's follow the double-quote standard in new files.

templates/Gemfile.erb
```diff
@@ -41,6 +40,7 @@ group :development, :test do
   gem "pry-byebug"
   gem "pry-rails"
   gem "rspec-rails", "~> 3.4.0"
+  gem "refills"
```
tute: Do suspended applications need `refills` in the `test` block?

tute:
In an application generated with this branch (with the suspenders gem pointed to the path) I can't get the new tasks to list with `rake -T`, nor can I run `suspenders:static`. Am I missing something while trying it out?

mike-burns:
It's a generator, so you want to try `rails generate suspenders:static`.

That said: I also don't get the `suspenders:static` generator. Thanks for trying this, we'll need to fix that.

tute:
I do get them:

```diff
diff --git a/Gemfile.lock b/Gemfile.lock
index d14b331..0a12e04 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,11 @@
+PATH
+  remote: ../suspenders
+  specs:
+    suspenders (1.38.1)
+      bitters (~> 1.3)
+      bundler (~> 1.3)
+      rails (~> 4.2.0)
```

```
$ rails g
# [...]
Suspenders:
  suspenders:app
  suspenders:static
```

tute:
I'm loving this! `suspenders:static` works, but not `suspenders:app` yet:

```
rails g suspenders:app
Running via Spring preloader in process 51856
Usage:
  rails new APP_PATH [options]

Options:
  -r, [--ruby=PATH]                                      # Path to the Ruby binary of your choice
                                                         # Default: /Users/tutec/.rbenv/versions/2.3.1/bin/ruby
  -m, [--template=TEMPLATE]                              # Path to some application template (can be a filesystem path or URL)
      [--skip-gemfile], [--no-skip-gemfile]              # Don't create a Gemfile
  -B, [--skip-bundle], [--no-skip-bundle]                # Don't run bundle install
                                                         # Default: true
  -G, [--skip-git], [--no-skip-git]                      # Skip .gitignore file
      [--skip-keeps], [--no-skip-keeps]                  # Skip source control .keep files
  -O, [--skip-active-record], [--no-skip-active-record]  # Skip Active Record files
  -S, [--skip-sprockets], [--no-skip-sprockets]          # Skip Sprockets files
      [--skip-spring], [--no-skip-spring]                # Don't install Spring application preloader
  -d, [--database=DATABASE]                              # Configure for selected database (options: mysql/oracle/postgresql/sqlite3/frontbase/ibm_db/sqlserver/jdbcmysql/jdbcsqlite3/jdbcpostgresql/jdbc)
                                                         # Default: postgresql
  -j, [--javascript=JAVASCRIPT]                          # Preconfigure for selected JavaScript library
                                                         # Default: jquery
  -J, [--skip-javascript], [--no-skip-javascript]        # Skip JavaScript files
      [--dev], [--no-dev]                                # Setup the application with Gemfile pointing to your Rails checkout
      [--edge], [--no-edge]                              # Setup the application with Gemfile pointing to Rails repository
      [--skip-turbolinks], [--no-skip-turbolinks]        # Skip turbolinks gem
                                                         # Default: true
  -T, [--skip-test-unit], [--no-skip-test-unit]          # Skip Test::Unit files
                                                         # Default: true
      [--rc=RC]                                          # Path to file containing extra configuration options for rails command
      [--no-rc], [--no-no-rc]                            # Skip loading of extra configuration options from .railsrc file
  -H, [--heroku], [--no-heroku]                          # Create staging and production Heroku apps
      [--heroku-flags=HEROKU_FLAGS]                      # Set extra Heroku flags
  -G, [--github=GITHUB]                                  # Create Github repository and add remote origin pointed to repo
      [--path=PATH]                                      # Path to the gem

Runtime options:
  -f, [--force]                    # Overwrite files that already exist
  -p, [--pretend], [--no-pretend]  # Run but do not make any changes
  -q, [--quiet], [--no-quiet]      # Suppress status output
  -s, [--skip], [--no-skip]        # Skip files that already exist

Rails options:
  -h, [--help], [--no-help]        # Show this help message and quit
  -v, [--version], [--no-version]  # Show Rails version number and quit

Description:
    Create suspenders files for app generator.
```

mike-burns:
`suspenders:app` isn't supposed to be there; I should figure out how to hide it for now.

It turns out suspenders has always been a generator, so adding it to your Gemfile will make it appear as a valid generator. This will not work until another hundred or so PRs makes it work.

lib/suspenders.rb
```diff
@@ -1,5 +1,6 @@
 require 'suspenders/version'
 require 'suspenders/generators/app_generator'
+require 'suspenders/generators/static_generator'
```
tute: Should these generators be required in `suspenders/generators/app_generator.rb`?

lib/suspenders/generators/app_generator.rb
```diff
@@ -233,6 +238,14 @@ def remove_routes_comment_lines
       build :remove_routes_comment_lines
     end
 
+    def generate_all
+      run("spring stop")
```
tute: Why do we need to add this?

mike-burns: This line took me five months to find. I have nothing polite to say about this line.

I'm keeping it as a separate commit when I merge: https://github.com/thoughtbot/suspenders/pull/676/commits/406eedafdf684420f25b15a38e4c64500d584eb4

tute: It looks like people chiming into suspenders are typically against `spring`, yet we never managed to drop it.

mike-burns: I typed a longer thing, but instead I'll reiterate: I have nothing polite to say about this line, or about Spring.

lib/suspenders/app_builder.rb
mike-burns: Dunno; you tell me.

I always follow the existing coding style. A mix of coding style throughout a codebase seems odd to me. Maybe we should make a big commit to change this code style?

templates/Gemfile.erb
mike-burns: Suspenders' test suite itself needs `refills` in the `test` block.

Debugging trick for the test suite:

```diff
diff --git a/bin/suspenders b/bin/suspenders
index 787d6af..e0478a4 100755
--- a/bin/suspenders
+++ b/bin/suspenders
@@ -20,4 +20,14 @@ templates_root = File.expand_path(File.join("..", "templates"), File.dirname(__F
 Suspenders::AppGenerator.source_root templates_root
 Suspenders::AppGenerator.source_paths << Rails::Generators::AppGenerator.source_root << templates_root
 
+class Thor
+  module Shell
+    class Basic
+      def stdout
+        $stderr
+      end
+    end
+  end
+end
+
 Suspenders::AppGenerator.start
```

(I'd like to get that, or something more clear than that, into `master`, hidden behind e.g. `ENV['DEBUG_SUSPENDERS_TO_STDERR']`.)

lib/suspenders.rb
mike-burns: The generators need to be required in `lib/suspenders.rb` so that users can add `gem "suspenders"` to their `Gemfile` and this will give them the generators.

The _methods_ defined in `suspenders/generators/static_generator` are not called from within `suspenders/generators/app_generator.rb`, so we do not need to require it in there.

mike-burns:
Done:

- Hid the `suspenders:app` generator.
- Force `Gemfile` overwrite.
- Don't remove the file before removing the directory it is in, too.
- Rename `#generate_all` to `#generate_default`.
- Use double quotes in the new file.

Remaining: should we mix styles in existing files?

Anything else before I merge?

tute:
Thanks for the `Thor::Shell::Basic#stdout` tip. I see two things:

* `rails g` lists `suspenders:app` and not `:static`
* A generated app on this branch doesn't install high voltage

Everything else is 🎉

blackjid:
Wow, awesome, I really like this PR. I just finish reading the whole discussion in #511 and here, I'm super excited. One of the reasons we build http://github.com/platanus/potassium at platanus, instead of using suspenders, was because we wanted to be able to add features to a "suspended" project after the fact.

I really like the simplicity of the idea of just using generators.

I would love to talk and see a way to customize the bundled generators and add new ones without having to maintain a fork.

> The basic pattern for activating something should be adding a gem to Gemfile, bundle install, running a generator, and then pulling in our opinionated config.

What happens when some stuff just skips one of the steps in the pattern?? For example setting up CI, or Hound, Heroku... most of those does have configuration but no gem or generators... Where these kind of stuff are going to be added? in the base template or as a slimmed down generator??

mike-burns:
> I would love to talk and see a way to customize the bundled generators and add new ones without having to maintain a fork.

The way forward might be:

```ruby
gem 'suspenders'
gem 'potassium'
```

And:

```
rails g suspenders:static
rails g potassium:exception_notifier
rails g suspenders:style
```

(Or:

```
rails g potassium:default
```
)

That is, make your own generators, and the `potassium:default` generator could make use of the `suspenders:...` generators for the ones you want, and custom ones for the ones you want changed:

```ruby
require 'rails/generators'
module Potassium
  class DefaultGenerator < Rails::Generators::Base
    def do_it_all
      generate('suspenders:static')
      generate('potassium:exception_notifier')
      generate('suspenders:style')
      # ...
    end
  end
end
```

> What happens when some stuff just skips one of the steps in the pattern?

We'll have to figure it out when we get there, but my thinking is: each generator will be isolated and stand by itself, so it will need to ensure that all of its dependencies are met. This is true for configuration/non-gem generators, too.

mike-burns:
@tute I figured out the issue. In order to try this locally you need to pass the `--path` to the development suspenders gem. Here's how I did it:

```
../suspenders/bin/suspenders --path=/home/mike/suspenders tinder_for_dogs
```

Does that work? Should I squash and merge?

mike-burns:
I have opened https://github.com/thoughtbot/suspenders/pull/765 and https://github.com/thoughtbot/suspenders/pull/766 off of this branch.

tute:
> @tute I figured out the issue. In order to try this locally you need to pass the --path to the development suspenders gem.

Forgot about that. Ready for merge! 🎉  😃  ❤️

tute:
Oh! Before merging, should it run all generators when we call `suspenders`?

tute:
> Oh! Before merging, should it run all generators when we call suspenders?

It does, and I prefer it so, at least for now. Great to merge!

templates/Gemfile.erb
tute: `refills` is not alphabetically sorted.

mike-burns:
I re-ordered the Gemfile as requested, and merged. Whew.

PR #678, opened 2015-12-10
Generate `rack_mini_profiler`

seanpdoyle:
Uses [rack-mini-profiler][gem] gem to help debug slow queries /
requests.

Enable it by setting `ENV["RACK_MINI_PROFILER"] = 1`.

[gem]: https://github.com/MiniProfiler/rack-mini-profiler#rails

### DIFF
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -42,6 +42,10 @@ group :development, :test do
   gem "rspec-rails", "~> 3.4.0"
 end
 
+group :development, :staging do
+  gem "rack-mini-profiler", require: false
+end
+
 group :test do
   gem "capybara-webkit"
   gem "database_cleaner"
--- a/templates/rack_mini_profiler.rb
+++ b/templates/rack_mini_profiler.rb
@@ -1,0 +1,5 @@
+if ENV["RACK_MINI_PROFILER"].to_i > 0
+  require "rack-mini-profiler"
+
+  Rack::MiniProfilerRails.initialize!(Rails.application)
+end
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 master (unreleased)
 
 * Switch from Airbrake to Honeybadger
+* Generate applications with `rack_mini_profiler` (disabled by default, enabled
+  by setting `RACK_MINI_PROFILER=1`)
 
 1.35.0 (December 30, 2015)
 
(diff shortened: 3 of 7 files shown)


templates/Gemfile.erb
```diff
@@ -16,6 +16,7 @@ gem "normalize-rails", "~> 3.0.0"
 gem "pg"
 gem "puma"
 gem "rack-canonical-host"
+gem "rack-mini-profiler", require: false
```
gylaz: Should this be in the `dev/staging` group?

seanpdoyle: Would leaving the default value of `RACK_MINI_PROFILER=0` in production be enough?

tute: I'd rather put it in dev/staging group, so we don't load unnecessary code.

mcmire:
Seems fine to me.

templates/rack_mini_profiler.rb
```diff
@@ -0,0 +1,5 @@
+if ENV["RACK_MINI_PROFILER"].to_i > 0
```
derekprior: whats the scenario where you wouldnt want this on? Are we concerned with the visual pollution? We could start it hidden (option-p, IIRC, toggles it).

geoffharcourt:
Having not used this previously, I was very impressed with the tool on the project @seanpdoyle and I just worked on. +1

templates/rack_mini_profiler.rb
seanpdoyle: According to [the docs](https://github.com/MiniProfiler/rack-mini-profiler#configuration-options):

> toggle_shortcut (default Alt+P) - a jquery.hotkeys.js-style keyboard shortcut, used to toggle the mini_profiler's visibility. See https://github.com/jeresig/jquery.hotkeys for more info.

jferris: We've generally been moving towards using environment variables to configure changes between environments, vs configuring based on the environment name. We're experimenting with eliminating the `staging` Rails environment (as Heroku recommends), so this could be useful in that case.

seanpdoyle: @jferris has that change landed yet?

Is this PR good to go without it, or does it depend on the resolution of that PR?

jferris: That change hasn't landed yet, and I don't believe this pull request is being held up. I just meant that configuring things using environment variables (as the pull request is currently doing) seems good to me, because it will work well when there are no separate staging/production environments.

PR #681, opened 2015-12-18
View each failure in specs with multiple expects

gylaz:
When we write tests with multiple expects, only the first failure is
reported when the test fails. `aggregate_failures` allows us to see
any or all the failures that happened in a specific test.

https://www.relishapp.com/rspec/rspec-core/docs/expectation-framework-integration/aggregating-failures

### DIFF
--- a/templates/spec_helper.rb
+++ b/templates/spec_helper.rb
@@ -16,6 +16,10 @@ RSpec.configure do |config|
     mocks.verify_partial_doubles = true
   end
 
+  config.define_derived_metadata do |meta|
+    meta[:aggregate_failures] = true
+  end
+
   config.example_status_persistence_file_path = "tmp/rspec_examples.txt"
   config.order = :random
 end


mcmire:
Nice. :+1:

tute:
:+1:

zamith:
This is definitely cool in situations where you have multiple expectations, but is that a common thing for you? Is this more of a feature spec thing?

gylaz:
@zamith We don't typically subscribe to one expectation per test, and thus it isn't uncommon for us to have multiple expects in one test.

zamith:
Oh ok. Then this change makes even more sense. :+1:

jferris:
Have you tried this out on a project yet?

I frequently write specs where the first expectation would guarantee that the remaining expectations fail. For example:

``` ruby
expect(results).to be_present
expect(results.first.title).to eq("Hello")
expect(results.length).to eq(1)
```

There's no reason to run the second two expectations if the first one fails. Running all three would print several meaningless failures (`NoMethodError: undefined method 'title' for nil:Nilclass`) that you have to wade through to find the actual error.

I feel the same way about some of the examples given in the documentation:

``` ruby
expect(response.status).to eq(200)
expect(response.body).to eq('{"message":"OK"}')
```

If the response status is a redirect or something, there's little point in checking the body. You'd almost certainly get one useful failure (`Expected 200, got 302`) and one meaningless failure (`Expected {"message":"OK"}, got ""`).

I think it would be good to try this out on a project and see if it provides useful results before adding it to Suspenders. When trying it out, I'd be interested to know if it's acceptable to use `aggregate_failures` only for the specs where multiple, independent expectations are used, which I would guess is the exception and not the rule.

tute:
> I frequently write specs where the first expectation would guarantee that the remaining expectations fail.

I agree with this.

I haven't used this, and I don't feel I have much value to gain or lose from using it. Are we running an experiment, @gylaz? Should we hold off until we see results, close the PR, or merge it?

gylaz:
We are running this as an experiment on Hound. It's a nice feature, but as @jferris pointed out might not we as helpful if we are good about writing our expectations. However, I don't see any harm in having this enabled...

tute:
@gylaz it seems like nobody is strongly against or in favor of this. And I like resolving open issues/code submissions. :) Please, merge or close at will. Thank you!

gylaz:
I'm going to close for now.

PR #683, opened 2015-12-21
Remove database host config for Linux support

graysonwright:
Extracted from https://github.com/thoughtbot/administrate/pull/359.
Issue first reported in https://github.com/thoughtbot/administrate/issues/296.

## Problem:

Ubuntu (and possibly other Linux distributions)
don't recognize `localhost` as a valid host parameter
in `config/database.yml`.

Running `bin/setup` gives them the error:

```
Couldn't create database for {"adapter"=>"postgresql",
"database"=>"administrate-prototype_development",
"encoding"=>"utf8",
"host"=>"localhost",
"min_messages"=>"warning",
"pool"=>2,
"timeout"=>5000}
```

## Solution:

Remove the `host` key from `config/database.yml`.

Without the `host` key, OS X defaults to `localhost`
and Ubuntu defaults to its Linux-y equivalent.

### DIFF
--- a/templates/postgresql_database.yml.erb
+++ b/templates/postgresql_database.yml.erb
@@ -2,7 +2,6 @@ development: &default
   adapter: postgresql
   database: <%= app_name %>_development
   encoding: utf8
-  host: localhost
   min_messages: warning
   pool: <%%= Integer(ENV.fetch("DB_POOL", 5)) %>
   reaping_frequency: <%%= Integer(ENV.fetch("DB_REAPING_FREQUENCY", 10)) %>


tute:
LGTM. I'm surprised this didn't get an earlier report.

jferris:
Does this actually default to localhost, or does it default to using a local socket?

It used to use a local socket without any host specified, and many Postgres installs on OS X do not create a local socket. Before this is merged, I would test on some of the various ways people configure Postgres on OS X, such as Postgres.app and Homebrew.

zamith:
I tested it with Postgres.app on El Capitan and it seems to work.

tute:
With homebrew Postgres this works well, too.

Before merging I want to double check: @jferris do you see any other potential drawback to this? Thanks!

jferris:
This should be fine for Linux users. Most Postgres installs I've encountered on Linux listen on a local socket by default.

tute:
Thank you all!

PR #684, opened 2015-12-24
Fix/airbrake update

patbenatar:


patbenatar:
Oops, wrong merge base.

Issue #686, opened 2015-12-29
warning: constant ::TimeoutError is deprecated


delphaber:
https://github.com/thoughtbot/suspenders/blob/master/templates/errors.rb#L26 uses `TimeoutError` but now (ruby 2.3.0) that constant is deprecated.

Should I remove it from the array?

bernerdschaefer:
Good catch. It looks like it should be replaced with `Timeout::Error`, as is done for `HTTP_ERRORS` above it.

sublimecoder:
Just submitted a PR to fix this issue. Had it show up in a few of my suspenders generated apps I've been upgrading to ruby 2.3.0.

tute:
Thank you! Fixed with #688.

PR #688, opened 2015-12-29
Fix SMTP TimeoutError warning

sublimecoder:
Fixing issue mentioned in #686 
TimeoutError has been deprecated and in favor of using Timeout::Error in ruby 2.3.0

### DIFF
--- a/templates/errors.rb
+++ b/templates/errors.rb
@@ -23,7 +23,7 @@ SMTP_SERVER_ERRORS = [
   Net::SMTPAuthenticationError,
   Net::SMTPServerBusy,
   Net::SMTPUnknownError,
-  TimeoutError
+  Timeout::Error
 ]
 
 SMTP_CLIENT_ERRORS = [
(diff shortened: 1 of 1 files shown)


templates/errors.rb
```diff
@@ -23,7 +23,7 @@
   Net::SMTPAuthenticationError,
   Net::SMTPServerBusy,
   Net::SMTPUnknownError,
-  TimeoutError
+  Timeout::Error
```
bernerdschaefer: I think it makes sense to ignore this, since the rest of the file doesn't use that style.

gylaz: Should we strive to adhere to our style-guide as we touch lines? Eventually, this file would be inline with our preferred style, as opposed to using stale styles.

sublimecoder: I agree since the style guide enforces this the other lines are wrong and should be updated. I will add that change to this PR.

tute:
Thank you! :)

PR #690, opened 2015-12-30
Call create only once for the heroku pipeline

### LINKED ISSUE #689: Pipelines: separate pipeline apps created
Hey guys,

Thanks for all your hard work on v1.35.0, it's come a long way since #617. Some nice touches added, especially running Heroku commands against staging as default!

However, when creating a new app the staging and production apps are generated as usual, but two pipeline apps are also generated instead of just the one. One has the staging app set as staging, the other has the production app set as production...

Thought I'd report this behaviour so we can start debugging. Ping #639. I'll try and take a look in the new year.

zamith:
Why:

* The pipelines `create` command was being called twice, which creates
  two pipelines, one for each environment

* Creating the pipeline with one of the environments and then adding the
  other one

Fixes #689

### DIFF
--- a/spec/support/fake_heroku.rb
+++ b/spec/support/fake_heroku.rb
@@ -44,7 +44,7 @@ class FakeHeroku
 
   def self.has_setup_pipeline_for?(app_name)
     commands_ran =~ /^pipelines:create #{app_name} -a #{app_name}-staging --stage staging/ &&
-      commands_ran =~ /^pipelines:create #{app_name} -a #{app_name}-production --stage production/
+      commands_ran =~ /^pipelines:add #{app_name} -a #{app_name}-production --stage production/
   end
 
   def self.commands_ran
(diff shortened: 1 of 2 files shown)


spec/support/fake_heroku.rb
```diff
@@ -44,7 +44,7 @@ def self.has_configured_vars?(remote_name, var)
 
   def self.has_setup_pipeline_for?(app_name)
     commands_ran =~ /^pipelines:create #{app_name} -a #{app_name}-staging --stage staging/ &&
-      commands_ran =~ /^pipelines:create #{app_name} -a #{app_name}-production --stage production/
+      commands_ran =~ /^pipelines:add #{app_name} -a #{app_name}-production --stage production/
```
zamith: I'm not sure how to fix this without affecting readability.

tute:
Thank you! Merged in.

zamith:
:+1:

PR #691, opened 2016-01-04
Update Bitters to v1.2

tysongach:


tute:
If there's no markup to change, LGTM! :)

PR #692, opened 2016-01-07
Replace NewRelic with Skylight

jferris:
thoughtbot developers who have used Skylight on recent projects
universally prefer it to NewRelic.

https://trello.com/c/MvWCeBNB


geoffharcourt:
I had a fantastic experience with it on a long-running project that I was working alone on. It has great capabilities around finding expensive queries, N+1 stuff, etc. :+1:

lib/suspenders/app_builder.rb
```diff
@@ -113,10 +113,6 @@ def set_up_hound
       copy_file "hound.yml", ".hound.yml"
     end
 
-    def configure_newrelic
```
derekprior: should we generate the analogous skylight configuration so you know what ENV variables you need to provide? Maybe it could have comments on how to set things up?

jferris: The [Skylight docs](https://docs.skylight.io/getting-set-up/) say you don't need to do that:

> For Heroku, setup is as simple as setting the desired ENV vars and pushing your repo. Nothing fancy required!

I'm not experienced enough with Skylight to recommend options, but I'm open to suggestions.

gylaz: I can confirm, just having setup SkyLight on a project. All I had to do was add the gem and set the Heroku config variable. However, to enable SkyLight on staging (in a `staging` environment), I had to provide `config.skylight.environments << "staging"` in `config/environments/staging.rb`.

jferris: Interesting. Do we want Skylight on in staging by default?

derekprior: I'm not a big fan of having suspenders dependencies that you have to know to configure to use. 

If we think monitoring is required and we think skylight is the tool for that, then should we make it impossible to deploy without it? That was my thinking by suggesting the configuration file. While It's not *needed*, I think it is possible to use one and would let us use `ENV.fetch` so we get errors if we forget to set it up.

jferris: I think that makes sense. I'll look around for an application with this setup to use an example.

gylaz: > Do we want Skylight on in staging by default?

Probably not.

wagenet:
I'm one of the lead developers on Skylight. It looks like you're on a good track, but if you've got any questions or issues, please don't hesitate to ask.

wagenet:
Also, we're all really glad to hear that you enjoy Skylight!

lib/suspenders/app_builder.rb
toobulkeh: I've found it useful to find issues in staging and production. Why would you not use it in staging? Notifications / priority? Too much noise?

wagenet: The important thing is to make sure that different production environments don't share the same Skylight token. The steps for setting up staging are documented here (https://docs.skylight.io/getting-set-up/#setting-up-multiple-environments) and, as we're well aware, are less than ideal. given the extra effort to set up another environment it probably makes sense to not have it on by default.

That said, we are currently working on some stuff that should make multiple environments much more streamlined. Once this all lands then I think it makes perfect sense to enable staging by default as we'll be able to track separate environments with one token.

toobulkeh: Thanks for the pointers! I've been using it wrong then.

Would it be safe to say that once this is in place, staging by default would be useful and recommended by Skylight?

Not sure how far along you are with it, but in case it helps: 

I personally enjoy the UX perspective of Airbrake, Honeybadger and Segment, as they handle separate environments quite well in-app as well as their installation process (which are mainly different keys or auto-parsing ENV information). From a security standpoint, different keys per environment are ideal (we often can give access to a developer to all-expect-production environments), though not a major gain (it's very rare we've used that).

It looks like you're still in very early product mode with Skylight -- just Intercom and email for support -- if you host a public [product feedback forum](https://digitalocean.uservoice.com/forums/136585-digitalocean), I'd be happy to contribute.

wagenet: I think that's correct. If Skylight can track two environments by default then there's not really any good reason why you wouldn't want to go with that.

tute:
@jferris what's the status of this PR? Thanks!

jferris:
@tute I forgot about it, thanks for the reminder! I'll try and poke at it again soon.

tute:
Ping @jferris when you get some time! :)

tute:
@jferris, don't forget!

screen shot 2016-04-01 at 5 15 58 pm

toobulkeh:
That picture reminds me of a scene from GATTACA. ❤️

tute:
Ping @jferris! I never used Skylight myself yet, so I don't to take over your work.

jferris:
@tute I rebased my changes on top of the latest master.

The only outstanding issue is this comment from @derekprior:

> I'm not a big fan of having suspenders dependencies that you have to know to configure to use.

I agree with that comment: I don't like including gems included that silently do nothing until you realize you're supposed to set an API key or something.

However, that's also the way things were working before, with NewRelic, so this is not a regression; it replaces NewRelic with Skylight as the default and preserves the existing workflow.

I think it might be worth taking a pass in a second pull request that ensures all production-necessary services are configured and refusing to boot if they aren't. If we're okay with that approach, I think this pull request is ready to go as-is.

tute:
> However, that's also the way things were working before, with NewRelic, so this is not a regression; it replaces NewRelic with Skylight as the default and preserves the existing workflow.
> 
> I think it might be worth taking a pass in a second pull request that ensures all production-necessary services are configured and refusing to boot if they aren't. If we're okay with that approach, I think this pull request is ready to go as-is.

This all makes sense to me.

---

We can drop a spec:

```
  1) Suspend a new project with default configuration ensures newrelic.yml reads NewRelic license from env
     Failure/Error: newrelic_file = IO.read("#{project_path}/config/newrelic.yml")
     
     Errno::ENOENT:
       No such file or directory @ rb_sysopen - /home/travis/build/thoughtbot/suspenders/tmp/dummy_app/config/newrelic.yml
     # ./spec/features/new_project_spec.rb:96:in `read'
     # ./spec/features/new_project_spec.rb:96:in `block (2 levels) in <top (required)>'
Finished in 3 minutes 59.7 seconds (files took 0.16696 seconds to load)
40 examples, 1 failure
```

toobulkeh:
What about a warning on boot instead of a failure to boot?

```
ratings master % heroku local
[OKAY] Loaded ENV .env File as KEY=VALUE Format
[OKAY] Trimming display Output to 169 Columns
[WARNING] No skylight key found
[WARNING] No XXXXX key found
11:17:30 AM web.1 |  [87849] Puma starting in cluster mode...
11:17:30 AM web.1 |  [87849] * Version 3.0.2 (ruby 2.3.0-p0), codename: Plethora of Penguin Pinatas
11:17:30 AM web.1 |  [87849] * Min threads: 2, max threads: 2
11:17:30 AM web.1 |  [87849] * Environment: development
```

jferris:
@tute I pushed a commit to remove that test

@toobulkeh 

> What about a warning on boot instead of a failure to boot?

I would worry that nobody would see that failure. When deploying to Heroku, I'd probably see a warning that printed during the build, but maybe not, and I definitely wouldn't see one hiding in the production log.

tute:
Extracted a new issue to make this configuration required: https://github.com/thoughtbot/suspenders/issues/757

Joe, can you please add notes to https://github.com/thoughtbot/suspenders/blob/master/NEWS.md pelase? Then it's good to merge I think. Thank you!

jferris:
I updated NEWS and did a rebase/squash. I'm waiting for CI to pass.

tute:
Green CI Joe! 🎉

jferris:
@tute thank you for your many reminders. I finally merged this into master. :toot:

tute:
Thank you!

PR #693, opened 2016-01-08
Change SimpleCov output dir while run in CircleCI

delphaber:
Following [this guide](https://circleci.com/docs/code-coverage) I was able to integrate `simplecov` in `circleci`. Integration means I can see `simplecov`'s html report as artifact in `circleci`.

By the way I placed the new code inside `if ENV.fetch("COVERAGE", false)`. This means you should run your test with `COVERAGE=true rake`. Is it good for you?

In my suspeders' fork I run always `simple_cov`. So my `spec_helper` is:

```ruby
require "simplecov"

if ENV['CIRCLE_ARTIFACTS']
  dir = File.join(ENV['CIRCLE_ARTIFACTS'], "coverage")
  SimpleCov.coverage_dir(dir)
end

SimpleCov.start "rails"

[...]
```

### DIFF
--- a/templates/circle.yml.erb
+++ b/templates/circle.yml.erb
@@ -3,4 +3,4 @@ database:
     - bin/setup
 test:
   override:
-    - bin/rake
+    - COVERAGE=true bin/rake
--- a/templates/spec_helper.rb
+++ b/templates/spec_helper.rb
@@ -1,5 +1,11 @@
 if ENV.fetch("COVERAGE", false)
   require "simplecov"
+
+  if ENV["CIRCLE_ARTIFACTS"]
+    dir = File.join(ENV["CIRCLE_ARTIFACTS"], "coverage")
+    SimpleCov.coverage_dir(dir)
+  end
+
   SimpleCov.start "rails"
 end
 


mcmire:
Does this mean that CircleCI would need to run tests with `COVERAGE` set to `true`?

delphaber:
@mcmire yes and that's because I tried to keep thoughtbot code which use `COVERAGE`. If you tell me that you do not use `COVERAGE` anymore, I'll edit my PR :)

mcmire:
I'm not sure if we use `COVERAGE`. I would guess that we don't, since it isn't inside of the `circle.yml` that Suspenders generates. Should we? Anyone else have thoughts on this?

jferris:
It looks like it was added here: https://github.com/thoughtbot/suspenders/pull/524

I'm guessing the intended usage is that you can run `COVERAGE=true rake` to generate coverage locally.

cc @tute

derekprior:
I've never done that and don't have any ionterest in doing so either. So that's one data point for not needing to maintain it...

mcmire:
Yeah, I'm probably never going to use `COVERAGE=true` locally, either. Is this something we want to run on Circle though?

delphaber:
I think the real question should be: "does thoughbot want to always run coverage on development machine?"

derekprior:
I don't ever want to run coverage on my development machine. I want to leave that to CI if I want it at all. Coverage is something I'm generally only curious about when I come onto an application and then again if it becomes a problem. It's interesting to see it in large chunks over time, I suppose, but unintersting to see on every run.

I've never worked with anyone at thoughtbot who looked at the coverage locally. Not once.

jferris:
I also only run it when I think there's an issue and I want to know what the situation is with coverage.

It's somewhat nice to have the snippet in there so that you can easily look at the coverage when you want to, but I don't think it's worth keeping the configuration around. I'm fine having it run on CI.

mcmire:
Okay, in that case, can we tell Circle to set `COVERAGE` to `true`?

jferris:
I thought we had Circle set up to send coverage stats to Code Climate.

I'm not strongly for or against having these artifacts on CI, as I don't look at coverage frequently, but the configuration required is fairly small.

I haven't used coverage locally very often and I almost never look at coverage stats on Code Climate.

geoffharcourt:
I used to pay lots of attention to coverage stats. Now I would generally only use them if I were auditing or inheriting a legacy project and wanted to get a very quick sense the size coverage gap.

In the event I needed to get a coverage stat for one of the above scenarios, I would probably branch, install SimpleCov, run my test suite, and then remove the branch.

mcmire:
Alright. Are we interested in merging this PR, then?

delphaber:
If you want to merge this PR, then I have to edit circle.yml to add COVERAGE env variable (https://circleci.com/docs/configuration#environment)

tute:
> I'm guessing the intended usage is that you can run `COVERAGE=true rake` to generate coverage locally.

That's indeed my idea, but I agree that local coverage are not that useful in greenfield thoughtbot projects. Being it such simple configuration, I don't mind having it though. I never looked at reports in Code Climate nor CircleCI; every once in a while I use it locally to verify.

> I have to edit circle.yml to add COVERAGE env variable

I would add that indeed, and I'd be +1 on merging this if nobody feels strongly against it.

tute:
> I have to edit circle.yml to add COVERAGE env variable

@delphaber do you want to add this to your PR? Then rebase, and we'll merge. Thanks!

delphaber:
@tute sure thing! I can edit the PR this weekend.

bf4:
Dunno if anyone's come across this, but SimpleCov and CircleCI can't do parallelism very easily due to containerized builds without a shared filesystem :(

delphaber:
done!

tute:
Thanks!

PR #694, opened 2016-01-08
Remove CoffeeScript and add babel

MattMSumner:
Why:

* We prefer to write vanilla javascript with future features made
  available via a transpiler such as [babel](https://babeljs.io/).

This PR:

* Removes "coffee-rails" gem.
* Adds "sprockets-es6", this will be pulled into sprockets 4.x at which
  point this can be removed.
* Adds "sprockets" >= 3.0.0, a dependency for "sprockets-es6"


MattMSumner:
This is following a PR to guides suggesting we stop using CoffeeScript in favor of JavaScript with the babel transpiler: https://github.com/thoughtbot/guides/pull/387

mcmire:
Ok :disappointed:

calebthompson:
:-1: I'm a huge fan of coffeescript, and recent changes to JS only incorporate the worse features of CS into JS. It's a net detriment.

mcmire:
I like CoffeeScript too, and seriously lament the fact that it has certain nice things that JavaScript doesn't have. But if you're using a framework these days, then I wager you're probably going to use vanilla JavaScript. And if you're not using a framework, then you might not have that much JavaScript anyway, in which case it *might* not be that painful to use vanilla JS. Whether that's really true or not, I feel like there is a certain direction that the JavaScript community at large is going in, and I feel like if we don't jump on board now, we'll have to eventually. I know this is a bit of a touchy subject though.

calebthompson:
> But if you're using a framework these days,

Most of our Rails projects aren't using front end javascript frameworks (I assume this is what you mean) I think. Even if they are, if they're using the asset pipeline then it doesn't matter what language they're writing it in because it'll get compiled into JS.

BlakeWilliams:
Big :+1:. CoffeeScript isn't as relevant as it once was and isn't doing a great job at keeping up with JavaScript.

MattMSumner:
The past couple of Rails projects I've been on have used ES6 and not CoffeeScript. I found this easier to manage. There is more discussion on the research card the prompted this PR: https://trello.com/c/EGpNLWB1/393-es6-instead-of-coffeescript

christoomey:
I'm a :+1: on this as well. Admitedly I'm more focused on React and client side apps, but my take is that at least that community has moved nearly 100% to es6 and coffeescript is almost never used.

Beyond personal preferences, I would very uncomfortable starting a client app with coffeescript at this point. This is the main reason I support removing it from suspenders.

MattMSumner:
I'm going to wait on the outcome of https://github.com/thoughtbot/guides/pull/387 before moving this forward.

geoffharcourt:
I'm reluctant to move to Sprockets 3.x for another few weeks until the recent gem breakages on asset precompilation are resolved. Other than that, I'm OK with Suspenders incorporating these changes. If anyone wants to work with coffeescript in a project it's easy enough to change the app after generation.

tute:
Frameworks and browsers are increasingly adopting ES6 as the standard, [our guides prefer babel over coffeescript](https://github.com/thoughtbot/guides/pull/387), and it's easy to override this decision if that makes sense for a certain project or team. Should we now merge this change?

dakull:
Regarding `sprockets-es6` gem from their docs, maybe it need some updating?

>This plugin is primarily experimental and will never reach a stable 1.0. The purpose is to test out BabelJS features on Sprockets 3.x and include it by default in Sprockets 4.x.

Issue #695, opened 2016-01-08
Point default Hound configurations to our Guides repo


tysongach:
Some discussion over on https://github.com/thoughtbot/guides/pull/379 happened around pointing the [default Hound configuration](https://github.com/thoughtbot/suspenders/blob/master/templates/hound.yml) here in Suspenders to look at standardized linter configs in our Guides repo.

- This means we could establish standard configurations in our Guides repo, that match our style, right next to all of our other documentation regarding code style
- It also means new Suspenders projects will kick-off with Hound configured directly to our thoughtbot style, whereas right now it is whatever Hound is defaulted to (which we are split on whether that should be _our_ guidelines, or the default of each linter we use, or something else)

tysongach:
So I believe it would be something like this: (as an example)

```yaml
# hound.yml

scss:
  config_file: https://raw.githubusercontent.com/thoughtbot/guides/tg-scss-lint/style/sass/scss-lint-config.yml
  enabled: true
```

geoffharcourt:
This is a cool idea. I wonder if it might be best to freeze the styleguide at the time the repo is created (`curl` the file from Github and store it in the repo) rather than pointing Hound to a remote location. 

I'm advocating for this approach because I think it might be better for a project to have a consistent style over time rather than evolve over time.

If really do we want the style to be whatever our guides is currently doing, then your suggestion as is (with added styles for JS/Ruby) would be awesome.

tysongach:
@geoffharcourt Really good points! I like that.

Also, if you just duplicate the config from our Guides upon creation, you then have a ready-to-go config file that actually exists in your new project, so that you can just start tweaking it right away (for project-specific linter changes).

tute:
What is the status of this PR? Is it ok to close the issue in favor of discussion on the PR? Thank you both!

geoffharcourt:
@tute, let's close this and just keep the PR open.

PR #696, opened 2016-01-08
Download and use latest styleguides from thoughtbot

geoffharcourt:
When users build a new project, they should have styleguides encouraging
the latest best practices available for local linting as well as through
Hound.

This change downloads the latest Ruby, JavaScript, and Sass
styleguides that we have in our https://github.com/thoughtbot/guides
repo for immediate use. It uncomments the relevant Hound configuration
to enable their immediate use.

Thanks to @tysongach and @teoljungberg for the idea to incorporate these
stylesheets directly, suggested in #695 and thoughtbot/guides#379.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -375,6 +375,21 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
       directory("dotfiles", ".")
     end
 
+    def download_style_guides
+      download_file(
+        ".jscsrc",
+        "https://raw.githubusercontent.com/thoughtbot/guides/master/style/javascript/.jscsrc",
+      )
+      download_file(
+        ".rubocop.yml",
+        "https://raw.githubusercontent.com/thoughtbot/guides/master/style/ruby/.rubocop.yml",
+      )
+      download_file(
+        ".scss-lint.yml",
+        "https://raw.githubusercontent.com/thoughtbot/guides/master/style/sass/.scss-lint.yml",
+      )
+    end
+
     def init_git
       run 'git init'
     end
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -376,17 +376,10 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
     end
 
     def download_style_guides
+      download_file(config_url("thoughtbot/guides", "style/javascript/.jscsrc"))
+      download_file(config_url("thoughtbot/guides", "style/ruby/.rubocop.yml"))
       download_file(
-        ".jscsrc",
-        "https://raw.githubusercontent.com/thoughtbot/guides/master/style/javascript/.jscsrc",
-      )
-      download_file(
-        ".rubocop.yml",
-        "https://raw.githubusercontent.com/thoughtbot/guides/master/style/ruby/.rubocop.yml",
-      )
-      download_file(
-        ".scss-lint.yml",
-        "https://raw.githubusercontent.com/thoughtbot/guides/master/style/sass/.scss-lint.yml",
+        config_url("thoughtbot/guides", "style/sass/.scss-lint.yml"),
       )
     end
 
--- a/templates/hound.yml
+++ b/templates/hound.yml
@@ -1,17 +1,17 @@
 # See https://houndci.com/configuration for help.
 coffeescript:
-  # config_file: .coffeescript-style.json
+  config_file: .coffeescript-style.json
   enabled: true
 haml:
   # config_file: .haml-style.yml
   enabled: true
 javascript:
-  # config_file: .javascript-style.json
+  config_file: .javascript-style.json
   enabled: true
   # ignore_file: .javascript_ignore
 ruby:
-  # config_file: .ruby-style.yml
+  config_file: .ruby-style.yml
   enabled: true
 scss:
-  # config_file: .scss-style.yml
+  config_file: .scss-style.yml
   enabled: true
--- a/templates/hound.yml
+++ b/templates/hound.yml
@@ -1,17 +1,17 @@
 # See https://houndci.com/configuration for help.
 coffeescript:
-  config_file: .coffeescript-style.json
+  # config_file: .coffeescript-style.json
   enabled: true
 haml:
   # config_file: .haml-style.yml
   enabled: true
 javascript:
-  config_file: .javascript-style.json
+  config_file: .jscsrc
   enabled: true
   # ignore_file: .javascript_ignore
 ruby:
-  config_file: .ruby-style.yml
+  config_file: .rubocop.yml
   enabled: true
 scss:
-  config_file: .scss-style.yml
+  config_file: .sass-lint.yml
   enabled: true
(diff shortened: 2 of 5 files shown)


templates/hound.yml
```diff
@@ -1,17 +1,17 @@
...
+  config_file: .javascript-style.json
   enabled: true
   # ignore_file: .javascript_ignore
 ruby:
-  # config_file: .ruby-style.yml
+  config_file: .ruby-style.yml
   enabled: true
 scss:
-  # config_file: .scss-style.yml
+  config_file: .scss-style.yml
```
tysongach: We need to update this to `.scss-lint.yml`

I like specifically using this filename because you can run `scss-lint` from the command line (if you have it installed locally, obviously) and it defaults to looking for a `.scss-lint.yml` in the project root to base it's linting off of.

templates/hound.yml
```diff
@@ -1,17 +1,17 @@
 # See https://houndci.com/configuration for help.
 coffeescript:
-  # config_file: .coffeescript-style.json
+  config_file: .coffeescript-style.json
```
tysongach: Should this one remain commented out; is there no config we are pulling from thoughtbot/guides, as it doesn't exist?

templates/hound.yml
```diff
@@ -1,17 +1,17 @@
...
   # config_file: .haml-style.yml
   enabled: true
 javascript:
-  # config_file: .javascript-style.json
+  config_file: .javascript-style.json
   enabled: true
   # ignore_file: .javascript_ignore
 ruby:
-  # config_file: .ruby-style.yml
+  config_file: .ruby-style.yml
```
tysongach: Each of the `config_file` values in this file (or at least the ones we are actively pulling from thoughtbot/guides) should match the name we are passing from the download in `lib/suspenders/app_builder.rb`.

For example, this one should be `.rubocop.yml`…

_Or_, we change the download action to name this to `.ruby-style.yml`

tysongach:
THIS IS GONNA BE AWESOME

templates/hound.yml
geoffharcourt: The Ruby one should definitely be `.rubocop.yml`, as that's where Rubocop looks, and Ruby developers will think to look there.

templates/hound.yml
geoffharcourt: OK, cool. That matches the reasoning I have above for using the Rubocop default name.

geoffharcourt:
Great way to close out a Friday! :tada:

templates/hound.yml
tysongach: :+1:

templates/hound.yml
geoffharcourt: Good call.

lib/suspenders/app_builder.rb
```diff
@@ -375,6 +375,21 @@ def copy_dotfiles
       directory("dotfiles", ".")
     end
 
+    def download_style_guides
+      download_file(
+        ".jscsrc",
+        "https://raw.githubusercontent.com/thoughtbot/guides/master/style/javascript/.jscsrc",
```
teoljungberg: What do you think about extracting something alongside the lines of `config_url("thoughtbot/guides", "style/javascript/.jscsrc")` ?

mcmire:
Sweet!

lib/suspenders/app_builder.rb
geoffharcourt: @teoljungberg I had the same idea and then decided that I didn't want to box us in to using code from Github. Looking at it again, I think you've sold me.

geoffharcourt:
One issue I haven't yet resolved: the tests for this change are actually pulling the config from the web. My instinct here would be to use VCR, but that's a new development dependency. Any suggestions as to how to best make these tests not reliant on a working network connection?

gylaz:
@geoffharcourt Instead of VCR, how about using Webmock and returning a fixture file?

geoffharcourt:
@gylaz that works. Thanks for the recommendation!

geoffharcourt:
Gah! Here's where I'm at. Using WebMock (or VCR) doesn't seem to work here, I suspect because the feature execution is somehow outside of the hooks WebMock implements in RSpec. (I have WebMock running with `disable_net_connect!` but outside connections are still made. Any ideas how to get this working?

geoffharcourt:
One approach could be to use cURL instead of Net::HTTP to do the download (and then to fake/stub `curl` as we have `git` and `heroku`), but that seems like changing our implementation simply to make it testable. Is there something I'm missing that we can do?

teoljungberg:
@geoffharcourt I'll gladly pair with you on this on Friday, timezones apart. What do you think?

geoffharcourt:
@teoljungberg sounds great. I suspect I'm missing something straightforward, if I figure it out before then I'll let you know. Otherwise we'll get a time early on my Friday so our times line up.

Thanks!

tute:
I love this! :)

gylaz:
Just an FYI, we are working on supporting an organization-level config on Hound itself.

tute:
> we are working on supporting an organization-level config on Hound itself.

That would render this changeset unnecessary. @gylaz, does Hound already have this? Do you see it coming in the foreseeable futuer? Thank you!

gylaz:
We've started work on this front, but the progress slowed -- not a lot of investment availability and we are focusing on stability. I would say in the next few months is likely.

tute:
@geoffharcourt what is the status of this PR? Thank you!

geoffharcourt:
@tute we need to do some work on the specs to make them not dependent on external web connectivity.

templates/hound.yml
```diff
@@ -6,12 +6,12 @@ haml:
   # config_file: .haml-style.yml
   enabled: true
 javascript:
-  # config_file: .javascript-style.json
+  config_file: .jscsrc
```
thorncp: The `javascript` key is an alias for the JSHint linter, so giving it a config file for JSCS would probably cause Hound to fail. Typically people use both JSHint and JSCS. For that, hound.yml would look like:

```yaml
javascript:
  enabled: true
  config_file: .jshintrc

jscs:
  enabled: true
  config_file: .jscsrc
```

However, JSCS is [merging] (or has merged?) with ESLint, and Hound won't support JSCS for much longer because of that. It seems to me like ESLint is becoming the dominant JavaScript linter, so maybe it would make sense to configure that in hound.yml in this PR.

[merging]: http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint

```yaml
javascript:
  enabled: false

eslint:
  enabled: true
```

geoffharcourt: @thorncp what's the timetable for the merging as you understand it?

thorncp: I just realized my markdown was janky and the link was broken. [Fixed](http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint).

They didn't give any concrete timeline, but in an effort to push people away from it, JSCS has already been [unlisted](https://github.com/houndci/hound/pull/1129) from Hound's documentation.

> We expect this effort to take several months, and as such, we still encourage current JSCS users to upgrade to JSCS 3.0.0 and provide feedback to the team. We’ll announce when all of the changes necessary to support JSCS users in ESLint are complete and will start encouraging JSCS users to switch to ESLint at that time.

geoffharcourt:
It looks like this is going to be supplanted by (superior) Hound functionality in the future, so I'm closing this out.

PR #697, opened 2016-01-11
Some useful gems added, shoulda init bug fixed

kvokka:
Little description for gems I leave in gemfile (and i think it is good practics for some gems)

Shoulda patch added.

### DIFF
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -22,27 +22,37 @@ gem "sass-rails", "~> 5.0"
 gem "simple_form"
 gem "title"
 gem "uglifier"
+gem 'slim-rails'                            # Slim support
 
 group :development do
+  gem "meta_request"    # Rails meta panel in chrome console. https://chrome.google.com/webstore/detail/railspanel/gjpfobpafnhjhbajcjgccbbdofdckggg
+  gem "better_errors"   # Much more useful browser error message   
   gem "quiet_assets"
   gem "refills"
   gem "spring"
   gem "spring-commands-rspec"
   gem "web-console"
+  gem 'guard'
+  gem 'guard-livereload', '~> 2.4', require: false            # makes frontend developing more fun
+  gem 'rails_db'        # DB viewer on http://localhost:3000/rails/db
+  gem 'axlsx_rails'     # Addition for xls data export
 end
 
 group :development, :test do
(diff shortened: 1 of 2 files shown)


mike-burns:
A discussion about comments occurred in #511.

mcmire:
Thanks for the PR. Suspenders is opinionated, in that it reflects what we use day-to-day in our projects. So the gems that it includes come out of discussions we have internally on .

If you'd like to introduce a new gem that you've found useful, that's fine, but we likely won't accept a PR that introduces a bunch of stuff at once.

That said, here are some comments on some of them:

* We haven't added Haml or Slim because we're split across the company on which one we like (and some of us don't like either).
* `meta_request` (along with RailsPanel) looks nice. I haven't seen this before, so I'll have to try it out.
* `rails_db` is interesting, although I think that if I ever have to inspect anything, using the Rails console or dropping into the Postgres console works for me -- I imagine that my colleagues might feel the same way.
* I thought we already had Faker in the Gemfile. Out of all the gems, I'd vote we should add this, because it's definitely useful in factories.

mcmire:
Also, in terms of requiring Bundler in shoulda-matchers, you shouldn't have to do this. Was there an issue you ran into? If so, please file it on the [shoulda-matchers](http://github.com/thoughtbot/shoulda-matchers) repo.

kvokka:
it is good news to know, that sholda matchers need not more this addon, becourse i remember that i spent some time in solution search. You can see it in this link https://github.com/thoughtbot/shoulda/issues/203#issuecomment-170299361

And I understood your position about slim, but, maybe it will be good to add some questions for user about it, link not default settings?

Also, mee to isn`t shure about rails_db, but its pretty & looks unproblem. If one day it helps- perfect

I`ll remake gem and make another pull request. Thank you for cooperating

mike-burns:
I've never used Faker and can't imagine why I would want to in conjunction with FactoryGirl.

geoffharcourt:
In past use I found Faker to be more of a headache than it was worth. A thoughtfully-designed factory with sequences communicates more during a failure about what happened and how much data was created.

Using Faker also had the unintended side effect of creating real-looking data in a development database when `rake` was running specs in development mode. It's much better to have the factory generate almost-plausible data, but records that can clearly be traced back to a factory's work.

PR #700, opened 2016-01-12
User selection feature added, rubocop oops fixed

kvokka:
Added possibility of adding list of features on creation
It`s easy scalable. All hollywars gems moved to user choice section
I think to update it as much as I need

PS: rubocop conflict with hound in quotes selection
PSS: Ruby 2.3.0 support lay down today nigth on travis after update :(

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,4 @@ require 'rspec/core/rake_task'
 RSpec::Core::RakeTask.new(:rspec)
 
 desc 'Run the test suite'
-task :default => :rspec
+task default: :rspec
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,4 @@ require 'rspec/core/rake_task'
 RSpec::Core::RakeTask.new(:rspec)
 
 desc 'Run the test suite'
-task default: :rspec
+task :default => :rspec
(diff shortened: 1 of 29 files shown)


kvokka:
Ruby 2.3.0 support lay down today nigth on travis after update :(

zamith:
@kvokka I don't think this is the idea of suspenders. It is supposed to be opinionated and take away the decision of what to use from the developer. Furthermore, it should prompt the user for input as little as possible and hopefully not at all. I know at least @mike-burns is strongly opposed to this.

However, the approach we're taking is to extract most functionality into smaller generators that can be composed into a bigger opinionated generator, or run at will in a project. You can follow that feature in #676.

Hope this helps.

kvokka:
Sad, but I understand this position
I have another mind about this type of tool.
Thank you for your time

geoffharcourt:
@kvokka To be very clear, we appreciate the input. However, because there's a large number of people at our company and outside who rely on Suspenders, we are somewhat conservative about what gems we use and our strategy for getting the application generated.

An interactive Rails app generator might be a great idea for a separate project, maybe leveraging some parts of suspenders that already exist.

PR #701, opened 2016-01-13
Usefull flags added

kvokka:
Hello again
i made flag, for possibility of custom gems install
also, leave some pry gems be defoult

if start with out flags nothig changes, as was needed
but not there is the way to solve hollywars at the begining


mcmire:
Thanks for the PR, but there are too many changes here for us to consider.

Suspenders is a reflection of the tools that we use at thoughtbot, and as such, we don't usually add new gems until we've tried them out internally. Thanks for bringing `rails_db` to our attention -- we'll check it out.

PR #702, opened 2016-01-14
Remove the fix_i18n_deprecation_warning method

gssbzn:
`fix_i18n_deprecation_warning` is no longer needed since this
configuration is the default in Rails now

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -318,14 +318,6 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
       configure_environment "test", "config.active_job.queue_adapter = :inline"
     end
 
-    def fix_i18n_deprecation_warning
-      config = <<-RUBY
-    config.i18n.enforce_available_locales = true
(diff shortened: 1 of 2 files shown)


geoffharcourt:
Removing the workaround for the deprecation warning looks like a great idea, but can you remove the unrelated changes? We'd like to keep the focus of the change small. Some of these refactorings contradict our [style guide](https://github.com/thoughtbot/guides/tree/master/style/ruby), but in any event they should be in a separate PR/commit.

gssbzn:
@geoffharcourt done, I'll check the style guides to make another pull request with this fixes if that's ok
I did it in the same PR since some projects try to avoid PR with only code style fixes.

geoffharcourt:
@gssbzn thanks!

bernerdschaefer:
Source:

> Rails 4.1 now defaults the I18n option enforce_available_locales to true.
> -- http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#i18n-enforcing-available-locales

tute:
Thank you! Merged in. :)

PR #705, opened 2016-01-19
Disable CSS transitions in tests

gylaz:
CSS transitions often break tests. For example, a popup can take some
time to be fully shown if there is a transition time applied to its
show effect. The test will see that the popup is present and will try to
use it, but the element is ready yet.

This disables the CSS transitions in the test environment and removes
timing issues related to CSS.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -231,6 +231,13 @@ end
       copy_file '_javascript.html.erb', 'app/views/application/_javascript.html.erb'
     end
 
+    def create_shared_css_overrides
+      copy_file(
+        "_css_overrides.html.erb",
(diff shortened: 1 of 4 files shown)


mcmire:
This makes sense to me, but what do you think about putting this in a partial?

geoffharcourt:
Is there a possibility that a test would want to confirm that those transition properties are present that might fail because of this change?

bernerdschaefer:
This has definitely bitten me before! :+1:

tute:
:+1:

gylaz:
Extracted a file for CSS overrides.

geoffharcourt:
I like the partial. LGTM.

tute:
@gylaz :ship:!

PR #711, opened 2016-01-29
Fix refills installation to use bundle

gssbzn:
Refills generator was not using bundle, with internal generate command this should fix bundle problems


tute:
Merged in. Thank you! :)

PR #713, opened 2016-02-05
Remove `RAILS_ENV` declarations

seanpdoyle:
**DISCLAIMER:** I’m not sure if this is a good idea, just questioning the status quo

The proper way to access Rails' environment is through `Rails.env`,
which checks `RAILS_ENV`, falling back to `RACK_ENV`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,5 @@
+* Remove `RAILS_ENV` definitions
+
 1.36.0 (February 26, 2016)
 
 * Update Bitters to v1.2
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ This:
 
 * Creates a staging and production Heroku app
 * Sets them as `staging` and `production` Git remotes
-* Configures staging with `RACK_ENV` and `RAILS_ENV` environment variables set
+* Configures staging with `RACK_ENV` environment variable set
   to `staging`
 * Adds the [Rails Stdout Logging][logging-gem] gem
   to configure the app to log to standard out,
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -25,7 +25,7 @@ module Suspenders
       end
 
       def create_staging_heroku_app(flags)
-        rack_env = "RACK_ENV=staging RAILS_ENV=staging"
+        rack_env = "RACK_ENV=staging"
         app_name = heroku_app_name_for("staging")
 
         run_toolbelt_command "create #{app_name} #{flags}", "staging"
--- a/templates/app.json.erb
+++ b/templates/app.json.erb
@@ -17,9 +17,6 @@
     "RACK_ENV":{
       "required":true
     },
-    "RAILS_ENV":{
-      "required":true
-    },
     "SECRET_KEY_BASE":{
       "generator":"secret"
     },
--- a/templates/rails_helper.rb
+++ b/templates/rails_helper.rb
@@ -1,4 +1,4 @@
-ENV["RAILS_ENV"] = "test"
+ENV["RACK_ENV"] = "test"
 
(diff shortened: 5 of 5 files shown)


seanpdoyle:
The motivation behind this PR is that the less code we have to maintain, the better.

Why is it that we define both `RAILS_ENV` and `RACK_ENV` -- doesn't `RAILS_ENV` (and `Rails.env`) inherit the value from `RACK_ENV`?

geoffharcourt:
If this doesn't impact test environments at all I'm in favor. Less is better.

derekprior:
If we're going to set only 1 of them, would it make more sense for it to be `RAILS_ENV`? Why set the fallback and not the primary?

seanpdoyle:
@derekprior I had a hunch that some middleware in the stack is Rack-based and not Rails-based. 

If a middleware doesn't know anything about Rails, it wouldn't be checking `RAILS_ENV`.

derekprior:
Good point.

mike-burns:
I found where the build pack cares about `RACK_ENV`: https://github.com/heroku/heroku-buildpack-ruby/blob/master/lib/language_pack/rack.rb#L45

And `RAILS_ENV`: https://github.com/heroku/heroku-buildpack-ruby/blob/b017193b6774476d86bcbc87d526e79f541bf99a/lib/language_pack/rails4.rb#L27

Mostly dropping my knowledge, not any opinions.

geoffharcourt:
We're only concerned about what happens in development and test here, correct? Heroku's going to set these variables in production/staging/demo environments without our intervention.

mike-burns:
Two of the three Ruby files changed exist only for Heroku -- `heroku.rb` and `app.json`. So does this change affect Heroku?

geoffharcourt:
@mike-burns you are correct! I think until we resolve how staging environments get set up (Heroku discourages having environment names now), this in fact does alter Heroku configuration. Disregard my last comment.

tute:
@schneems can you please tell us if Heroku needs `RAILS_ENV`, or we can drop it in favor of `RACK_ENV`? And in general, why would we need the framework specific one, if rack's seems to suffice? Thank you in advance!

schneems:
RACK_ENV != RAILS_ENV if you want to be technical about it. The idea behind RACK_ENV is that "production" doesn't have any special debugging middleware, that's about it. t, while RAILS_ENV is like a feature flipper. Over they years the've been munged to basically mean the same thing. Heroku will set both to "production" so you don't have to do anything. If you're using a Rails app, I would set RAILS_ENV over RACK_ENV.

tute:
Thank you, Richard. I don't see why would we need both, and when would we set them differently.

I have never seen that happen at thoughtbot yet, so I personally :+1: this change.

tute:
And anyway, in Heroku `RAILS_ENV` will be set. Want to merge, @seanpdoyle?

seanpdoyle:
@tute merge as is (after rebase), preferring `RACK_ENV` over `RAILS_ENV`?

tute:
I just thought that before merging I should test this in my projects, to see if it has any unintended consequences. If you have been using it already then yes, otherwise, I'll test it and get back to this. Thank you!

seanpdoyle:
@tute I haven't yet tested this -- go ahead.

tute:
Now I'm testing this on a project, and test, review apps and staging work well. Want to rebase and merge, @seanpdoyle?

tute:
Yay! Thank you

tute:
We just saw a staging Heroku app created with suspenders with `RAILS_ENV` defined as `production`. This might be a bug in suspenders.

tute:
> We just saw a staging Heroku app created with suspenders with RAILS_ENV defined as production. This might be a bug in suspenders.

I tested for this, it works well out of the box.

PR #717, opened 2016-02-23
Add missing dependencies to gemspec

### LINKED ISSUE #715: Streamline development setup
I was trying to setup suspenders, to help out with #676, but ran into a few issues.

I ran `bin/setup`, then I had to manually install the below gems with `gem install`:

* `simple_form`
* `title`
* `quiet_assets`
* `capybara-webkit`
  * Needed to install `brew install qt`

not that it was a huge issue, but thought I would mention it since its wasn't documented anywhere. 

__Possible improvements__:

* Document in `README`
* Maybe an idea would be to just print it in `bin/setup` and mention the dependency on `qt`?
* Lastly those gems could be installed in `bin/setup, however:
  - `capybara-webkit`s dependency on `qt` is a problem (easily installed with `brew` in my case, but what about other systems? Detect the system an install accordingly seems perhaps a bit excessive?)
  - Installing global gems in`bin/setup` isn't perhaps a good idea

I'd be happy to do a PR for any of the above.

Thanks!

buren:
* simple_form
* title
* quiet_assets
* capybara-webkit

Closes #715

### DIFF
--- a/suspenders.gemspec
+++ b/suspenders.gemspec
@@ -32,4 +32,8 @@ rush to build something amazing; don't use it if you like missing deadlines.
   s.add_dependency 'rails', Suspenders::RAILS_VERSION
 
   s.add_development_dependency 'rspec', '~> 3.2'
+  s.add_development_dependency 'simple_form', '~> 3.2'
+  s.add_development_dependency 'title', '~> 0.0'
+  s.add_development_dependency 'quiet_assets', '~> 1.1'
+  s.add_development_dependency 'capybara-webkit', '~> 1.8'
 end
--- a/suspenders.gemspec
+++ b/suspenders.gemspec
@@ -27,11 +27,11 @@ rush to build something amazing; don't use it if you like missing deadlines.
   s.test_files = `git ls-files -- {test,spec,features}/*`.split("\n")
   s.version = Suspenders::VERSION
 
-  s.add_dependency 'bitters', '~> 1.2'
   s.add_dependency 'bundler', '~> 1.3'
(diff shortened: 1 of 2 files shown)


suspenders.gemspec
```diff
@@ -30,6 +30,10 @@ rush to build something amazing; don't use it if you like missing deadlines.
   s.add_dependency 'bitters', '~> 1.2'
   s.add_dependency 'bundler', '~> 1.3'
   s.add_dependency 'rails', Suspenders::RAILS_VERSION
+  s.add_dependency 'simple_form', '~> 3.2'
+  s.add_dependency 'title', '~> 0.0'
+  s.add_dependency 'quiet_assets', '~> 1.1'
+  s.add_dependency 'capybara-webkit', '~> 1.8'
```
tute: I think these (and bitters) should be `development_dependency`s (used for running tests). Generated app's Gemfile will take care of those while using suspenders.

buren: Thanks @tute, mistake on my part :)

buren:
Didn't get any errors about `bitters` not being installed. @tute should I add that as a development dependency too?

tute:
> Didn't get any errors about bitters not being installed. @tute should I add that as a development dependency too?

That's surprising. Maybe you already had it installed? I think it is a development dependency, yes. I'll test this cleaning up my gems when you are ready. Thank you!

buren:
@tute totally missed the fact that `bitters` :tired_face: already was in the gemfile.. moved it from a runtime dependency to a development dependency in 347af87.

tute:
Thank you! Rebased and merged in! :) :tada:

PR #724, opened 2016-03-01
Update browser support

tysongach:
- According to http://caniuse.com/usage-table, global usage of IE 9 and 10 are
both around 0.6%.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 * Remove `RAILS_ENV` definitions
 * Update Bourbon to v5.0.0.beta.3
 * Update Bitters to v1.3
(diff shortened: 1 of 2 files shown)


tute:
We can drop 10 already? That is great! Can you add links with information to the commit message, please? Thank you!

brendastorer:
:tada: :tada: :tada:

tysongach:
@tute Updated, and provided better messaging.

I also renamed this PR and added another commit to drop iOS 7, which has little usage these days, as well.

cc: @brendastorer

tute:
:+1: Thank you!

PR #725, opened 2016-03-04
Set development delivery_method to :file

tute:
`:test` sends the emails to the logs, and to see them one has to scroll
back through them. If we configure delivery_method to be `:file`,
mails will also be created in the `./tmp/mails/` directory

One file will be created for each email address sent to that contains
all email content delivered to that address. This allows to easily
review any and all emails sent, without searching through logs.

See:
* https://github.com/thoughtbot/suspenders/pull/521 (https://github.com/thoughtbot/suspenders/commit/8b45fd42)
* http://blog.mojotech.com/a-decade-of-rails/

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -51,7 +51,7 @@ module Suspenders
     def set_test_delivery_method
       inject_into_file(
         "config/environments/development.rb",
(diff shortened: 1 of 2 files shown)


bernerdschaefer:
Nice. :+1:

tute:
Thanks for your review!

PR #729, opened 2016-03-13
Updates version to 1.37.0

lorentrogers:
Updates News.md

[ci skip]

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,15 @@
+1.37.0 (March 13, 2016)
+
 * Remove `RAILS_ENV` definitions
+* Set development `action_mailer.delivery_method` to `:file`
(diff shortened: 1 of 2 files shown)


tute:
Thank you, @lorentrogers! This looks great.

I'd add a comment for "Set development delivery_method to :file" with a bit more context, to let developers know that they can see their outbound emails from Rails in `tmp/mails`. They might find it useful.

With that, we'll merge and release. Thank you!

tute:
Something like:

```
* Set development `action_mailer.delivery_method` to `:file`
  so that mails are copied to `./tmp/mails/` directory for easy development
  access.
```

lorentrogers:
Updated!

tute:
Thank you very much!

Issue #731, opened 2016-03-14
Bullet error during run


lorentrogers:
I'm getting an error while creating an app with `bundle exec bin/suspenders /tmp/testjunk`

    ...
    generate  rspec:install
    /tmp/testjunk2/config/environments/development.rb:19:in `block (2 levels) in <top (required)>': uninitialized constant Bullet (NameError)
        from /home/lorentrogers/.rvm/gems/ruby-2.3.0/gems/activesupport-4.2.6/lib/active_support/lazy_load_hooks.rb:36:in `execute_hook'
        from /home/lorentrogers/.rvm/gems/ruby-2.3.0/gems/activesupport-4.2.6/lib/active_support/lazy_load_hooks.rb:45:in `block in run_load_hooks'
        from /home/lorentrogers/.rvm/gems/ruby-2.3.0/gems/activesupport-4.2.6/lib/active_support/lazy_load_hooks.rb:44:in `each'
        from /home/lorentrogers/.rvm/gems/ruby-2.3.0/gems/activesupport-4.2.6/lib/active_support/lazy_load_hooks.rb:44:in `run_load_hooks'
        from /home/lorentrogers/.rvm/gems/ruby-2.3.0/gems/railties-4.2.6/lib/rails/application/finisher.rb:62:in `block in <module:Finisher>'
        from /home/lorentrogers/.rvm/gems/ruby-2.3.0/gems/railties-4.2.6/lib/rails/initializable.rb:30:in `instance_exec'
        from /home/lorentrogers/.rvm/gems/ruby-2.3.0/gems/railties-4.2.6/lib/rails/initializable.rb:30:in `run'
        from /home/lorentrogers/.rvm/gems/ruby-2.3.0/gems/railties-4.2.6/lib/rails/initializable.rb:55:in `block in run_initializers'
        from /home/lorentrogers/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/tsort.rb:228:in `block in tsort_each'
        from /home/lorentrogers/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/tsort.rb:350:in `block (2 levels) in each_strongly_connected_component'
        from /home/lorentrogers/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/tsort.rb:431:in `each_strongly_connected_component_from'
[...]

PR #732, opened 2016-03-14
Adds missing gems for development

lorentrogers:
Adds the `pg` and `bullet` gems into the gemspec. Seems to fix the errors found in #731

### DIFF
--- a/suspenders.gemspec
+++ b/suspenders.gemspec
@@ -36,4 +36,6 @@ rush to build something amazing; don't use it if you like missing deadlines.
   s.add_development_dependency 'title', '~> 0.0'
   s.add_development_dependency 'quiet_assets', '~> 1.1'
   s.add_development_dependency 'capybara-webkit', '~> 1.8'
+  s.add_development_dependency 'bullet', '~> 5.0'
+  s.add_development_dependency 'pg'
 end


tute:
We also did some at https://github.com/thoughtbot/suspenders/pull/717. I cleaned up all my gems and I found we needed execjs, and then more. This is all weird, because tests run `suspenders`, and that runs `bundle install`. All these dependencies are needed for the generated app, not suspenders itself, so I now don't think we should keep adding dependencies to the gemspec.

I just cleaned up again all my gems, ran `suspenders new-app`, and then the tests, and I didn't see any errors. I think we need to setup a new application before running any tests, but I'm not sure yet how to do that. This PR, if it achieves that, should actually remove dependencies.

What do you think of this? Thanks for your contribution!

tute:
Hey I think I found it! Tests start with:

```ruby
drop_dummy_database
remove_project_directory
run_suspenders
```

`drop_dummy_database` goes into the dummy rails app for testing and runs `rake db:drop`, but rake won't find the gems if they are not yet there! `bundle happens later, on `run_suspenders`.

So this patch makes it work well:

```diff
diff --git a/spec/support/suspenders.rb b/spec/support/suspenders.rb
index a585a17..4f371f0 100644
--- a/spec/support/suspenders.rb
+++ b/spec/support/suspenders.rb
@@ -24,6 +24,7 @@ module SuspendersTestHelpers
     if File.exist?(project_path)
       Dir.chdir(project_path) do
         Bundler.with_clean_env do
+          `bundle check || bundle install`
           `rake db:drop`
         end
       end
```

Better would be to define:

```ruby
  def setup_app_dependencies
    if File.exist?(project_path)
      Dir.chdir(project_path) do
        Bundler.with_clean_env do
          `bundle check || bundle install`
        end
      end
    end
  end
```

I think we can remove many dependencies:

```diff
-  s.add_development_dependency 'bitters', '~> 1.3'
-  s.add_development_dependency 'simple_form', '~> 3.2'
-  s.add_development_dependency 'title', '~> 0.0'
-  s.add_development_dependency 'quiet_assets', '~> 1.1'
-  s.add_development_dependency 'capybara-webkit', '~> 1.8'
-  s.add_development_dependency 'bullet', '~> 5.0'
-  s.add_development_dependency 'pg'
```

And call it where appropriate. What do you think of this?

Thanks for your work!

tute:
If you use `rbenv`, this is the script I used for cleaning up my env:

```bash
#!/usr/bin/env bash

uninstall() {
  list=`gem list --no-versions`
  for gem in $list; do
    gem uninstall $gem -aIx
  done
  gem list
  gem install bundler
}

#rbenv versions --bare
RBENVPATH=`rbenv root`
echo $RBENVPATH
RUBIES=`ls $RBENVPATH/versions`
for ruby in $RUBIES; do
  echo '---------------------------------------'
  echo $ruby
  rbenv local $ruby
  uninstall
done
```

PR #733, opened 2016-03-15
Replaces static error pages with controller

### LINKED ISSUE #673: Stylize 404 and 500 error pages
Rails should show by default 404 and 500 pages with the application's layout, instead of the normal Rails-default bare html pages.

Sample (a bit pseudo) code from `app/controllers/application_controller.rb` from a project I have worked on in the past:

``` ruby
  if Rails.env.production?
    rescue_from ActionController::RoutingError, with: :route_not_found
    rescue_from ActiveRecord::RecordNotFound, with: :route_not_found
    # Are there others?

    rescue_from NoMethodError, with: :internal_server_error
  end

  def internal_server_error
    notify_exception
    render "errors/500.html", status: :internal_server_error
  end

  def route_not_found
    notify_exception
    render "errors/404.html", status: :not_found
  end
```

lorentrogers:
This adds basic generators for error pages that use
controllers and views instead of static html pages
with inline css, etc.

Closes #673

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -443,17 +443,53 @@ you can deploy to staging and production with:
       copy_file "json_encoding.rb", "config/initializers/json_encoding.rb"
     end
 
-    def customize_error_pages
-      meta_tags =<<-EOS
-  <meta charset="utf-8" />
-  <meta name="ROBOTS" content="NOODP" />
-  <meta name="viewport" content="initial-scale=1" />
-      EOS
+    def inject_error_handling
+      page_methods = <<-EOS
 
-      %w(500 404 422).each do |page|
-        inject_into_file "public/#{page}.html", meta_tags, after: "<head>\n"
-        replace_in_file "public/#{page}.html", /<!--.+-->\n/, ''
-      end
+  if Rails.env.production?
+    rescue_from ActionController::RoutingError, with: :route_not_found
+    rescue_from ActiveRecord::RecordNotFound, with: :route_not_found
+
+    rescue_from NoMethodError, with: :internal_server_error
+  end
+
+  def internal_server_error
+    notify_exception
+    render "errors/internal_server_error", status: :internal_server_error
+  end
+
+  def route_not_found
+    notify_exception
+    render "errors/not_found", status: :not_found
+  end
+      EOS
+      inject_into_file "app/controllers/application_controller.rb",
+                       page_methods,
+                       after: "protect_from_forgery with: :exception\n"
+    end
+
+    def copy_error_controller
+      copy_file "errors_controller.rb", "app/controllers/errors_controller.rb"
+    end
+
+    def inject_error_routes
+      error_routes = <<-EOS
+  match "/404", :to => "errors#not_found", :via => :all
+  match "/500", :to => "errors#internal_server_error", :via => :all
+       EOS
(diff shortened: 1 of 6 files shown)


lorentrogers:
I couldn't quite tell how to test that the error pages are properly caught in production... I assume this should be done by hand on a built project?

lorentrogers:
Also, this post noted an additional configuration that I didn't include - may be useful https://mattbrictson.com/dynamic-rails-error-pages/#4-tell-rails-to-use-our-routes-for-error-handling

tute:
> I couldn't quite tell how to test that the error pages are properly caught in production... I assume this should be done by hand on a built project?

What if instead of `if Rails.env.production?` we run those rescues only `if !Rails.env.development?`? Then we'd be able to assert on test environment, but we'd lose valuable stacktraces when the errors happen in the test environment. This is tricky indeed, not sure what's best yet.

---

I think we'll need to implement `notify_exception`, which will be something like `Honeybadger.notify(exception)`.

---

If we `render` right from `ApplicationController`, we don't need `ErrorsController` I think. Right?

---

I'm now testing your branch and I don't see the layout, checking why. Thanks for your work!

lib/suspenders/app_builder.rb
```diff
@@ -443,17 +443,53 @@ def copy_miscellaneous_files
...
-        replace_in_file "public/#{page}.html", /<!--.+-->\n/, ''
-      end
+    def copy_error_controller
+      copy_file "errors_controller.rb", "app/controllers/errors_controller.rb"
+    end
+
+    def inject_error_routes
+      error_routes = <<-EOS
+  match "/404", :to => "errors#not_found", :via => :all
+  match "/500", :to => "errors#internal_server_error", :via => :all
```
tute: Do we need the `ErrorsController` and routes, if we only `render` in `ApplicationController`? We might not.

tute:
Comments from @mcmire in #673:

> Instead of using Rails.env.production?, what do you think about using the Rails configuration option that controls whether exception detail pages are shown (they're usually shown in development)? I think what you might want to say is, if exception detail pages are NOT being shown, then render a 404 or 500 page. A consequence of this is that 404/500 pages would then be rendered in test mode -- but I think that might be what you want, to simulate what would happen in production mode.

lib/suspenders/app_builder.rb
tute: If we do need the routes, I'd do `get "/#{code}"`, and not `via: :all`.

jferris:
Have we tried this out in a number of our applications? In general, Suspenders reflects what we've tried and decided to repeat rather than things which seem like potential best practices for the future.

This feels like too much logic for Suspenders. If we want to change how errors are handled in Rails applications, I'd suggest creating a new gem that provides any generators and shared logic necessary. If we think we want to do that in Suspenders, we could depend on that gem.

If the goal is to make sure error pages are correctly styled, I'd prefer to look for a way to generate static HTML pages during asset compilation. The error pages are used when something has already gone wrong, so we don't know what we can depend on when rendering HTML. The default approach of rendering static pages seems wise to me.

tute:
I agree this is too much code. Would love a solution that hooks into the asset pipeline to generate the appropriate HTML file.

lorentrogers:
@jferris This makes a lot of sense to me. I'm going to close this one out for now.

PR #734, opened 2016-03-18
Adds handling for ARGV.empty?

### LINKED ISSUE #730: `suspenders help` should provide actual help ...not generate an app...
Just made this mistake, expecting help. If the app is actually called "help", there should probably be a `--name` option or something.

lorentrogers:
Instead of printing default output, provide a more helpful
prompt with more info on what went wrong and
directions on how to get help.

Closes #730

### DIFF
--- a/bin/suspenders
+++ b/bin/suspenders
@@ -6,7 +6,12 @@ $LOAD_PATH << source_path
 
 require 'suspenders'
 
-if ['-v', '--version'].include? ARGV[0]
+if ARGV.empty?
+  puts "Please provide a path for the new application"
+  puts
+  puts "See --help for more info"
+  exit 0
+elsif ['-v', '--version'].include? ARGV[0]
   puts Suspenders::VERSION
   exit 0
 end


bin/suspenders
```diff
@@ -6,7 +6,12 @@ $LOAD_PATH << source_path
 
 require 'suspenders'
 
-if ['-v', '--version'].include? ARGV[0]
+if ARGV.empty?
+  puts "Please provide a path for the new application"
```
tute: Should it be name instead of path?

tute:
I like this. One small question. Thank you, @lorentrogers!

lorentrogers:
I considered that, but the argument is actually a path, right? I've been building from my `~/dev/suspenders` directory to `/tmp/fakeapp1`. The path includes the name of the destination app at the end, right? (Or would this make an app literally called "/tmp/fakeapp1"? In which case, it should make it in the current directory, and be documented.)

tute:
I always treated it as a name, but it's a path. TIL! :)

tute:
Thank you!

bin/suspenders
```diff
@@ -6,7 +6,12 @@ $LOAD_PATH << source_path
...
+  puts
+  puts "See --help for more info"
```
tute: It's passing help to Rails, not suspenders. This smells like a bug:
```
suspenders master % suspenders --help
Usage:
  rails new APP_PATH [options]
```

PR #736, opened 2016-03-22
Update CSS override

tysongach:
- Remove webkit prefix
- This prefix hasn't been needed since Safari 6 and iOS Safari 6.1
- Reference: http://caniuse.com/#feat=css-transitions


tute:
LGTM :+1:

PR #739, opened 2016-03-24
Make Bitters a runtime dependency

### LINKED ISSUE #738: Out-of-the-box SCSS misconfiguration
After scaffolding out an app from a fresh `gem install suspenders` -- without any modifications -- my first feature spec is failing to run with the following error:

```
     Failure/Error: $base-font-family: $helvetica;
     
     ActionView::Template::Error:
       Undefined variable: "$helvetica".
     # ./app/assets/stylesheets/base/_variables.scss:2
     # ./app/assets/stylesheets/base/_base.scss:6
     # ./app/assets/stylesheets/application.scss:8
```

tysongach:
- Bitters was made a development dependency rather than a runtime dependency in
  3b7ce010560c965010b9a72f55979720e175362c, likely by accident.
- The reason we need Bitters to be a runtime dependency is so we can force
  installation of a specific version of Bitters, rather than leaning on a
  version the person might have installed already.
- Fixes #738

### DIFF
--- a/suspenders.gemspec
+++ b/suspenders.gemspec
@@ -27,11 +27,11 @@ rush to build something amazing; don't use it if you like missing deadlines.
   s.test_files = `git ls-files -- {test,spec,features}/*`.split("\n")
   s.version = Suspenders::VERSION
 
+  s.add_dependency 'bitters', '~> 1.3'
   s.add_dependency 'bundler', '~> 1.3'
   s.add_dependency 'rails', Suspenders::RAILS_VERSION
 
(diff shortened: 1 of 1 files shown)


tute:
`bitters` is both a runtime dependency and a development dependency (we generate apps while testing). I think making it a runtime dependency, and then running `bundle` while testing would fix these issues (see https://github.com/thoughtbot/suspenders/pull/732#issuecomment-198181368).

I'll work tomorrow on this, and try to fix the dependencies issues we've been having in these two PRs.

PR #741, opened 2016-03-30
Return heroku run exit code status

jsom:
I noticed that migrations were failing on my deploys but heroku was still registering it as a successful deploy. It turns out that `heroku run` does not return the status code from the command you want to run.

Adding `--exit-code` to the `heroku run` commands will correctly return the status code to the invoker.

Documentation: https://github.com/heroku/heroku-run

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+1.38.0
+
+* Successful deploys with failing DB migrations on Heroku bugfix
+
 1.37.0 (March 13, 2016)
 
 * Remove `RAILS_ENV` definitions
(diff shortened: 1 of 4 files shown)


tute:
This is great, thank you! Can you please quote or add a link to Heroku's documentation for this? Then, merge at will.

tute:
Oh, also, a note in the NEWS file is :+1:'d. :)

jsom:
Updated with documentation and NEWS.

NEWS.md
```diff
@@ -1,3 +1,7 @@
+1.38.0
+
+* Fix `heroku run` exit code not returning on migrations
```
tute: Would something like "Successful deploys with failing DB migrations on Heroku bugfix" explain this better?

tute:
Left one question on the NEWS comment, looks good to me otherwise. Merge when you like!

PR #742, opened 2016-04-01
Default `EMAIL_RECIPIENTS` to be empty

seanpdoyle:
Some applications might not need to configure email interception out of
the box.

Without additional configuration, a newly `suspender`'d application
won't properly run in a `RACK_ENV=staging` environment.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 master
 
+* Default an unspecified `EMAIL_RECIPIENTS` to be the empty string
 * Drops staging environment in favor of configuration through env variables
 
 1.38.0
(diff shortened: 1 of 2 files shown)


tute:
:+1:

tute:
This code was moved to `smtp.rb`.

seanpdoyle:
@tute I've updated this PR.

tute:
Now that I review it, because of the `if ENV["EMAIL_RECIPIENTS"].present?` condition, this piece of code won't fail any longer. I think this PR can be closed now. Thank you, Sean!

PR #743, opened 2016-04-10
Drop staging environment file

### LINKED ISSUE #506: No separate Rails staging environment
Use `production` config on Heroku staging:

https://devcenter.heroku.com/articles/deploying-to-a-custom-rails-environment

tute:
The only custom configuration of suspended apps for the staging
environment was the email interceptor, which can live in `production.rb`
and be configured through environmental variables.

* Simplifies configuration
* Enforces staging/production parity
* Follows Heroku recommendations (https://devcenter.heroku.com/articles/multiple-environments)

[fixes #506]

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+master
+
+* Drops staging environment in favor of configuration through env variables
+
 1.38.0
 
 * Successful deploys with failing DB migrations on Heroku bugfix
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -25,11 +25,9 @@ module Suspenders
       end
 
       def create_staging_heroku_app(flags)
-        rack_env = "RACK_ENV=staging"
         app_name = heroku_app_name_for("staging")
 
         run_toolbelt_command "create #{app_name} #{flags}", "staging"
-        run_toolbelt_command "config:add #{rack_env}", "staging"
       end
 
       def create_production_heroku_app(flags)
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -196,20 +196,6 @@ module Suspenders
       )
     end
 
-    def setup_staging_environment
-      staging_file = 'config/environments/staging.rb'
-      copy_file 'staging.rb', staging_file
-
-      config = <<-RUBY
-
-Rails.application.configure do
-  # ...
-end
-      RUBY
-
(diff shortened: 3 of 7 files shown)


tute:
cc @jferris for review. Thank you!

jferris:
This looks good to me!

tute:
Thanks for your review Joe!

blackjid:
With this PR now staging and production apps runs with the `production` environment. How do you deal with services that aggregate data based on the environment?

For example in honeybadger there is `env` option that by default use the `Rails.env` value. 

Without this PR, that would work out of the box, but always using `production` honeybadger would assume that is the same app as the share the same api token.

Probably there is something i'm missing. 
Thanks

tute:
You can set the hostname or env via ENV variables in Honeybadger's configuration file: http://docs.honeybadger.io/lib/ruby.html. `HONEYBADGER_ENV` is an option.

blackjid:
In the current state of suspenders one needs to know this, and manually configure it, to prevent mixing up staging and production logging on honeybadger?? 

What do you think about adding this ENV when heroku apps are created?? with staging and production values for each application?

tute:
I think it does make sense to set this up automatically. Thanks for
bringing this up.

tute:
Update: Skylight doesn't seem to differentiate in the interface between environments. They need to be treated as separate apps, and thus, it doesn't make sense to add new behavior to suspenders yet.

tute:
Whoops, wrong issue, deleting

blackjid:
no problem

tute:
Created a new issue to track the Honeybadger's environment: https://github.com/thoughtbot/suspenders/issues/759

ivobenedito:
Hey guys, am I missing something here or this update requires more changes on other suspenders Rails file templates?

Example:
- [suspenders/templates/Gemfile.erb](https://github.com/thoughtbot/suspenders/blob/master/templates/Gemfile.erb)
- [suspenders/templates/postgresql_database.yml.erb](https://github.com/thoughtbot/suspenders/blob/a930e6bf9fbc4fb11bf59fbf637349a0e3733fad/templates/postgresql_database.yml.erb)
- ...

The **Gemfile.erb** file has many gems scoped to :staging env.
The **database.yml** file has :staging env defined.
Etc ...

tute:
That is right! Also useless in `templates/secrets.yml`. Would you like to send a PR cleaning that up? Thanks!

PR #745, opened 2016-04-15
Fix suspenders development and runtime dependencies

tute:
Suspenders' tests generate an application to run assertions on it. This
means its runtime dependencies are themselves development dependencies.

Those dependencies were not being installed by suspenders' tests,
neither specified in its `.gemspec` as development dependencies, but
they were needed for new installation tests to run.

We address this issue by adding a `setup_app_dependencies` step in the
specs that need it, after generating an application (`run_suspenders`).
This ensures all dependencies are installed before we need them.

This makes it unnecessary to declare app dependencies in
`suspenders.gemspec`, because Rails and Bundler (only declared runtime
dependencies) will take care of those.

[fixes #732]
[fixes #739]

Pairing: @tute and @robskrob.

### DIFF
--- a/spec/features/github_spec.rb
+++ b/spec/features/github_spec.rb
@@ -9,6 +9,7 @@ RSpec.describe "GitHub" do
   it "suspends a project with --github option" do
     repo_name = 'test'
     run_suspenders("--github=#{repo_name}")
+    setup_app_dependencies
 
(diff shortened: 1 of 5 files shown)


derekprior:
Makes sense to me. I'm not super familiar with how the tests run, but this looks sane.

Issue #746, opened 2016-04-19
Instructions for non-optional configuration


klaaspieter:
I just deployed my first suspenders app and was first greeted by an SMTP configuration error. After removing the SMTP config, I received error: `KeyError: key not found: "APPLICATION_HOST"`. 

In #532 @tute mentioned suspenders is not meant to be plug and play. Does it make sense to add something to the README or after suspenders finishes to explain what still needs configuring?

tute:
We've been documenting through exceptions until now, which feels very unnatural to me. In https://github.com/thoughtbot/suspenders/pull/722#issuecomment-210556755 I state that I wish for a suspended app not to fail upon deploy to Heroku, but I've had some pushback in moving towards that direction.

Specifically for this case, your PR fixes the `APPLICATION_HOST` issue (thank you!). `SMTP` could be conditional or not, but if we don't make it so, I vote for having what's needed in the README.

mike-burns:
Uses for `APPLICATION_HOST`:

- Used in `Rack::CanonicalHost`. This way, requests to other hostnames (such as the Heroku hostname) will redirect. This helps with TLS certs, SEO, and other problems.
- Used as the default `config.action_controller.asset_host`, though can be overridden by setting `ASSET_HOST`. This value is used in e.g. `image_url`, `stylesheet_link_tag`, etc. This ensures that it makes the request using the canonical hostname, which helps with TLS and browser security checks.
- Used as the production `Action::Mailer` host. This way emails will appear to come from the correct domain, which helps ensure reliable delivery.

^ This might all be useful for writing documentation on this.

tute:
`APPLICATION_HOST` is now fixed, `SMTP` variables are next (failing tests in https://github.com/thoughtbot/suspenders/pull/722 are targeting that).

I'm not sure what variables will fail next, but I do plan to move #722 forward, which leads to fixing this issue.

If in the end we get green CI in different environments, but developer input is still needed (for example, SMTP credentials), then I'll add post-generate instructions so that developers know what to expect.

Thank you all for your input.

PR #747, opened 2016-04-19
Set the default application host

klaaspieter:
This sets the default application host to "[app-name]-[environment].herokuapp.com" to prevent the error I talked about in #746. 

Does this make sense? This configures the app to heroku defaults to make suspenders more plug & play, but still doesn't explain how to change this after acquiring a custom domain. Is a change even required in that case?

### DIFF
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -65,7 +65,6 @@ module Suspenders
           exit 1
         end
 
-        heroku_app_name = app_builder.app_name.dasherize
         run_toolbelt_command(
           "pipelines:create #{heroku_app_name} \
             -a #{heroku_app_name}-staging --stage staging",
@@ -88,6 +87,15 @@ module Suspenders
         end
       end
 
+      def set_heroku_application_host
+        %w(staging production).each do |environment|
+          run_toolbelt_command(
+            "config:add APPLICATION_HOST=#{heroku_app_name}-#{environment}.herokuapp.com",
+            environment,
+          )
+        end
+      end
+
       private
 
       attr_reader :app_builder
@@ -105,8 +113,12 @@ fi
         SHELL
       end
 
+      def heroku_app_name
+        app_builder.app_name.dasherize
+      end
+
       def heroku_app_name_for(environment)
-        "#{app_builder.app_name.dasherize}-#{environment}"
+        "#{heroku_app_name}-#{environment}"
       end
 
       def generate_secret
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -13,6 +13,7 @@ module Suspenders
                    :provide_review_apps_setup_script,
                    :set_heroku_rails_secrets,
                    :set_heroku_remotes,
+                   :set_heroku_application_host,
                    :set_heroku_serve_static_files,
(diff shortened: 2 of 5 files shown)


lib/suspenders/adapters/heroku.rb
```diff
@@ -88,6 +88,16 @@ def set_heroku_serve_static_files
         end
       end
 
+      def set_heroku_application_host
+        heroku_app_name = app_builder.app_name.dasherize
```
tute: `heroku_app_name = app_builder.app_name.dasherize` is repeated in line 68. Can we extract a method from both?

lib/suspenders/app_builder.rb
```diff
@@ -13,8 +13,10 @@ class AppBuilder < Rails::AppBuilder
                    :provide_review_apps_setup_script,
                    :set_heroku_rails_secrets,
                    :set_heroku_remotes,
+                   :set_heroku_application_host,
                    :set_heroku_serve_static_files,
-                   :set_up_heroku_specific_gems
+                   :set_up_heroku_specific_gems,
+                   :set_heroku_application_host
```
tute: Is there a reason why the alphabetical order was changed?

tute:
Bugfix and code look good to me. Left two small comments, after that, feel free to squash and merge. Thanks for this! :)

mike-burns:
> Does this make sense?

I'm only like 15% opposed to this patch.

> This configures the app to heroku defaults to make suspenders more plug & play, but still doesn't explain how to change this after acquiring a custom domain. Is a change even required in that case?

A change is required in that case. My worry is that this autonomy will lull devs into a false sense of security. That's my 15% of opposition.

I would be 0% opposed to a documentation patch, regardless of whether this goes in.

klaaspieter:
@mike-burns I would include the README change in this PR, but honestly I don't know the purpose of the `APPLICATION_HOST` variable. I feel any good documentation would explain that at a minimum. 

Perhaps someone can the explanation as part of #746?

lib/suspenders/app_builder.rb
klaaspieter: Because I didn't realize they were alphabetical 🎱

lib/suspenders/adapters/heroku.rb
```diff
@@ -105,8 +114,12 @@ def command_to_join_heroku_app(environment)
         SHELL
       end
 
+      def heroku_app_name
+        app_builder.app_name.dasherize
+      end
```
tute: We can now drop `heroku_app_name = app_builder.app_name.dasherize` at line 91.

mike-burns:
@klaaspieter documentation added: https://github.com/thoughtbot/suspenders/issues/746#issuecomment-212930477

lib/suspenders/app_builder.rb
```diff
@@ -13,8 +13,9 @@ class AppBuilder < Rails::AppBuilder
                    :provide_review_apps_setup_script,
                    :set_heroku_rails_secrets,
                    :set_heroku_remotes,
+                   :set_heroku_application_host,
                    :set_heroku_serve_static_files,
-                   :set_up_heroku_specific_gems
+                   :set_up_heroku_specific_gems,
```
klaaspieter: Oops. This `,` causes readme to be added as a delegator.

PR #752, opened 2016-04-28
Drop initializer for disabling XML parser

tute:
Vulnerabillity was fixed three years ago
(https://www.kb.cert.org/vuls/id/380039).

Also, this code doesn't work in Rails 5.


mike-burns:
Heh, I was wondering about that.

+1 kill it.

tute:
Thanks for your review.

PR #753, opened 2016-04-28
Upgrade to Rails 5.0.0

tute:
Upgrade to Rails 5.0.0

* Replaces default Puma setup configuration that comes with Rails 5.
  We were already using it as per Heroku preference.
* No need of static files custom logic
  See https://blog.heroku.com/archives/2016/5/2/container_ready_rails_5

I needed to add the `listen` gem in development. See related PR to
Rails and its related issues: https://github.com/rails/rails/pull/24243

### DIFF
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -23,7 +23,7 @@ gem "title"
 gem "uglifier"
 
 group :development do
-  gem "quiet_assets"
+  gem "listen"
   gem "spring"
   gem "spring-commands-rspec"
   gem "web-console"
@@ -37,7 +37,8 @@ group :development, :test do
   gem "factory_girl_rails"
   gem "pry-byebug"
   gem "pry-rails"
-  gem "rspec-rails", "~> 3.4.0"
+  gem "refills"
+  gem "rspec-rails", "~> 3.5.0.beta4"
 end
 
 group :development, :staging do
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 master
 
+* Update to Rails 5
+* Drop quiet_assets
 * Drop unneeded `suspenders` aliases: `--skip-test-unit`, `--skip-turbolinks`,
   `--skip-bundle`. Drops `-G` that clashes with Rails’ `--skip-git` alias.
 
--- a/README.md
+++ b/README.md
@@ -74,8 +74,6 @@ And development gems like:
   pre-loading
 * [Web Console](https://github.com/rails/web-console) for better debugging via
   in-browser IRB consoles.
-* [Quiet Assets](https://github.com/evrone/quiet_assets) for muting assets
-  pipeline log messages
 
 And testing gems like:
 
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -70,15 +70,6 @@ module Suspenders
         )
       end
 
-      def set_heroku_serve_static_files
-        %w(staging production).each do |environment|
-          run_toolbelt_command(
-            "config:add RAILS_SERVE_STATIC_FILES=true",
-            environment,
-          )
-        end
-      end
-
       def set_heroku_application_host
         %w(staging production).each do |environment|
           run_toolbelt_command(
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -13,8 +13,7 @@ module Suspenders
                    :create_review_apps_setup_script,
                    :set_heroku_rails_secrets,
                    :set_heroku_remotes,
-                   :set_heroku_application_host,
-                   :set_heroku_serve_static_files
+                   :set_heroku_application_host
 
     def readme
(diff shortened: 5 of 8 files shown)


johnpaulashenfelter:
Also now beta4 is out @tute

tute:
Thanks JP! Also, we might be able to simplify the generated code: https://blog.heroku.com/archives/2016/5/2/container_ready_rails_5

tute:
cc @schneems for review too. Thank you!

templates/Procfile
```diff
@@ -1,2 +0,0 @@
-web: bundle exec puma -p $PORT -C ./config/puma.rb
-worker: bundle exec rake jobs:work
```
tute: Are these default tasks in Heroku?

schneems: Kinda. We use `rails server -p $PORT`. https://github.com/heroku/heroku-buildpack-ruby/blob/52b7238c748678fffc70c8f2a51e795305ef15c9/lib/language_pack/rails3.rb#L27

tute: I assume the practical difference is that we use Rails' default (https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/config/puma.rb) instead of our configuration file (https://github.com/thoughtbot/suspenders/blob/77312c9629535e3ccb279bc7316e77ace4ed0fb4/templates/puma.rb). @derekprior @calebthompson @jferris do we need our custom configuration on Rails 5?

jferris: @tute did you compare our config with the Rails default? There are some differences I saw right away:

* They default to a much higher number of threads
* They don't use multiple processes
* They use the `tmp_restart` plugin, which doesn't seem useful with Heroku/`heroku local`

Our total concurrency (2 threads * 2 workers) is similiar to theirs (5 threads * 1 worker), but I'm guessing we actually have more concurrency, as multiple Ruby threads will block each other on non-IO actions like rendering views.

schneems: Rails defaults needs to support jruby which doesn't support workers

---
Richard Schneeman
http://www.schneems.com




On Wed, May 4, 2016 at 7:06 AM -0700, "Joe Ferris" <notifications@github.com> wrote:












In templates/Procfile:
> @@ -1,2 +0,0 @@
> -web: bundle exec puma -p $PORT -C ./config/puma.rb
> -worker: bundle exec rake jobs:work



@tute did you compare our config with the Rails default? There are some differences I saw right away:


They default to a much higher number of threads
They don't use multiple processes
They use the tmp_restart plugin, which doesn't seem useful with Heroku/heroku local





Our total concurrency (2 threads * 2 workers) is similiar to theirs (5 threads * 1 worker), but I'm guessing we actually have more concurrency, as multiple Ruby threads will block each other on non-IO actions like rendering views.

tute:
Rolled back the `Procfile` and `puma.rb` changes, keeping our custom configuration for that. This is getting ready-ish, and I'd merge when Rails `5.0.0.its_real` is released.

tute:
Can we drop `quiet_assets` in favor of Rails 5 `config.assets.quiet = true`?

mike-burns:
Yes: https://github.com/evrone/quiet_assets#deprecation

templates/Gemfile.erb
```diff
@@ -37,7 +37,8 @@ group :development, :test do
   gem "factory_girl_rails"
   gem "pry-byebug"
   gem "pry-rails"
-  gem "rspec-rails", "~> 3.4.0"
+  gem "refills"
+  gem "rspec-rails", "~> 3.5.0.beta4"
```
tute: Should we merge even when `rspec-rails` is still in beta?  https://github.com/rspec/rspec-rails/releases

teoljungberg: It looks like there's going to be a release cut today - https://twitter.com/samphippen/status/748639259549114368

mike-burns: We would be losing these commits:

https://github.com/rspec/rspec-rails/commit/1fad32257f0970936f598b64de8ef4a40ede058d
https://github.com/rspec/rspec-rails/commit/5612e0d45ac4994fa72db41e611b4c6cbc2a33b6

It seems fine to do without those, and we can upgrade Suspenders next week after `rspec-rails` 3.5.0 is out.

mike-burns:
LGTM.

templates/Gemfile.erb
tute: > It seems fine to do without those, and we can upgrade Suspenders next week after rspec-rails 3.5.0 is out.

👍

tute:
Thanks for your reviews!

schneems:
🎉🎉🎉

tute:
> Suspenders, our Rails template, generates Rails 5 applications as of version 1.41, released today. That changeset removes a gem and logic that is not necessary anymore, because 5.0 is the easiest to deploy to Heroku.
> https://robots.thoughtbot.com/thoughtbot-ruby-open-source-is-ready-for-Rails-5

Thank you, @schneems!

Issue #735, opened 2016-03-21
Logs are missing (possible Puma+Quiet Assets interaction?)


sumeetjain:
Is it desired behavior for logs to not show for a Rails application generated by Suspenders? With a fresh application, running `heroku local` correctly serves the application, but no logs show in Terminal. But they do show in development.log, as one would expect.

I found https://github.com/evrone/quiet_assets/issues/40, which suggests a possible interaction between Puma and Quiet Assets. But when I run `rails s` (as opposed to `heroku local`) on a new Suspenders-generated application, the logs _do_ show–and Puma is still being used. Could there be some side effect of using forego?

I'm not even sure this is even a true issue. It's possible y'all prefer to only have logs in development.log, but I wanted to confirm before spending more time debugging this.

mike-burns:
We do want logs in production, and I have not seen problems like this on Heroku. Do you see them on Heroku, or only in `heroku local`?

tute:
> Do you see them on Heroku, or only in `heroku local`?

ping @sumeetjain for that question ^

sumeetjain:
At least on my computer, this is a problem locally. Here's the script I followed to test it:

```bash
➜  Code gem install suspenders
Fetching: bitters-1.3.1.gem (100%)
Successfully installed bitters-1.3.1
Fetching: suspenders-1.38.1.gem (100%)
Successfully installed suspenders-1.38.1
2 gems installed
➜  Code suspenders do_logs_show
[truncated]
➜  Code cd do_logs_show
➜  do_logs_show git:(master) ✗ rails g scaffold user name
Running via Spring preloader in process 60568
[truncated]
➜  do_logs_show git:(master) ✗ bundle exec rake db:migrate
== 20160428225518 CreateUsers: migrating ======================================
-- create_table(:users)
   -> 0.0073s
== 20160428225518 CreateUsers: migrated (0.0074s) =============================

➜  do_logs_show git:(master) ✗ heroku local
```

Then I navigated to **/users** and clicked through the application to make a new user. I expected some kind of output in the terminal, but nothing changed there at all. However, logs/development.log does reflect the correct actions. It just wasn't tailed to terminal.

mike-burns:
Can you try that again, but with the `rails_stdout_logging` gem? So after the `cd do_logs_show` step, add that gem to the `Gemfile`, bundle, and then see if that changes anything.

sumeetjain:
[...]

PR #756, opened 2016-05-02
Make new apps "deployable to Heroku" by default

sebarmano:
This change makes new suspender apps "deployable to Heroku" by default,
moving the Heroku setup in the local machine to be run by default (i.e.
with the --heroku flag off). All the setup that requires to hit the
Heroku API to create the staging and production apps is kept to be run
only with the flag on.

### DIFF
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -47,6 +47,7 @@ module Suspenders
       invoke :setup_dotfiles
       invoke :setup_git
       invoke :setup_database
+      invoke :create_local_heroku_setup
       invoke :create_heroku_apps
       invoke :create_github_repo
       invoke :setup_segment
@@ -57,11 +58,6 @@ module Suspenders
 
     def customize_gemfile
       build :set_ruby_to_version_being_used
-
-      if options[:heroku]
-        build :set_up_heroku_specific_gems
-      end
-
       bundle_command 'install'
       build :configure_simple_form
     end
@@ -165,18 +161,22 @@ module Suspenders
       end
     end
 
+    def create_local_heroku_setup
+      say "Creating local Heroku setup"
+      build :create_review_apps_setup_script
+      build :create_deploy_script
+      build :create_heroku_application_manifest_file
+    end
+
     def create_heroku_apps
       if options[:heroku]
         say "Creating Heroku apps"
         build :create_heroku_apps, options[:heroku_flags]
-        build :provide_review_apps_setup_script
         build :set_heroku_serve_static_files
         build :set_heroku_remotes
         build :set_heroku_rails_secrets
         build :set_heroku_application_host
-        build :create_heroku_pipelines_config_file
         build :create_heroku_pipeline
-        build :provide_deploy_script
         build :configure_automatic_deployment
       end
     end
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -223,6 +223,39 @@ RSpec.describe "Suspend a new project with default configuration" do
     expect(File).to exist("#{project_path}/spec/factories.rb")
   end
 
+  it "creates review apps setup script" do
+    bin_setup_path = "#{project_path}/bin/setup_review_app"
+    bin_setup = IO.read(bin_setup_path)
+
+    expect(bin_setup).to include("heroku run rake db:migrate --exit-code "\
+                                 "--app #{app_name.dasherize}-staging-pr-$1")
+    expect(bin_setup).to include("heroku ps:scale worker=1 "\
+                                 "--app #{app_name.dasherize}-staging-pr-$1")
+    expect(bin_setup).to include("heroku restart "\
+                                 "--app #{app_name.dasherize}-staging-pr-$1")
+    expect(File.stat(bin_setup_path)).to be_executable
+  end
+
+  it "creates deploy script" do
+    bin_deploy_path = "#{project_path}/bin/deploy"
+    bin_deploy = IO.read(bin_deploy_path)
+
+    expect(bin_deploy).to include("heroku run rake db:migrate --exit-code")
+    expect(File.stat(bin_deploy_path)).to be_executable
+  end
+
+  it "creates heroku application manifest file with application name in it" do
+    app_json_file = IO.read("#{project_path}/app.json")
+
+    expect(app_json_file).to match(/"name":"#{app_name.dasherize}"/)
+  end
+
+  it "sets up heroku specific gems" do
+    gemfile_file = IO.read("#{project_path}/Gemfile")
+
+    expect(gemfile_file).to include %{gem "rails_stdout_logging"}
+  end
+
   def app_name
     SuspendersTestHelpers::APP_NAME
   end
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 master
 
 * Update to Ruby 2.3.1
+* Make new apps "deployable to Heroku" by default.
 
 1.38.1 (April 20, 2016)
 
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -16,14 +16,6 @@ module Suspenders
         app_builder.append_file "bin/setup", remotes
       end
 
-      def set_up_heroku_specific_gems
-        app_builder.inject_into_file(
-          "Gemfile",
-          %{\n\s\sgem "rails_stdout_logging"},
-          after: /group :staging, :production do/,
-        )
-      end
-
       def create_staging_heroku_app(flags)
         app_name = heroku_app_name_for("staging")
 
@@ -45,7 +37,7 @@ module Suspenders
         end
       end
 
-      def provide_review_apps_setup_script
+      def create_review_apps_setup_script
         app_builder.template(
           "bin_setup_review_app.erb",
           "bin/setup_review_app",
@@ -54,7 +46,7 @@ module Suspenders
         app_builder.run "chmod a+x bin/setup_review_app"
       end
 
-      def create_heroku_pipelines_config_file
+      def create_heroku_application_manifest_file
         app_builder.template "app.json.erb", "app.json"
       end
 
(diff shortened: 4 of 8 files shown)


tute:
Thank you! Linking to related issues:

* https://github.com/thoughtbot/suspenders/issues/735
* https://github.com/thoughtbot/suspenders/pull/604

tute:
While we are editing, can we improve the changed lines to comply with the style guide, and please Hound CI? Thanks!

spec/features/new_project_spec.rb
```diff
@@ -223,6 +223,43 @@
...
+        staging:
+          branch: master
+          commands:
+            - bin/deploy staging
+    YML
+  end
+
+  it "adds app.json file" do
+    expect(File).to exist("#{project_path}/app.json")
+  end
```
tute: This spec is a subset of the next one (if this fails, the next one sure will). We can drop it.

spec/features/new_project_spec.rb
```diff
@@ -223,6 +223,43 @@
...
+    expect(bin_setup).to include("heroku ps:scale worker=1 --app #{app_name.dasherize}-staging-pr-$1")
+    expect(bin_setup).to include("heroku restart --app #{app_name.dasherize}-staging-pr-$1")
+    expect(File.stat(bin_setup_path)).to be_executable
+
+    bin_deploy_path = "#{project_path}/bin/deploy"
+    bin_deploy = IO.read(bin_deploy_path)
+
+    expect(bin_deploy).to include("heroku run rake db:migrate --exit-code")
+    expect(File.stat(bin_deploy_path)).to be_executable
+
```
tute: Let's add an `it` section:

> it "creates a circle.yml file with staging deployment commands" do

tute: > it "creates a circle.yml file with staging deployment commands" do

Now that I think of this, it will fail if there is no Heroku app set up. Circle CI configuration should only happen with `--heroku true` flag.

tute:
Left a few comments and questions. This is great work, thank you @sebarmano!

tute:
This is already a step forward. Question: do you think it's worth it to extract a shell script that performs the stateful operations?

* Create Heroku apps, pipelines, and remotes
* Configure CI autodeploys
* Set static files, rails secret, and application host constants

This would require some refactoring in suspenders. We could work on it as part of a different PR, too.

spec/features/new_project_spec.rb
sebarmano: Yes, I'll remove this test. Also, I think that the app.json test should be inside the "creates the local heroku setup" block, as it is part of it. What do you think?

tute: > Also, I think that the app.json test should be inside the "creates the local heroku setup" block, as it is part of it. What do you think?

That's true, but I like having smaller specs with more specific descriptions. I'd leave them separate (even though the project does not follow this rule in the different spec files).

tute:
Great, thanks! Let me know when you consider this ready for another review.

sebarmano:
> Let me know when you consider this ready for another review.

I think it's ready for review @tute. I fixed the Hound issues and deleted the redundant app.js test. Please, let me know if you think of anything else that can be improved regarding this PR.

tute:
> I think it's ready for review @tute. I fixed the Hound issues and deleted the redundant app.js test. Please, let me know if you think of anything else that can be improved regarding this PR.
>  Show all checks

What do you think of dividing the spec with more descriptive sections? https://github.com/thoughtbot/suspenders/pull/756#discussion_r61895648

tute:
Do you think we should solve https://github.com/thoughtbot/suspenders/pull/756#issuecomment-216561336 as part of another PR?

sebarmano:
> What do you think of dividing the spec with more descriptive sections?
I think it's a good idea. Right now we are "suspending" only one app to run all the tests in the block. Splitting the tests will add the time of creating a new app for each one of them (which I don't think is an issue at this point given the amount of tests). 
Also, right now there are three tests inside the block and one outside (the app.js one). I think splitting the tests will improve consistency in the feature spec file.

> Do you think we should solve #756 (comment) as part of another PR?
As I understand, the idea is to be able to have fully functional apps without the need of creating the staging and production Heroku counterparts. I think that if we keep the circle.yml file by default and someone intends to use Circle CI it just for their local app, it won't work (please correct me if I'm wrong on this). So I think the best call is to keep that file to be created only under when the --heroku flag is on. Also, I think that the change is part of this same PR as otherwise we would be merging something that is not 100% functional for some use cases.

spec/features/new_project_spec.rb
sebarmano: I missed this comment. I'll move the Circle CI confuguration back for it to be under the --heroku flag.

sebarmano:
Ok, I think that the PR is ready for another review. I moved back the Circle CI configuration based on [this comment](https://github.com/thoughtbot/suspenders/pull/756#discussion_r61896289). Also, I've divided the specs in new_project feature spec file to be more descriptive.

tute:
> Splitting the tests will add the time of creating a new app for each one of them

Because we use `before(:all)` instead of `before(:each)`, we recreate a suspenders app once regardless of the number of examples in the spec file, so there should be no performance penalty.

> I think that we can work on that in another PR, as it is related to this same issue but is not exactly the same thing. This PR is solving a problem related to not having the Heroku setup for local apps. The comment you mention seems more like a new feature of suspenders, right?

Makes sense to me. 👍

spec/features/new_project_spec.rb
```diff
@@ -223,6 +223,33 @@
...
+    expect(bin_setup).to include("heroku run rake db:migrate --exit-code "\
+                                 "--app #{app_name.dasherize}-staging-pr-$1")
+    expect(bin_setup).to include("heroku ps:scale worker=1 "\
+                                 "--app #{app_name.dasherize}-staging-pr-$1")
+    expect(bin_setup).to include("heroku restart "\
+                                 "--app #{app_name.dasherize}-staging-pr-$1")
+    expect(File.stat(bin_setup_path)).to be_executable
+  end
+
+  it "creates heroku pipelines config file" do
```
tute: This description should be something like "creates deploy script".

lib/suspenders/generators/app_generator.rb
```diff
@@ -165,18 +166,22 @@ def setup_git
       end
     end
 
+    def create_local_heroku_setup
+      say "Creating local Heroku setup"
+      build :provide_review_apps_setup_script
+      build :create_heroku_pipelines_config_file
+      build :provide_deploy_script
```
tute: While we are on this, should we make these names more homogeneous? They all create files, but two start with `provide` and one with `create`.

tute:
Looking great, left two minor comments. Thank you! 🎉

lib/suspenders/generators/app_generator.rb
sebarmano: Yes, makes sense. I'll change this and what you mention in the previous comment.

sebarmano:
I just pushed the last changes and I think it's again ready for another review. In addition to the changes you mentioned, I changed the name of the methods and tests related to app.js file to mention the "manifest". I think that name is more accurate and is the name Heroku uses. In this way, for example, if someone wants to know more about what this file does, it will be easier to google for it after the methods' names and their contents. Feel free to tell me if this doesn't make sense, I can rename back the methods.

lib/suspenders/generators/app_generator.rb
```diff
@@ -165,18 +166,22 @@ def setup_git
       end
     end
 
+    def create_local_heroku_setup
+      say "Creating local Heroku setup"
+      build :create_review_apps_setup_script
+      build :create_heroku_application_manifest_file
+      build :create_deploy_script
```
tute: We need to move up here (and the related spec) the `set_up_heroku_specific_gems` step.

Thank you!

tute:
Looking great! Added a note on a missing step. Thanks!

sebarmano:
Ready for another review (thanks for taking the time to review this like a million times!). I moved the Heroku gems method to be run by default. 
The only thing that I'm not 100% sold with is the test. I left the test as it was before (just moved it to the right place), but it's testing just that the gem is in the Gemfile, not that is inside the production & staging block. I couldn't find a way of testing it in an easy way, but if you think of one I'm happy to add that too.

tute:
Hey, thank you for your patience with the iterative work and feedback! :)

You know what? If we always add that gem into the Gemfile, we can directly edit the `template/Gemfile.erb` file, and add the gem in the production environment, after `rack-timeout` so we keep the alphabetical order. We can then drop the code for setup heroku gems, but I'd leave the test where it is.

Thanks again for your work and patience! :)

sebarmano:
Sure! This is good because I learn how the codebase works with each step.
I just pushed another version for review. As we removed the method that was adding the `rails_stdout_logging` I was able to also remove a method from the Heroku adapter and its test that were there just because of that gem. This is just a heads up so you can tell me if I went too far removing that.

tute:
The spec would document why we have that line there, and in that sense it is useful. Anyway Rails 5 is just around the corner and we won't need that gem anymore, so this looks good. Will test soon-ish, and when I see everything works as we expect, I'll merge. Thank you! 🎉

sebarmano:
Just to clarify, I didn't remove the test in the new project feature spec referencing that gem, which is still there and testing it. 
In the heroku adapter there was a method to add that gem (and the corresponding spec in the heroku adapter tests). I removed that method (and test) which I think is not needed anymore because the gem is added via the Gemfile template. 

Also, just found a typo in that I'll fix now.

tute:
> Just to clarify, I didn't remove the test in the new project feature spec referencing that gem, which is still there and testing it. 
> In the heroku adapter there was a method to add that gem (and the corresponding spec in the heroku adapter tests). I removed that method (and test) which I think is not needed anymore because the gem is added via the Gemfile template.

That is perfect. 👏 👍

sebarmano:
Just fixed the typo in this last push, so it's basically the same as before.

tute:
Thanks so much! 🎉

PR #760, opened 2016-05-09
Provide suspenders-specific help output

pacso:
This updates the command line output when you run `suspenders -h`.

Rather than receiving the rails usage and help messages, it will now provide instructions on how to use suspenders instead.

### DIFF
--- a/USAGE
+++ b/USAGE
@@ -1,0 +1,13 @@
+Description:
+    Suspenders is a Rails template with thoughtbot standard
+    defaults, ready to deploy to Heroku.
+
+    For full details on the changes we've made compared to
+    a vanilla Rails app, check our GitHub project:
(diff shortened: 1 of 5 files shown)


USAGE
```diff
@@ -0,0 +1,12 @@
+Description:
+    Suspenders is the base Rails application used at thoughtbot.
```
tute: I'd say as a first line:

> Suspenders is a Rails template with thoughtbot standard defaults, ready to deploy to Heroku.

tute:
This is great, thank you! Left a small comment around suspenders tagline. Can you please address, and squash your commits together? Then I'll merge. Thanks! :)

pacso:
@tute - should all be sorted. :)

tute:
Thank you!

PR #762, opened 2016-05-12
Bugfix: configure static_cache_control in production

tute:
`1.year.to_i` wasn't being parsed nor raising an error, failing silently
to add that line to the `production.rb` file.

### DIFF
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -123,6 +123,13 @@ RSpec.describe "Suspend a new project with default configuration" do
     )
   end
 
+  it "configures static_cache_control in production" do
+    prod_env_file = IO.read("#{project_path}/config/environments/production.rb")
+    expect(prod_env_file).to match(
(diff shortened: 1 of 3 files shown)


spec/features/new_project_spec.rb
```diff
@@ -123,6 +123,13 @@
     )
   end
 
+  it "configures static_cache_control in production" do
+    prod_env_file = IO.read("#{project_path}/config/environments/production.rb")
+    expect(prod_env_file).to match(
+      /config.static_cache_control = "public, max-age=.+"/,
```
teoljungberg: What hinders us from now asserting against the year aswell?

tute: > What hinders us from now asserting against the year aswell?

I don't think the value is relevant enough. We say "a year" to mean "CDNs and browsers can cache this for long", and we might write the year as an integer, as `1.year.to_i`, etc.

teoljungberg: Sure!

PR #765, opened 2016-05-20
Add a JavaScript base generator

mike-burns:
We use different foundations for JavaScript in our projects: jQuery,
Ember, React, and so on. Move the `jquery-rails` gem into a generator so
that we can easily change it as desired.

This adds a `suspenders:javascript_base` generator that adds our desired
JavaScript base library.

This is a branch off of `gen-high_voltage`. Paired with Justin Moore.


derekprior:
What about jquery-ujs? It will not function without jquery. 

If you are building a server rendered HTML app with Rails, there's a lot you need to know about if you decide not to use jquery/jquery-ujs.

To me it seems like the more common distinction is API or not? If we move suspenders to generators then you'd kick off your new app using the correct rails generator flags for this.

mike-burns:
You raise two good points:

- The generator would need to be in charge of `app/assets/javascripts/application.js`.
- If we switch away from jQuery as the default, there's a bit more to change.
- Rails assumes `require jquery_ujs` will work.

This PR moves the `gem "jquery"` into a separate generator, which will allow us to kill the `suspenders` executable (well, eventually). This generator should also handle the `app/assets/javascripts/application.js` file. Whoever wants to change _that_ generator -- hopefully in the far future -- will need to deal with losing `jquery_ujs`.

But: `gem "jquery-rails"` is in the default Rails `Gemfile`, uncommented by default. So it is perhaps it is far too early/controversial/irrelevant for me to introduce generators for these:

* sqlite3
* sass-rails
* uglifier
* coffee-rails
* jquery-rails
* turbolinks
* jbuilder
* sdoc
* byebug
* web-console
* spring
* rails

What does the audience say?

derekprior:
If I understand the goals of this endeavor, and I may not, the point is to have the suspenders binary go away and to generate rails applications with `rails new`? Then we'd add `suspenders` as a development dependency so you can use its generators to add/remove things from there?

Assuming that's correct, I don't think we should have generators that add rails' default JavaScript stack to a project. We should just use the flags that `rails new` already provides us. We can skip javascript, turbolinks, and more with those flags. We can strip the gemfile down to a bare minimum with the `--api` flag (assuming Rails 5). That generates the following Gemfile (comments removed):

```ruby
source 'https://rubygems.org'

gem 'rails', '>= 5.0.0.rc1', '< 5.1'
gem 'sqlite3'
gem 'puma', '~> 3.0'

group :development, :test do
  gem 'byebug', platform: :mri
end

group :development do
  gem 'listen', '~> 3.0.5'
  gem 'spring'
  gem 'spring-watcher-listen', '~> 2.0.0'
end

gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```

More directly, what projects have we worked on where we've said, "I wish I didn't have jquery in this project?" that wouldn't already be covered by `--skip-javascript` or `--api` on `rails new`?

mike-burns:
Yup, I'm convinced. Closing this PR. Thanks for the guidance, Derek.

derekprior:
We are going to have a new problem though: We'll probably want to have a way to document the flags we'd typically start with for various types of projects. But we can cross that bridge when we come to it.

mike-burns:
My solution will be: a PR to dotfiles that adds an alias ... called `suspenders`!

PR #767, opened 2016-05-20
Introduce a stylesheet_base generator

mike-burns:
This introduces the `suspenders:stylesheet_base` generator, which adds
our typical foundations for CSS. In this case it is Bourbon, Neat, and
Refills.

Of note, the `application.scss` must list the imports in a specific
order.

Paired with Justin Moore.

### DIFF
--- a/lib/suspenders/generators/stylesheet_base_generator.rb
+++ b/lib/suspenders/generators/stylesheet_base_generator.rb
@@ -1,0 +1,27 @@
+require "rails/generators"
+
+module Suspenders
+  class StylesheetBaseGenerator < Rails::Generators::Base
+    source_root File.expand_path(
+      File.join("..", "..", "..", "templates"),
+      File.dirname(__FILE__))
+
+    def add_stylesheet_gems
+      gem "bourbon", "5.0.0.beta.6"
+      gem "neat", "~> 1.7.0"
+      gem "refills", group: [:development, :test]
+    end
+
+    def add_css_config
+      copy_file(
+        "application.scss",
+        "app/assets/stylesheets/application.scss",
+        force: true,
+      )
+    end
+
+    def remove_prior_config
+      remove_file "app/assets/stylesheets/application.css"
+    end
+  end
+end
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -264,7 +264,24 @@ RSpec.describe "Suspend a new project with default configuration" do
     expect(gemfile).to match(/high_voltage/)
   end
 
+  it "adds and configures bourbon, neat, and refills" do
+    gemfile = read_project_file("Gemfile")
+
+    expect(gemfile).to match(/bourbon/)
+    expect(gemfile).to match(/neat/)
+    expect(gemfile).to match(/refills/)
+  end
+
+  it "configures bourbon, neat, and refills" do
+    app_css = read_project_file(%w(app assets stylesheets application.scss))
+    expect(app_css).to match(/normalize-rails.*bourbon.*neat.*base.*refills/m)
+  end
+
   def analytics_partial
     IO.read("#{project_path}/app/views/application/_analytics.html.erb")
+  end
+
+  def read_project_file(path)
+    IO.read(File.join(project_path, *path))
   end
 end
--- a/lib/suspenders.rb
+++ b/lib/suspenders.rb
@@ -1,6 +1,7 @@
 require 'suspenders/version'
 require 'suspenders/generators/app_generator'
 require 'suspenders/generators/static_generator'
+require 'suspenders/generators/stylesheet_base_generator'
 require 'suspenders/actions'
 require "suspenders/adapters/heroku"
 require 'suspenders/app_builder'
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -318,12 +318,6 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
       copy_file "Procfile", "Procfile"
     end
 
-    def setup_stylesheets
-      remove_file "app/assets/stylesheets/application.css"
-      copy_file "application.scss",
-                "app/assets/stylesheets/application.scss"
-    end
-
     def install_refills
       generate "refills:import", "flashes"
       remove_dir "app/views/refills"
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -249,6 +249,7 @@ module Suspenders
       run("spring stop")
 
       generate("suspenders:static")
+      generate("suspenders:stylesheet_base")
 
       bundle_command "install"
     end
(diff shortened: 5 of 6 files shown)


lib/suspenders/generators/stylesheet_base_generator.rb
```diff
@@ -0,0 +1,23 @@
...
+      gem "neat", "~> 1.7.0"
+      gem "refills", group: [:development, :test]
+    end
+
+    def add_css_config
+      copy_file(
+        "application.scss",
+        "app/assets/stylesheets/application.scss",
+        force: true,
+      )
```
tute: What should we do if the host application has an `app/assets/stylesheets/application.css` (non SCSS) file?

tute:
This looks very happy to me. ❤️  🚢

lib/suspenders/generators/stylesheet_base_generator.rb
mike-burns: Interesting. How about: copy our file over and issue a warning message?

tute: > How about: copy our file over and issue a warning message?

You mean appending to their css, and changing their format? Or copying it too, as we do now?

I'm inclined to changing their file, but I'm not sure how-meaningful issues will arise from this.

mike-burns: I mean: leave this line as-is, but add an additional check for `app/assets/application.css` and if it exists print a warning.

I doubt this will come up, but it's considerate to give them the heads-up.

tute: > I mean: leave this line as-is, but add an additional check for app/assets/application.css and if it exists print a warning.

If we change and move the original file, you see the changes in the git diff and act accordingly. If we add the new file, it might come as a surprise that both files exist, and potentially be more confusing, which makes me prefer modifying the existing file.

If you'd rather add the warning message, we can keep this in mind, and if people trip on this we can later improve. Your call.

lib/suspenders/generators/stylesheet_base_generator.rb
```diff
@@ -0,0 +1,23 @@
+require "rails/generators"
+
+module Suspenders
+  class StylesheetBaseGenerator < Rails::Generators::Base
+    source_root File.expand_path(
```
teoljungberg: What do you think of keeping the style consistent and re-write this as:

```
source_root(
  File_expand_path("..", "..", "..", templates),
  File.dirname(__FILE__),
)
```

lib/suspenders/generators/stylesheet_base_generator.rb
```diff
@@ -0,0 +1,23 @@
...
+      File.join("..", "..", "..", "templates"),
```
teoljungberg: The feels clumsy to me, is there a better way we can write this?

spec/features/new_project_spec.rb
```diff
@@ -264,7 +264,24 @@ def app_name
     expect(gemfile).to match(/high_voltage/)
   end
 
+  it "adds and configures bourbon, neat, and refills" do
+    gemfile = read_project_file("Gemfile")
```
teoljungberg: Where is the generator run, are they run by default?

lib/suspenders/generators/stylesheet_base_generator.rb
teoljungberg: Thinking about this, isn't this the same as `Rails.root`?

lib/suspenders/generators/stylesheet_base_generator.rb
mike-burns: Everything about this line annoys me. Is there a better way to do this?

spec/features/new_project_spec.rb
mike-burns: In a `before(:all)`. It takes less than 10 minutes to run the generator, so we do it exactly once.

lib/suspenders/generators/stylesheet_base_generator.rb
mike-burns: I want that! I hate these three lines.

spec/features/new_project_spec.rb
teoljungberg: Cool, just asking.

lib/suspenders/generators/stylesheet_base_generator.rb
mike-burns: Should we also do something if e.g. `app/assets/stylesheets/application.less` exists? And `app/assets/stylesheets/application.sass`? Why should we treat `.css` specially?

Thinking about this more, I'm tempted to issue no warning and not touch anything except `app/assets/stylesheets/application.scss`.

What do you think -- is that setting someone up for hours of annoying debugging?

mike-burns:
OK so the status of this seems to be:

- Everyone hates the hideous `source_root` line.
- Concern over users who have an existing `application.css`.

@tute @teoljungberg Are these things that should be fixed in this PR?

tute:
> Everyone hates the hideous source_root line.

If we know of a better way we improve it yes, but if not, it shouldn't hold off this PR I think.

> Concern over users who have an existing `application.css`.

Around this your previous last comment said:

> Thinking about this more, I'm tempted to issue no warning and not touch anything except `app/assets/stylesheets/application.scss`.
>
> What do you think -- is that setting someone up for hours of annoying debugging?

I'm ok with that workaround. But there is a bug right now: `application.css` is not removed any longer:

```
test1 master % tree app/assets/stylesheets
app/assets/stylesheets
├── application.css
├── application.scss
└── base
    ├── _base.scss
    ├── ...
```

teoljungberg:
> > Everyone hates the hideous source_root line.

> If we know of a better way we improve it yes, but if not, it shouldn't hold off this PR I think.

I agree with @tute that that can be left for a future PR.

mike-burns:
>  But there is a bug right now: application.css is not removed any longer

Aha! I had missed that. Obvious now. Thanks, fixed.

Now the build is failing because of changes to the gem ecosystem. We should fix `master` before I merge.

mcmire:
I just saw this and this seems like a great addition. Thanks @mike-burns.

rikkipitt:
I haven't had chance to properly delve into this new generator, but are you guys getting the following when bundling? It's because the refills gem is already in the development/test block in the Gemfile while the new generator appends it again to the bottom of the file...

```
Your Gemfile lists the gem refills (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
```

mike-burns:
@rikkipitt I'm not seeing that. Attaching a sample `Gemfile` that I see. I did this to get it:

```sh
mkdir tmp && cd tmp
../suspenders/bin/suspenders sample
```
[Gemfile.txt](https://github.com/thoughtbot/suspenders/files/365551/Gemfile.txt)

rikkipitt:
@mike-burns Looks like someone else has found this to be the case too in #786

PR #770, opened 2016-05-22
Update the viewport meta tag

tysongach:
- Add `width=device-width` to set the virtual viewport to the width of
the device's screen
- Add `shrink-to-fit=no` to tell Safari 9 not to shrink overflowing
content to fit the width of the virtual viewport

References:
- https://css-tricks.com/snippets/html/responsive-meta-tag/
- https://bitsofco.de/ios-safari-and-shrink-to-fit/


tute:
LGT 🚢 !

PR #772, opened 2016-05-27
Fix absence of heroku-pipelines in heroku plugins listing.

rubylibre:
Since heroku-pipelines became publicly available it is no longer an external plugin. This prevents it to appear in the listing `heroku plugins` which is used to check for it's availability.

By using `heroku help` and grep for `pipelines` it is still possible to do this check. If I'm right this wil work with in both situations: before and after general availability of this plugin.

### DIFF
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -53,7 +53,7 @@ module Suspenders
       def create_heroku_pipeline
         pipelines_plugin = `heroku help | grep '^\s*pipelines'`
         if pipelines_plugin.empty?
-          puts "You need heroku pipelines plugin. Run: heroku plugins:install heroku-pipelines"
+          puts "You need heroku pipelines plugin. Run: heroku update"
           exit 1
         end
 
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -51,7 +51,7 @@ module Suspenders
       end
 
       def create_heroku_pipeline
-        pipelines_plugin = `heroku plugins | grep pipelines`
+        pipelines_plugin = `heroku help | grep '^\s*pipelines'`
         if pipelines_plugin.empty?
           puts "You need heroku pipelines plugin. Run: heroku plugins:install heroku-pipelines"
           exit 1
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -51,7 +51,7 @@ module Suspenders
       end
 
       def create_heroku_pipeline
-        pipelines_plugin = `heroku help | grep '^\s*pipelines'`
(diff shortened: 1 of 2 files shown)


spec/support/fake_heroku.rb
```diff
@@ -6,8 +6,9 @@ def initialize(args)
   end
 
   def run!
-    if @args.first == "plugins"
-      puts "heroku-pipelines@0.29.0"
+    if @args.first == "help"
+      # spaces in front are intentional
+      puts "  pipelines      #  manage collections of apps in pipelines"
```
zamith: I understand that the spaces are required because of the regex, but are they mandatory? Can't we change the regex?

rubylibre: The spaces are here to test the regex :)

zamith: Right. My question was, can we change the regex and keep the functionality?

rubylibre: You mean both? I suppose it can.

rubylibre: If you mean: does it work. Than yes it does.

zamith: I mean does it work, and does it provide the same level of confidence as the one with spaces. Because if it does, I'd make spaces optional, which would remove the need to have the comment.

rubylibre: The real `heroku help` outputs two spaces in the beginning. In the grep regex the spaces are optional. So I can leave them out of the test. But than the criteria of the test not correct, is it? I can leave the comment out if it fits better in you coding style.

zamith: In general I'm not a fan of comments, so if there's a way to remove them without losing information it would be great. Let's see what @tute thinks.

This is a small thing, so I think we can merge either way.

tute: I agree it's better not to need comments. Also, I think this feature will be more reliable if we only assert on the presence of `pipelines`, without any information on specific formatting. That way, if in a future the amount of spaces changes, behavior will always be the same.

Can we make it so that the spaces are not relevant?

Thank you for your contribution!

lib/suspenders/adapters/heroku.rb
```diff
@@ -51,9 +51,9 @@ def create_heroku_application_manifest_file
       end
 
       def create_heroku_pipeline
-        pipelines_plugin = `heroku plugins | grep pipelines`
+        pipelines_plugin = `heroku help | grep pipelines`
         if pipelines_plugin.empty?
-          puts "You need heroku pipelines plugin. Run: heroku plugins:install heroku-pipelines"
+          puts "You need heroku pipelines plugin. Run: heroku update"
```
tute: This is now `brew upgrade heroku-toolbelt`. Fixing.

tute:
Thank you! Merged into master. :D

PR #777, opened 2016-06-15
APPLICATION_HOST bug fix in production environment

### LINKED ISSUE #776: Set the APPLICATION_HOST atop the production.rb
The `config/environments/production.rb` looks like so:

```ruby
  config.cache_classes = true
  config.eager_load = true
  config.consider_all_requests_local       = false
  config.action_controller.perform_caching = true
  config.assets.js_compressor = :uglifier
  config.assets.compile = false
  config.assets.digest = true
  config.log_level = :debug
  config.action_controller.asset_host = ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))
  config.i18n.fallbacks = true
  config.active_support.deprecation = :notify
  config.log_formatter = ::Logger::Formatter.new
  config.active_record.dump_schema_after_migration = false
  if ENV.fetch("HEROKU_APP_NAME", "").include?("staging-pr-")
    ENV["APPLICATION_HOST"] = ENV["HEROKU_APP_NAME"] + ".herokuapp.com"
  end
  config.middleware.use Rack::CanonicalHost, ENV.fetch("APPLICATION_HOST")
  config.middleware.use Rack::Deflater
  config.static_cache_control = "public, max-age=31557600"
  config.action_mailer.default_url_options = { host: ENV.fetch("APPLICATION_HOST") }
end
Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
```

As you can see, it sets `ENV["APPLICATION_HOST"]` in the middle. However, it should be set at the top because it is used earlier, like so:

```ruby
if ENV.fetch("HEROKU_APP_NAME", "").include?("staging-pr-")
  ENV["APPLICATION_HOST"] = ENV["HEROKU_APP_NAME"] + ".herokuapp.com"
end

  config.cache_classes = true
  config.eager_load = true
  config.consider_all_requests_local       = false
  config.action_controller.perform_caching = true
  config.assets.js_compressor = :uglifier
  config.assets.compile = false
  config.assets.digest = true
  config.log_level = :debug
  config.action_controller.asset_host = ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))
  config.i18n.fallbacks = true
  config.active_support.deprecation = :notify
  config.log_formatter = ::Logger::Formatter.new
  config.active_record.dump_schema_after_migration = false
  config.middleware.use Rack::CanonicalHost, ENV.fetch("APPLICATION_HOST")
  config.middleware.use Rack::Deflater
  config.static_cache_control = "public, max-age=31557600"
  config.action_mailer.default_url_options = { host: ENV.fetch("APPLICATION_HOST") }
end

Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
```

This happens in https://github.com/thoughtbot/suspenders/blob/3cf23ed37a94a890f7678a8ec5cd072385f93bb4/lib/suspenders/app_builder.rb#L143-L153. It should not use `#configure_environment` but should instead do something like:

```ruby
inject_into_file("config/environments/#{rails_env}.rb", config, before: "Rails.application.configure do")
```

---

This issue came up when using Heroku PR pipelines: the asset path was incorrectly set for these PR deployments. I manually fixed the `production.rb` for the app, and that resolved it.

tute:
`ENV["APPLICATION_HOST"]` is used before it's defined in the production
environment.

By using `inject_into_file` instead of `configure_environment` we can
control the order of the steps.

[fixes #776]

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -142,14 +142,19 @@ module Suspenders
 
     def enable_rack_canonical_host
       config = <<-RUBY
-if ENV.fetch("HEROKU_APP_NAME", "").include?("staging-pr-")
+
+  if ENV.fetch("HEROKU_APP_NAME", "").include?("staging-pr-")
     ENV["APPLICATION_HOST"] = ENV["HEROKU_APP_NAME"] + ".herokuapp.com"
   end
 
   config.middleware.use Rack::CanonicalHost, ENV.fetch("APPLICATION_HOST")
       RUBY
 
-      configure_environment "production", config
+      inject_into_file(
+        "config/environments/production.rb",
+        config,
+        after: "Rails.application.configure do",
+      )
     end
 
     def enable_rack_deflater


mike-burns:
I see you used `:after` instead of `:before` -- looks like it works fine. Any reason to put it inside the block?

lib/suspenders/app_builder.rb
```diff
@@ -142,14 +142,19 @@ def configure_smtp
 
     def enable_rack_canonical_host
       config = <<-RUBY
```
teoljungberg: Since we are on ruby 2.3, we can use `<<~` to remove the leading whitespace.

tute:
> I see you used :after instead of :before -- looks like it works fine. Any reason to put it inside the block?

I can split the string and put the first part before, and the second after. Both ways looked good to me, and all inside was a bit simpler. Do you think we should split them?

> Since we are on ruby 2.3, we can use <<~ to remove the leading whitespace.

That is nice! If we take it out, I'll use this.

mike-burns:
Nah, I see your point now. LGTM.

PR #782, opened 2016-06-25
Extract methods in new_project_spec file

tute:
Reads environment files once per file run.


tute:
Started adding a feature (`log_level = :fatal` for test environment), but then I didn't think it was an improvement. I still kept the improvements to this spec file.

teoljungberg:
Looks good!

tute:
Thanks for your review, Teo!

PR #784, opened 2016-07-03
Only require refills once

jcmorrow:
Refills is added to the Gemfile after install via the
 `StylesheetBaseGenerator`. If it is also present in the Gemfile
 template then bundler will issue a warning that refills is required
 twice.

 Changes:

  - Remove refills from Gemfile template


tute:
Looks good to merge! Thank you, and happy 4th of July.

PR #790, opened 2016-07-20
Ensure stylesheet_base generator runs with a clean bundle

### LINKED ISSUE #568: Sass::SyntaxError: File to import not found or unreadable: refills/flashes.
I'm trying to deploy to Heroku.  I've done nothing but generate the app and am attempting to deploy.  I get this:

Sass::SyntaxError: File to import not found or unreadable: refills/flashes.

I'm using suspenders v 1.28.0

Thanks,
Peter

### LINKED ISSUE #787: Default scss setup for Bitters
It looks like the default generators create a standard setup, but the `*= require_tree .` sprockets directive needs to be replaced with the `@import`s that Bourbon likes. Any way this could happen on its own? May save folks like me some time trying to debug something that looks like this: https://github.com/thoughtbot/bitters/issues/234. (I thought I was having that same issue initially.)

tute:
Static and Stylesheet suspenders' generators add gems to the `Gemfile`,
but don't run `bundle install` after the fact, making installations with
clean gem sets fail with errors like:

```
 gemfile  high_voltage
generate  suspenders:stylesheet_base
Could not find gem 'high_voltage' in any of the gem sources listed in
your Gemfile or available on this machine.
Run `bundle install` to install missing gems.
     run  bundle install
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Fetching dependency metadata from https://rubygems.org/
Resolving dependencies...
Using rake 11.2.2
```

By adding `bundle install` calls after modifying the `Gemfile` we ensure
each step will have the necessary dependencies for application code to
run.

[fixes #568]
[fixes #787]

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -322,15 +322,6 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
       copy_file "Procfile", "Procfile"
     end
 
-    def install_refills
-      generate "refills:import", "flashes"
-      remove_dir "app/views/refills"
-    end
-
-    def install_bitters
-      run "bitters install --path app/assets/stylesheets"
-    end
-
     def setup_default_directories
       [
         'app/views/pages',
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -39,9 +39,6 @@ module Suspenders
       invoke :setup_secret_token
       invoke :create_suspenders_views
       invoke :configure_app
-      invoke :setup_stylesheets
-      invoke :install_bitters
-      invoke :install_refills
       invoke :copy_miscellaneous_files
       invoke :customize_error_pages
       invoke :remove_config_comment_lines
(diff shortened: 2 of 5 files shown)


jackvnimble:
LGTM

Though if bundle needs to be run after any generate maybe we could overwrite it here like: 
```
def generate(cmd)
  super
  bundle_command "install"
end
```

tute:
Thanks for your review. Indeed. Or run on steps that modify the Gemfile.

tute:
Added a commit which would render one of the `bundle install`s unnecessary.

jackvnimble:
Nice work, looks really good! I'm not familiar with rails generators though, does every instance var get run automatically inside `StylesheetBaseGenerator`?

tute:
What instance variables do you mean?

mike-burns:
LGTM.

jackvnimble:
Ah sorry @tute, I meant instance _methods_ (:install_refills, :install_bitters). Was getting late over here last night, I don't know how you do it ;) 

I'm gonna assume it's a meta thing.

tute:
> Does every instance method run automatically?

Yes, public methods defined in the class execute when the generator is run.

PR #791, opened 2016-07-20
Prefer using public_file_server.headers.

nickcharlton:
In Rails 5, deploying to production environments will print a deprecation
warning when configuring config.static_cache_control. This changes the
configuration to use the new style config.public_file_server.headers.


tute:
LGTM 🚀

PR #792, opened 2016-07-22
Remove turbolinks from application.js file

### LINKED ISSUE #788: Missing turbolinks gem
Seems like the application js file is sourcing turbolinks, but the gem is missing from the Gemfile. (I didn't have the turbolinks gem installed before, since this is a new machine.) Adding `gem "turbolinks"` fixed it for me locally -- not sure if this is something needed in Suspenders.

tute:
Suspenders doesn't use turbolinks by default. It doesn't include it in
the Gemfile, but the generated `application.js` was trying to include
it, resulting in a failure.

This fix make sure a newly generated application doesn't include
the turbolinks reference.

[fixes #788]

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -43,6 +43,14 @@ module Suspenders
         'raise_delivery_errors = false', 'raise_delivery_errors = true'
     end
 
+    def remove_turbolinks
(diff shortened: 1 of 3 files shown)


mike-burns:
Code LGTM.

Issue #793, opened 2016-07-25
Uses sprockets-es6 - Causes deprecation warnings.


ACPK:
sprockets-es6 is causing depreciation issues for Sprockets, and the gem author says that the gem will never reach v1.0.  May I ask how this gem is being used?

tute:
Related with https://github.com/thoughtbot/suspenders/issues/726

tute:
@ACPK: while sprockets 4 is not released, what is a good alternative? Cc @MattMSumner 

Thanks!

MattMSumner:
https://github.com/browserify-rails/browserify-rails seems like a possibility. @BlakeWilliams do you have any thoughts on alternative approaches?

MattMSumner:
@ACPK to answer your question, the gem is being used to allow writing ES6. The javascript community as a whole has moved towards using transpilers, such as [babel](https://babeljs.io), to get access to features we've come to expect of a modern programming language. In the past, we have used coffeescript to fill in the gaps we've had with ES5 but I've found that coffeescript has stopped moving forward while ES6 promises to provide all the features we're looking for without significantly changing the syntax, in a forwards compatible way.

I introduced sprockets-es6 to go along with [this PR](https://github.com/thoughtbot/guides/pull/387) on our guides.

MattMSumner:
Here's the issue on sprockets 4.0 milestone around babel and javascript modules https://github.com/rails/sprockets/issues/73

tute:
sprockets v4 beta3 recently released: https://github.com/rails/sprockets/releases/tag/v4.0.0.beta3

tute:
Closing as duplicate of https://github.com/thoughtbot/suspenders/issues/726. Soon we'll add `babel-transpiler` and `sprockets` 4 (I'm not against doing it while it's still in beta).
Thanks!

PR #796, opened 2016-09-02
Upgrade sprockets to 4.0.0.beta3

MattMSumner:
Why:

Currently, a newly generated suspenders rails app has deprecation
warnings [per this issue]. This is due to using the `sprockets-es6`
which was not designed to be a lasting solution to transpiling ES6 to
ES5.

[per this issue]: https://github.com/thoughtbot/suspenders/issues/793

This PR:

Removes `sprockets-es6` in favor of updating `sprockets` to 4.0.0.beta3
and adding the `babel-transpiler` gem. This resolves [this issue].

[this issue]: https://github.com/thoughtbot/suspenders/issues/793


MattMSumner:
Having generated an app, this works for transpiling es6. However, there seems to be an issue with bourbon and sprockets. I'm getting
```
Sprockets::FileNotFound in Posts#index
couldn't find file '...gems/bourbon-5.0.0.beta.6/core/bourbon/settings/_settings.scss'
```

It seems to be an issue with reassigning variables within a function but I'm still investigating.

tute:
@tysongach have you seen an error like this before?

tute:
Maybe @whmii has?

whmii:
@tute I have not. that is the right path though.

MattMSumner:
@whmii I seem to only have a problem with this setting https://github.com/thoughtbot/bourbon/blob/master/core/bourbon/settings/_settings.scss#L34. It looks like an issue being used here https://github.com/thoughtbot/bourbon/blob/master/core/bourbon/library/_modular-scale.scss#L70. It seems to me that these lines suggest this value should be a list https://github.com/thoughtbot/bourbon/blob/master/core/bourbon/library/_modular-scale.scss#L74-L75 not a singular value?

If I comment out this line https://github.com/thoughtbot/bitters/blob/a1a0ca9188d63596a67379fd07ae8053aed0e653/core/_typography.scss#L15 the page loads and everything works.

MattMSumner:
It seems like this also happens on master. I've created an issue to address: https://github.com/thoughtbot/suspenders/issues/797

tute:
@MattMSumner is this error exactly the same as what happens in master? If so, do you think we should merge this as a step in the right direction, and fix the error in another unrelated PR?

tute:
beta4 has been released: https://github.com/rails/sprockets/releases/tag/v4.0.0.beta4

MattMSumner:
I tried out beta4 with no luck. I've also fought with this for too long and am closing in favor of https://github.com/thoughtbot/suspenders/pull/802

PR #799, opened 2016-10-12
Prevents extra spacing in generated bin/setup with heroku

Dorian:
Not sure about the first lines, it was generated fine. But the last one had extra leading spaces.

e.g.:
```
else
  printf 'Ask for access to the "camera-production" Heroku app
'
fi


          git config heroku.remote staging
```

(Not sure about the consecutive empty lines either)

### DIFF
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -7,10 +7,10 @@ module Suspenders
 
       def set_heroku_remotes
         remotes = <<-SHELL.strip_heredoc
-          #{command_to_join_heroku_app('staging')}
-          #{command_to_join_heroku_app('production')}
+#{command_to_join_heroku_app('staging')}
+#{command_to_join_heroku_app('production')}
 
-          git config heroku.remote staging
+git config heroku.remote staging
         SHELL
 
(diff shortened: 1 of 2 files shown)


tute:
huh! I'm surprised. Why doesn't `strip_heredoc` do the trick?

Would you please adjust the assertion in https://github.com/thoughtbot/suspenders/blob/master/spec/features/heroku_spec.rb#L36 to catch as a regression? Then I'll merge.

Thanks!

Dorian:
@tute So [`strip_heredoc`](http://apidock.com/rails/String/strip_heredoc) looks for the line with the least leading spaces, but [`command_to_join_heroku_app`](https://github.com/thoughtbot/suspenders/blob/master/lib/suspenders/adapters/heroku.rb#L86-L97) doesn't have leading spaces so it doesn't remove them.

So, yes, I changed the specs.

tute:
Thank you! I reindented the code and used Ruby 2.3 squiggly heredoc operator. I kept your authorship, and merged in.

Thanks for your work!

Dorian:
@tute Wow, thanks, I didn't know about `<<~`, pretty cool!

Issue #797, opened 2016-09-16
New app fails with `Invalid null operation: "null gt null".`


MattMSumner:
Here's an example app https://github.com/MattMSumner/test-sus-2.

Visiting http://localhost:3000/high_scores produces:

![screen shot 2016-09-16 at 10 22 43 am](https://cloud.githubusercontent.com/assets/759721/18589333/2b9178d2-7bf8-11e6-8576-8da144547eaf.png)

Commenting out https://github.com/MattMSumner/test-sus-2/blob/master/app/assets/stylesheets/base/_typography.scss#L15 "fixes" the issue.

MattMSumner:
Seems like https://github.com/MattMSumner/test-sus-2 is working for other but https://github.com/MattMSumner/test-sus is an example where I've upgraded sprockets and this doesn't work. I get the following error:

![screen shot 2016-09-16 at 10 22 43 am](https://cloud.githubusercontent.com/assets/759721/18594960/79265320-7c10-11e6-9d35-b5c41dd088a3.png)

tute:
Does this still happen on latest release? Thank you!

tute:
cc @MattMSumner. Thank you!

tute:
I assume #802 addressed this issue. Thanks!

PR #802, opened 2016-10-29
Remove sprocket-es6

MattMSumner:
Why:

sprockets-es6 is causing depreciation issues for Sprockets, and the gem author says that the gem will never reach v1.0.

This PR:

Removes sprockets-es6. We would like to introduce the babel-transpiler gem in the future but it requires sprockets 4. While sprockets 4 has a beta release, this seems to be causing issues with bourbon neat and bitters per https://github.com/thoughtbot/suspenders/issues/797. Until this is resolved I suggest adding ES6 support on a project by project basis.

Resolves: https://github.com/thoughtbot/suspenders/issues/793


tute:
LGTM Matt. I'd consider using babel-transpiler and a beta release for Sprockets in the short term. I wouldn't add ES6 on a project by project basis, I'd vote for building the best solution we know for ES6 in Rails apps into suspenders (even if it requires beta sprockets).

This looks good to 🚢. Thank you! :)

tute:
> While sprockets 4 has a beta release, this seems to be causing issues with bourbon neat and bitters per #797.

Nevermind my idea, if we can't run on beta. Thanks!

tute:
After https://github.com/thoughtbot/suspenders/pull/796#issuecomment-257061826 I see there is no good solution yet. Read the threads in the opposite order.
This LGTM. Thank you!

mike-burns:
I don't understand why, but I have Suspenders plus this:

```
% bundle | grep sprockets
Using sprockets 3.7.1
Using sprockets-es6 0.9.2
Using sprockets-rails 3.2.0
```

I am able to write ES6 in files with names ending in `.es6`, deploy to Heroku, and see those files be served as JavaScript. I've seen no deprecation warnings or Bourbon issues, too.

Before I add sprockets-es6 back to Suspenders, what should I do to see the issues raised elsewhere?

PR #803, opened 2016-11-02
Add real support for MySQL

amree:
Previously postgresql is used as the database adapter even though mysql has been set as the database from the command line config. This will ensure the generator will create a MySQL rails app.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -225,6 +225,10 @@ config.public_file_server.headers = {
         force: true
     end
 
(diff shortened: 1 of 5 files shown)


tute:
Thank you for your contribution.

We don't use MySQL, so we wouldn't be able to ensure the quality and behavior of this new feature.

Moreover, Suspenders is intentionally opinionated: it picks the best defaults we have found for the new Rails projects.

We won't merge this PR. Thank you again for proposing and sharing your work!

amree:
Got it. Thanks for your time.

PR #806, opened 2016-11-07
Specify Bourbon version pessimistically

tysongach:
This matches how we specify the Bourbon dependency in Bitters and means
we won't have to coordinate releases between the two so closely.

Fixes https://github.com/thoughtbot/suspenders/issues/805


tute:
🚀

PR #810, opened 2016-12-04
Do not create test file in repo root folder

delphaber:
I noticed that line 122 creates a file named `production_config.rb` in the repo root path when you run `rspec spec` on my dev machine. I think the line might be a leftover from a previous commit. Am I right?


tute:
Whops, missed that one. Thanks! :)

PR #811, opened 2016-12-04
Update database_cleaner configuration

delphaber:
Update `database_cleaner` configuration, using the one from their readme (https://github.com/DatabaseCleaner/database_cleaner#rspec-with-capybara-example)

### DIFF
--- a/templates/database_cleaner_rspec.rb
+++ b/templates/database_cleaner_rspec.rb
@@ -1,21 +1,28 @@
 RSpec.configure do |config|
+  config.use_transactional_fixtures = false
+
   config.before(:suite) do
-    DatabaseCleaner.clean_with(:deletion)
+    DatabaseCleaner.clean_with(:truncation)
   end
 
   config.before(:each) do
     DatabaseCleaner.strategy = :transaction
   end
 
-  config.before(:each, js: true) do
-    DatabaseCleaner.strategy = :deletion
+  config.before(:each, type: :feature) do
+    driver_shares_db_connection_with_specs =
+      Capybara.current_driver == :rack_test
+
+    unless driver_shares_db_connection_with_specs
+      DatabaseCleaner.strategy = :truncation
+    end
   end
 
   config.before(:each) do
     DatabaseCleaner.start
   end
 
-  config.after(:each) do
+  config.append_after(:each) do
     DatabaseCleaner.clean
   end
 end
--- a/templates/rails_helper.rb
+++ b/templates/rails_helper.rb
@@ -16,7 +16,6 @@ RSpec.configure do |config|
   config.include Features, type: :feature
   config.infer_base_class_for_anonymous_controllers = false
   config.infer_spec_type_from_file_location!
-  config.use_transactional_fixtures = false
 end
 
 ActiveRecord::Migration.maintain_test_schema!


delphaber:
I had a problem with Capybara, AJAX calls and database_cleaner. It seemed that my `after` hook was being called too early. This resulted in database_cleaner removing records from db and tests failing.

After reading database cleaner readme, I noticed this:

> It's also recommended to use append_after to ensure DatabaseCleaner.clean runs after the after-test cleanup capybara/rspec installs.

I always used `after` instead of `append_after`!

So I tried to copy/paste [their configuration](https://github.com/DatabaseCleaner/database_cleaner#rspec-with-capybara-example) and tests passed again!

They also use a different clean strategy.

I don't know if you ever noticed my same problem, but I wanted to share this pull request anyway. Feel free to reject :)

tute:
Calling in @iwz to help me review this PR. I don't have experience configuring database cleaner, and you might. Thank you! :)

templates/database_cleaner_rspec.rb
```diff
@@ -1,21 +1,28 @@
...
 
-  config.before(:each, js: true) do
-    DatabaseCleaner.strategy = :deletion
+  config.before(:each, type: :feature) do
+    driver_shares_db_connection_with_specs =
+      Capybara.current_driver == :rack_test
+
+    unless driver_shares_db_connection_with_specs
+      DatabaseCleaner.strategy = :truncation
+    end
```
geoffharcourt: Does this fail if we're using debug mode in Capybara, etc? Is there a reason we shouldn't just use `js: true` from spec metadata as our criteria for this strategy?

templates/database_cleaner_rspec.rb
```diff
@@ -1,21 +1,28 @@
 RSpec.configure do |config|
+  config.use_transactional_fixtures = false
+
   config.before(:suite) do
-    DatabaseCleaner.clean_with(:deletion)
+    DatabaseCleaner.clean_with(:truncation)
```
geoffharcourt: 👍

templates/database_cleaner_rspec.rb
```diff
@@ -1,21 +1,28 @@
 RSpec.configure do |config|
+  config.use_transactional_fixtures = false
```
geoffharcourt: I like having this here, but I am mildly concerned a viewer won't realize it's been moved out.

templates/database_cleaner_rspec.rb
```diff
@@ -1,21 +1,28 @@
...
+      DatabaseCleaner.strategy = :truncation
+    end
   end
 
   config.before(:each) do
     DatabaseCleaner.start
   end
 
-  config.after(:each) do
+  config.append_after(:each) do
```
geoffharcourt: Nice! Clever.

templates/database_cleaner_rspec.rb
croaky: If the file name is `database_cleaner`, it feels unexpected to me to have this here. Feels better grouped next to the `ActiveRecord::Migration.maintain_test_schema!` line.

Could rename the file `database.rb` or something

croaky:
Would be good to try this on a couple thoughtbot apps like Hub to confirm the suite performs as expected

templates/database_cleaner_rspec.rb
teoljungberg: I like the rename to `database`, as that means more db configuration can be moved out of the pure spec-helper. Which is nice.

thiagogabriel:
It solves my intermittent "RecordNotFound" failures.
I'm in favor of moving `config.use_transactional_fixtures` back to `rails_helper.rb`. IMO `database_cleaner` has a decent amount of configuration, and in case we decide to remove `database_cleaner` in the future, it is easier to track what needs to be removed.

derekprior:
So it turns out, we no longer need Database Cleaner at all with Rails 5.1+. I recently merged a change that removes it entirely. 🎉

delphaber:
Less is more :D

PR #814, opened 2016-12-25
Update to Ruby 2.4.0

delphaber:
And merry xmas 😄


tute:
Thank you, and happy holidays! :)

PR #817, opened 2017-01-13
Remove 12factor gem component for logging

dijonkitchen:
No longer needed in Rails 5: https://github.com/heroku/rails_12factor#rails-5


tute:
Yep! Thank you!

PR #822, opened 2017-02-07
Update Segment snippet to 4.0.0

tysongach:
I can’t find a change log for the snippet, but I imagine we want to keep this updated.


tute:
🚢

PR #824, opened 2017-02-17
Add support for customization

monfresh:
**Why**:
- Some people might want to configure their Rails app differently from
what suspenders does by default

**How**:
- Allow the user to specify the `-c` flag, followed by a local path
that  contains the custom app generator, app builder, and templates.
For example: `suspenders app -c /path/to/custom/files`
- It is assumed that the custom files are organized in the same
directory structure as suspenders, and that the custom app generator
and builder class names are called `CustomAppGenerator` and
`CustomAppBuilder` respectively.

### DIFF
--- a/bin/suspenders
+++ b/bin/suspenders
@@ -14,10 +14,16 @@ if ARGV.empty?
 elsif ['-v', '--version'].include? ARGV[0]
   puts Suspenders::VERSION
   exit 0
+elsif ARGV[1] == '-c'
+  $LOAD_PATH << "#{ARGV[2]}/lib"
+  require 'suspenders/generators/custom_app_generator'
+  custom_templates_root = "#{ARGV[2]}/templates"
+  custom_generator_class = Suspenders::CustomAppGenerator
 end
 
-templates_root = File.expand_path(File.join("..", "templates"), File.dirname(__FILE__))
-Suspenders::AppGenerator.source_root templates_root
-Suspenders::AppGenerator.source_paths << Rails::Generators::AppGenerator.source_root << templates_root
+templates_root = custom_templates_root || File.expand_path(File.join("..", "templates"), File.dirname(__FILE__))
+generator_class = custom_generator_class || Suspenders::AppGenerator
+generator_class.source_root templates_root
+generator_class.source_paths << Rails::Generators::AppGenerator.source_root << templates_root
 
-Suspenders::AppGenerator.start
+generator_class.start


monfresh:
Hello thoughtbot folks. I love the way suspenders is designed, but I don't agree with everything it does. I've been trying to figure out the best way to make it customizable, and I think this solution works well. I tried it locally and it works great. Let me know what you think, and I'll add tests and a usage section in the README if you are open to supporting something like this.

rikkipitt:
@monfresh Did you know that you can just use the default Rails -m flag? I have a script I use that extends Suspenders with my own template file, it has been working for many Suspenders releases...

monfresh:
@rikkipitt Can you show me an example, please? For example, let's say I want everything that suspenders does, except I want to generate my own custom Gemfile, and I don't want any Heroku stuff.

rikkipitt:
Hi @monfresh I have a command line tool for building projects as mentioned above. Essentially, it calls this (with my Heroku flags removed):

```
suspenders app_name -m "path/to/template.rb"
```

What you do then is look up Thor and create a template file as per these docs: http://guides.rubyonrails.org/rails_application_templates.html

Hope this helps!

Rikki

monfresh:
That's not ideal in my opinion. I like the communicative method names in suspenders, and I think it's much easier to be able to remove calls to methods I don't need. For example, if I don't want any `capybara-webkit` stuff, as far as I can tell, the only way to do that with the `-m` option is to use `remove_file` in my custom template:
```ruby
remove_file 'spec/support/capybara_webkit.rb'
```

It doesn't feel right to me to have suspenders create the file, just to have it removed. It makes more sense to me to be able to tell suspenders to not create the file at all, and the way to do that, would be to remove the line `configure_capybara_webkit` from the `setup_test_environment` section in my `custom_app_generator.rb`.

Similarly, let's say I want to configure RSpec, but I want to use a different template. When using `-m`, I have to essentially recreate what suspenders is doing:

```ruby
remove_file "spec/rails_helper.rb"
copy_file "#{__dir__}/templates/rails_helper.rb", "spec/rails_helper.rb"
```

Here again, suspenders is creating a file that I then have to remove and replace with my own.
Also, my template now is just a collection of commands, without any of them being nicely wrapped inside methods such as `configure_rspec`.

With my proposed solution, all you have to do is provide your own `rails_helper.rb` file. I think this is much more user friendly. You retain the readability of suspenders, and you can easily remove the methods you don't want, and for those that you want to keep but modify, you provide your own template, or you override the method.

monfresh:
Any thoughts on this, @tysongach, @tute?

jessieay:
This is pretty neat! I just noticed [this comment](https://github.com/thoughtbot/suspenders/issues/827#issuecomment-282371265) from @mike-burns 

> Our hope (my hope? someone's hope) is to move suspenders away from being a command-line tool and toward being a set of rake tasks.

I wonder if this update is a good step in that direction, since it makes Suspenders more customizable? cc @croaky

mike-burns:
@monfresh in #511 we discussed the customizability issue of suspenders and picked a different direction: change suspenders away from being a command-line tool and into being a series of generators. Things like #767 and #676 are steps in that direction.

So while I'm not opposed to this proposed change, I therefore must also ask: what specific customization were you tackling here?

monfresh:
@mike-burns Thanks for the links to those issues. I'll check them out.

To answer your question, I was trying to generate my own ideal Rails app with suspenders, using the suspenders DSL, as well as be able to publish this "template" for others to use. For example, I might modify `suspenders_customization` such that it only runs certain things, or perhaps my own custom methods:

```ruby
def suspenders_customization
  invoke :customize_gemfile
  invoke :setup_development_environment
  invoke :setup_test_environment
  invoke :setup_production_environment
  invoke :setup_secret_token
  invoke :create_suspenders_views
  invoke :configure_app
  invoke :generate_default
  invoke :outro
end
```

and I would also provide my own `Gemfile.erb` template (among other templates I would customize) that sets up the gems I want, that differ from the suspenders defaults.

I would then publish these custom generators, app builders and templates files on GitHub, then someone who wanted to create a "monfresh-flavored" Rails app would download the files to their machine, then run suspenders with the `--custom` flag and point it to the directory that contains my templates.

Does that make sense?

toobulkeh:
@monfresh -- we forked suspenders when we had this need.

While I can't speak to the internal benefits of solidifying the internal suspenders DSL, I don't agree that customization is beneficial to suspenders. Suspenders is different from Rails in that it's an opinionated setup of a Rails config.

Opinionated is the opposite of customization.

monfresh:
@mike-burns I like the idea of having various generators. In the meantime though, I think this PR is still valuable. It is a non-breaking change, and doesn't affect any existing users. No one is forced to use the customization option provided in this PR. For those that do want to customize suspenders, they can easily do it with this PR without having to fork the gem and release their own gem.

If you agree, I will add instructions to the Readme and add a test.

toobulkeh:
There's more to evaluating the acceptance of code than "non-breaking change". As soon as it's in master, people (including you) start to rely on it, and then it becomes a breaking change when the strategy changes in the future.

Not speaking for @mike-burns -- but what if you rewrote this "customization" to follow the generator pattern for things that suspenders does not do? It would be more likely to be accepted then.

monfresh:
@toobulkeh Those are great points. I'm going to close this for now and explore the generators some more.

PR #825, opened 2017-02-17
Update Neat to 2.0.0

whmii:
😍

### DIFF
--- a/lib/suspenders/generators/stylesheet_base_generator.rb
+++ b/lib/suspenders/generators/stylesheet_base_generator.rb
@@ -8,7 +8,7 @@ module Suspenders
 
     def add_stylesheet_gems
       gem "bourbon", "~> 5.0.0.beta.7"
-      gem "neat", "~> 2.0.0.beta.1"
+      gem "neat", "~> 2.0.0"
       gem "refills", group: [:development, :test]
       Bundler.with_clean_env { run "bundle install" }
     end


tute:
So happy! ❤️ 

🚢

croaky:
Congrats! Looks good.

lib/suspenders/generators/stylesheet_base_generator.rb
```diff
@@ -8,7 +8,7 @@ class StylesheetBaseGenerator < Rails::Generators::Base
 
     def add_stylesheet_gems
       gem "bourbon", "~> 5.0.0.beta.7"
-      gem "neat", "~> 2.0.0.beta.1"
+      gem "neat", "~> 2.0.0"
```
tysongach: Should we remove the patch number here, so we that we always get minor release updates? Specifying the patch with a twiddle-wakka equates to `['>= 2.0.0', '< 2.1.0']` (i.e. patch releases only, but not `2.1` for example)? By dropping the patch, it will equate to `['>= 2.0', '< 3.0.0']`.

See: http://guides.rubygems.org/patterns/#pessimistic-version-constraint

croaky: > Use a pessimistic version in the Gemfile for gems that follow semantic versioning, such as rspec, factory_girl, and capybara.
> Use a versionless Gemfile declarations for gems that are safe to update often, such as pg, thin, and debugger.
> Use an exact version in the Gemfile for fragile gems, such as Rails.

https://github.com/thoughtbot/guides/tree/master/best-practices#bundler

tysongach:
@whmii Any update here?

tute:
I'd love this to get merged. Don't forget, @whmii! :)

tysongach:
@whmii 👋

PR #826, opened 2017-02-18
Schedule Heroku Backups and Capture backup of existing staging database for Review Apps

toobulkeh:
When a PR Review environment is created, the latest backup is currently used. If no backups exist, or if the backup is old and outdated, recent data used for testing may not exist.

This forcibly creates a backup right before migration -- leading to the most recent data.

Another way to solve this would be to automatically add `pg:backups:schedule` on all staging environments on original app scaffolding, but that would be up to 24 hours difference.

### DIFF
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -37,6 +37,15 @@ module Suspenders
         end
       end
 
+      def set_heroku_backup_schedule
+        %w(staging production).each do |environment|
+          run_toolbelt_command(
+            "pg:backups:schedule DATABASE_URL --at '02:00 America/Los_Angeles'",
+            environment,
+          )
+        end
+      end
+      
       def create_review_apps_setup_script
         app_builder.template(
           "bin_setup_review_app.erb",
--- a/lib/suspenders/adapters/heroku.rb
+++ b/lib/suspenders/adapters/heroku.rb
@@ -40,7 +40,7 @@ module Suspenders
       def set_heroku_backup_schedule
         %w(staging production).each do |environment|
           run_toolbelt_command(
-            "pg:backups:schedule DATABASE_URL --at '02:00 America/Los_Angeles'",
+            "pg:backups:schedule DATABASE_URL --at '10:00 UTC'",
             environment,
           )
         end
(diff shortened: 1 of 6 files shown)


tute:
> Another way to solve this would be to automatically add `pg:backups:schedule` on all staging environments on original app scaffolding, but that would be up to 24 hours difference.

This is what I normally do on my Rails projects. Regardless of how we merge this PR, I'd like production and staging to have daily backups by default (I thought this was already the case).

I don't feel strongly either way, and we can merge as is or instead turn this PR into adding the daily backup instruction on app creation, and then keeping this script as is. What do you prefer?

Thanks for sharing your work!

toobulkeh:
both

tute:
This LGTM. Calling in @iwz for another review before I merge. Thank you! :)

lib/suspenders/adapters/heroku.rb
```diff
@@ -37,6 +37,15 @@ def set_heroku_rails_secrets
         end
       end
 
+      def set_heroku_backup_schedule
+        %w(staging production).each do |environment|
+          run_toolbelt_command(
+            "pg:backups:schedule DATABASE_URL --at '02:00 America/Los_Angeles'",
```
iwz: Perhaps convert this to UTC?

toobulkeh: It's the Heroku default recommended in their docs -- to convert it do you mean 2AM UTC or 10AM UTC?

iwz: Yeah, I saw that in their docs. 10AM UTC would be fine.

iwz:
LGTM. I'm not sure what others think about the hound line wrapping, though...

@tute @croaky @jferris

iwz:
@toobulkeh this looks good to go. Do you want to clean up any commits before we merge? Also, do you think the README should mention anything about the scheduled backups?

toobulkeh:
Added to the README @iwz. I'd appreciate a Squash and Merge. Is that something I can do?

iwz:
@toobulkeh done. thanks for implementing this!

croaky:
Ooh ooh ooh, good one.



Shall we bump the gem version?

iwz:
@croaky good idea. i'm not set up for that atm, are you?

PR #832, opened 2017-03-10
Fix fatal git failures in tests

mike-burns:
The tests had started printing git error messages:

```
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
```

These were coming from three places:

- `suspenders.gemspec`, which collects all `Gem::Specification#files=`
  by shelling out to `git ls-files`.
- `suspenders.gemspec`, which collects all
  `Gem::Specification#test_files=` by shelling out to
  `git ls-files -- {test,spec,features}/*`.
- Honeybadger, which figures out which revision to tag the error with by
  invoking `rev-parse`:

      def from_git
        return nil unless File.directory?('.git')
        `git rev-parse HEAD`.strip rescue nil
      end

Although I am unsure why these things started to matter now, I found two
changes that improve the situation:

- In the test suite, make an initial commit after suspender'ing a
  project.
- When suspender'ing a project, make the git repo last.

### DIFF
--- a/spec/support/suspenders.rb
+++ b/spec/support/suspenders.rb
@@ -17,6 +17,12 @@ module SuspendersTestHelpers
         `
           #{suspenders_bin} #{APP_NAME} #{arguments}
         `
+        Dir.chdir(APP_NAME) do
+          with_env("HOME", tmp_path) do
+            `git add .`
+            `git commit -m 'Initial commit'`
+          end
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


spec/support/suspenders.rb
```diff
@@ -17,6 +17,12 @@ def run_suspenders(arguments = nil)
         `
           #{suspenders_bin} #{APP_NAME} #{arguments}
         `
+        Dir.chdir(APP_NAME) do
+          with_env("HOME", tmp_path) do
```
mike-burns: This file is all over the double quotes.

PR #837, opened 2017-04-25
.git/safe is opt-in

mike-burns:
The idea behind the .git/safe is that I run it manually and explicitly
after I trust the directoy and the team behind the project.

Creating it as part of the setup script defeats the point of that. Doing
so without explcitly telling the user comes across as sneaky, perhaps.
(I do expect all devs to read setup scripts before running them -- but I
also know that they do not!)

Switch this script to inform the user that they have no `.git/safe`
directory but might like to make one.

### DIFF
--- a/templates/bin_setup
+++ b/templates/bin_setup
@@ -13,8 +13,12 @@ bundle check || bundle install
 # Set up database and add any development seed data
 bin/rails dev:prime
 
-# Add binstubs to PATH via export PATH=".git/safe/../../bin:$PATH" in ~/.zshenv
-mkdir -p .git/safe
+if [ ! -d .git/safe ] && echo $PATH | grep .git/safe > /dev/null; then
+  echo "-----------------------------------------------------------------------"
+  echo
+  echo "-> When you trust this repo, remember to run: mkdir -p .git/safe"
+  echo
+fi
 
 # Only if this isn't CI
 # if [ -z "$CI" ]; then
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 `master`
 
+* No longer generate `mkdir -p .git/safe` as part of `bin/setup`
 * Update Bitters to 1.7
 * Update Neat to 2.1
 * Update Bourbon to 5.0.0.beta.8


mike-burns:
This can be considered WIP: I don't have the time/bandwidth to test this right now, but it seems to have the right syntax.

mike-burns:
Confirmed this, verified the output, and it looks great. Truly a work of art.

Bikeshedding over the output is welcome, but this is now good to review.

templates/bin_setup
```diff
@@ -13,8 +13,13 @@ bundle check || bundle install
 # Set up database and add any development seed data
 bin/rake dev:prime
 
-# Add binstubs to PATH via export PATH=".git/safe/../../bin:$PATH" in ~/.zshenv
-mkdir -p .git/safe
+if [ ! -d .git/safe ] && echo $PATH | grep .git/safe > /dev/null; then
+  # Add binstubs to PATH via export PATH=".git/safe/../../bin:$PATH" in ~/.zshenv
```
georgebrock: I'm not sure this comment makes sense anymore.

croaky:
A short blog post specifically about .git/safe might be a helpful explanation for many https://www.google.com/search?client=safari&hl=en-us&ei=nIEDWZy-LMWojwPKuLuwDg&q=%22.git%2Fsafe%22+site%3Arobots.thoughtbot.com

templates/bin_setup
```diff
@@ -13,8 +13,12 @@ bundle check || bundle install
 # Set up database and add any development seed data
 bin/rails dev:prime
 
-# Add binstubs to PATH via export PATH=".git/safe/../../bin:$PATH" in ~/.zshenv
-mkdir -p .git/safe
+if [ ! -d .git/safe ] && echo $PATH | grep .git/safe > /dev/null; then
+  echo "-----------------------------------------------------------------------"
+  echo
+  echo "-> When you trust this repo, remember to run: mkdir -p .git/safe"
```
croaky: Looks like we lost part of the help text from the previous comment?

```
Add binstubs to PATH via export PATH=".git/safe/../../bin:$PATH" in ~/.zshenv
```

Would be nice to drop the Z shell-specific part of that instruction.

The instruction "When you trust this repo, remember to run: mkdir -p .git/safe" doesn't fully accomplish the goal of using the project's binstubs unless they've also configured their `PATH`, though?

Worth adding to this instruction something like this?

> Add this project's binstubs to your PATH via export PATH=".git/safe/../../bin:$PATH"

Or we write the be-all, end-all blog post about this whole topic and then link to that blog post from this instruction?

mike-burns: The check on the line above makes sure to only show this message to people who have configured their PATH. That way we only show this to those of us who make use of such a feature.

PR #847, opened 2017-06-25
Update Rails to 5.1

delphaber:
Update Rails to 5.1

(adds support for webpack too)


delphaber:
Ouch... hound is barking out loud!

delphaber:
I didn't change `config/puma.rb`, but I noticed that in Rails 5.1 they changed it a little (see railsdiff http://railsdiff.org/5.0.4/5.1.1 ).

Heroku guide on puma (https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server) has not been updated since 2016 though.

So I don't know if `config/puma.rb` needs an update, because I'm not a big expert of puma on heroku :D. Any idea?

tute:
The puma diff has only comment changes, should be fine as is.

cc @nickcharlton for review, please! Thank you.

delphaber:
I noticed that rails 5.1 now:

`Initialize Git repo when generating new app, if option --skip-git is not provided. (Pull Request)`

So I think I can remove the `init_git` step in suspenders. Will work on it this evening (CEST time)

delphaber:
Ok I did it. The git repository is now created by Rails. `--skip-git` still exists in Rails.

delphaber:
Have you changed hound settings? :D I'm pretty sure I had to use double quotes for strings before!

nickcharlton:
@delphaber I _think_ that the organisation config got lost on Hound and had to be reset. If you pushed again it'll rerun and be happy.

delphaber:
@nickcharlton I pushed an empty commit, but those hound comments did not disappear from the PR. I think we can ignore them anyway, since hound check is ok!

nickcharlton:
@delphaber Thanks! I went through, confirmed by eye and then deleted the invalid comments.

Going to merge now!

PR #851, opened 2017-07-07
Install normalize.css via yarn

delphaber:


nickcharlton:
Great, thanks! Merging away.

tysongach:
Just curious: what’s the advantage of installing via Yarn instead of Rails Assets?

nickcharlton:
Mainly, I feel, that we're doing things the direction Rails is going and the growth of other `npm` related dependencies which are (presumably) likely on projects.

Issue #781, opened 2016-06-21
Auto-Generate & Deploy SSL for staging and production


toobulkeh:
With the [advent of Free SSL endpoints on Heroku](https://blog.heroku.com/archives/2016/5/18/announcing_heroku_free_ssl_beta_and_flexible_dyno_hours) and [Free SSL certificates using LetsEncrypt](https://letsencrypt.org/), Suspenders would be the ideal place to automatically configure and setup SSL certificates as part of all new rails applications -- even for both staging and production environments.

So far, we've configured applications thusly:
1. Deployed [Sabayon](https://github.com/dmathieu/sabayon) applications (a small Go app that gains Heroku permissions) for each environment app. We've done this using the naming scheme `sabayon-<app>-staging` and `sabayon-<app>-production`
2. Added the following config to rails' `config/routes.rb`:
   
   > ``` ruby
   > Rails.application.routes.draw do
   >   # Let's Encrypt for Sabayon (Heroku)
   >   if ENV['ACME_KEY'] && ENV['ACME_TOKEN']
   >     get ".well-known/acme-challenge/#{ ENV["ACME_TOKEN"] }" => proc { [200, {}, [ ENV["ACME_KEY"] ] ] }
   >   else
   >     ENV.each do |var, _|
   >       next unless var.start_with?("ACME_TOKEN_")
   >       number = var.sub(/ACME_TOKEN_/, '')
   >       get ".well-known/acme-challenge/#{ ENV["ACME_TOKEN_#{number}"] }" => proc { [200, {}, [ ENV["ACME_KEY_#{number}"] ] ] }
   >     end
   >   end
   > ...
   > ```
3. Forced SSL in Staging and changed ASSET_HOST to be protocol specific.  
4. Setup `bin/setup` to properly check for the sabayon apps.

Open questions we still have:
1. What is the best user/account strategy with Heroku to use the API KEY for Sabayon. We've created separate accounts for each client and have silo'd access using those keys -- which are not the same key. Heroku cloud now supports teams (like Enterprise) though there is no free tier of this.
2. We need help automating the deployment of the Sabayon apps. This includes the app names, keys, timeout periods, and domains.
3. Is this a separate generator / flag, since it includes information not previously required, or info that may be optional (like domain names)?

Hopefully this is close-enough to discuss. If it seems worthwhile, I'd be happy to contribute to the PR.

🍻 to an automated, secure, web.

tute:
[...]

PR #855, opened 2017-07-25
Enforce SSL in production environment

croaky:
This setting has been on for every thoughtbot Rails app I've touched
in the past year or two (maybe a dozen?).

* Use `configure_environment` helper method more consistently.
* Fixes https://github.com/thoughtbot/suspenders/issues/781

More about the pervasiveness of SSL in our ecosystem:

https://blog.heroku.com/announcing_heroku_free_ssl_beta_and_flexible_dyno_hours
https://letsencrypt.org/

### DIFF
--- a/lib/suspenders.rb
+++ b/lib/suspenders.rb
@@ -1,5 +1,6 @@
 require 'suspenders/version'
 require 'suspenders/generators/app_generator'
+require "suspenders/generators/enforce_ssl_generator"
 require 'suspenders/generators/static_generator'
 require 'suspenders/generators/stylesheet_base_generator'
 require 'suspenders/actions'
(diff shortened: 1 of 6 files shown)


mike-burns:
1. Yes, please.
2. Can this be a separate generator instead? See #767 for an example, but in short: new file under `lib/suspenders/generators`, modify `#generate_default` in `lib/suspenders/generators/app_generator.rb`, require it in `lib/suspenders.rb`.

croaky:
> Can this be a separate generator instead?

@mike-burns Like a `suspenders/generators/production_config_generator` that groups all the changes to the `config/environments/production.rb` file?

mike-burns:
I'd be OK with that, but I'm thinking more fine-grained: a `suspenders/generators/enforce_ssl_generator` that only makes this change.

The problem this is trying to solve is: people want suspenders to be more configurable, and want it to be easier to undo parts of suspenders. The long-term goal is to move each concept into its own generator, with one generator to rule them all.

croaky:
@mike-burns Ready for another review

PR #856, opened 2017-08-01
Prevent memory bloat in ActiveJob children

croaky:
`ActiveJob` attaches a single `LogSubscriber` to be notified of events.
When one job enqueues another, that's treated as a child event and
stored on the same stack as the parent. If a large number of children
are enqueued that leads to R14 errors on Heroku.

The fix here is to unsubscribe all existing "enqueue" listeners and
then add a new subscriber for just that event. The original handler
is copied from the parent class using `define_method` so that it's
picked up by `attach_to`.

Related:

https://github.com/rails/rails/issues/21036
https://github.com/rails/rails/pull/5932

Before and after:

Heroku metrics chart showing 1GB drop in memory usage after this change deployed

🎩  tip @trevororeilly

### DIFF
--- a/lib/suspenders.rb
+++ b/lib/suspenders.rb
@@ -1,6 +1,7 @@
 require 'suspenders/version'
 require 'suspenders/generators/app_generator'
 require "suspenders/generators/enforce_ssl_generator"
+require "suspenders/generators/initialize_active_job_generator"
 require 'suspenders/generators/static_generator'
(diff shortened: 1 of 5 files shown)


nickcharlton:
That's an impressive memory improvement!

The implementation looks good to me, but others might have some thoughts.

croaky:
The example chart obviously specific to our app, which had a lot of jobs kicking off child jobs as part of an ETL process.

PR #853, opened 2017-07-10
Configures action mailer asset host

delphaber:

Issue #843, opened 2017-06-01
Set ActionMailer's asset_host


jakecraige:
We currently set the maiiler's `default_url_options` but not the `asset_host`. Because of this when you use the `_url` helpers in mailers it doesn't include the URL which is unexpected. 

We should probably update the config to also set `action_mailer.asset_host` to the environment's asset host which we already have.

Thoughts?

nickcharlton:
We should do that. I'd happily merge such a PR!

delphaber:
https://github.com/thoughtbot/suspenders/pull/853 !

delphaber:
I've noticed I introduced a bug with that PR. Plus, we have to consider asset_host protocol. I'll open a new PR as soon as possibile

purinkle:
@delphaber, is the original issue now dealt with by #853? Can we close this issue and open a new one for the bug that was introduced?

delphaber:
@purinkle yes, it was introduced by #853. I already created a PR with the fix ( #858 ). But I have a question to ask you all. It's all in there :)

PR #858, opened 2017-08-04
Fix for action mailer asset_host: it should be a string, not an hash

delphaber:
This PR fixes a problem introduced in https://github.com/thoughtbot/suspenders/pull/853 and regarding https://github.com/thoughtbot/suspenders/issues/843.

`config.action_mailer.asset_host` accepts a plain string, not an hash like its cousin `config.action_mailer.default_url_options` does.

Plus, I had to add the protocol (`http` for `development` and `test`). Rails would not add the protocol for you and so if you have a logo in your email its url would be something like `//localhost:3000/assets/logo.png`. If you open the email in your native mail client, for instance, it won't load the image. This is why I had to specificy `http:` for `asset_host`.

How to manage this protocol problem in production? In my last application, I defined an `ASSET_HOST` env variable to be like my `APPLICATION_HOST` env variable with the protocol at the beginning. So `ASSET_HOST` is now a mandatory env variable for my project, and I think it's not what `suspenders` wants. Any idea? Shall I presume every assets hosts respond to https, like so?:

`config.action_mailer.asset_host = "https://#{ENV.fetch('ASSET_HOST', ENV.fetch('APPLICATION_HOST'))}"`

### DIFF
--- a/lib/suspenders/actions.rb
+++ b/lib/suspenders/actions.rb
@@ -15,7 +15,7 @@ module Suspenders
     end
 
     def action_mailer_asset_host(rails_env, host)
-      config = "config.action_mailer.asset_host = { host: #{host} }"
+      config = "config.action_mailer.asset_host = #{host}"
       configure_environment(rails_env, config)
     end
 
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -164,7 +164,7 @@ RSpec.describe "Suspend a new project with default configuration" do
   it "sets action mailer default host and asset host" do
     config_key = 'config\.action_mailer\.asset_host'
     config_value =
-      '{ host: ENV\.fetch\("ASSET_HOST", ENV\.fetch\("APPLICATION_HOST"\)\) }'
+      %q{ENV\.fetch\("ASSET_HOST", ENV\.fetch\("APPLICATION_HOST"\)\)}
     expect(production_config).to match(/#{config_key} = #{config_value}/)
   end
 
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -299,9 +299,9 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
 
     def configure_action_mailer
       action_mailer_host "development", %{"localhost:3000"}
-      action_mailer_asset_host "development", %{"localhost:3000"}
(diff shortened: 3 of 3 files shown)


spec/features/new_project_spec.rb
```diff
@@ -164,7 +164,7 @@
   it "sets action mailer default host and asset host" do
     config_key = 'config\.action_mailer\.asset_host'
     config_value =
-      '{ host: ENV\.fetch\("ASSET_HOST", ENV\.fetch\("APPLICATION_HOST"\)\) }'
+      'ENV\.fetch\("ASSET_HOST", ENV\.fetch\("APPLICATION_HOST"\)\)'
```
croaky:
```
Use %{} for single-line strings needing interpolation and double-quotes.
```

https://github.com/thoughtbot/guides/tree/master/style/ruby#ruby

lib/suspenders/actions.rb
```diff
@@ -15,7 +15,7 @@ def action_mailer_host(rails_env, host)
     end
 
     def action_mailer_asset_host(rails_env, host)
-      config = "config.action_mailer.asset_host = { host: #{host} }"
+      config = "config.action_mailer.asset_host = #{host}"
```
croaky: Does the `host` here pass quotes around itself? I think I'd expect something like:

```ruby
config = %{config.action_mailer.asset_host = "#{host}"}
```

delphaber: It could have quotes, but it has not for the production environment.
The `%{}` has been used in https://github.com/thoughtbot/suspenders/blob/master/lib/suspenders/app_builder.rb#L301 so I think we are good here

spec/features/new_project_spec.rb
delphaber: Will do!

lib/suspenders/actions.rb
croaky: Ah, makes sense

croaky:
Thinking this would be a good one to get in before the new Suspenders release.

spec/features/new_project_spec.rb
delphaber: Had to use `%q{}` instead of `%{}`, otherwise regexp won't match

delphaber:
Can I keep the protocol change for at least development and test environments? Without it images in emails won't work in development.

I will leave production untouched (no protocol assumption)

croaky:
Yup sounds good

Issue #834, opened 2017-03-27
Move bin/setup_review_app to app.json for auto-setup


toobulkeh:
Here's a [strategy](http://stackoverflow.com/a/36784098/1376201) for copying databases using `DATABASE_URL` env variable instead of requiring the Heroku API:

```js
// app.json
"scripts": {
  "postdeploy": "pg_dump $STAGING_DATABASE_URL | psql $DATABASE_URL && bundle exec rake db:migrate"
},
```

The downside of this is that it can cause problems if people use `app.json` for promoting apps to production from staging.

This would get rid of the step of typing `bin/setup_review_app ID` after every new PR.

toobulkeh:
We might want to keep `bin/setup_review_app` depending on how people use it -- as [subsequent deploys don't](https://devcenter.heroku.com/articles/app-json-schema#scripts) run the `postdeploy` script from `app.json`

toobulkeh:
@tute any thoughts on having this script only run on first deploy of PR env?

tute:
That'd be killer. We shouldn't need to remember to manually run this script after each new PR.

But also, we maybe shouldn't be reusing the staging database, and use development seeds instead, per best practices and Heroku recommendation. That's probably another issue though.

toobulkeh:
glad you like the concept! Just double-checking though -- is it a major problem or minor annoyance to not have this script run on every new deploy to an existing PR (branch and env)?

👍  on the concept of dev:prime > staging db. We try to practice that as well.

tute:
It's a problem because otherwise there's no database. If you push the branch, and an owner runs acceptance, the app will fail with a "users table not found", or the like.

toobulkeh:
Well -- this `app.json` script will run the first time the PR is created -- creating user tables.
it _won't_ run the second time (ie. someone pushes a fix to the same branch that the PR is based on.)

Sorry if I'm explaining something you already understand -- the language you're using just makes me think we're miscommunicating. 

For example, a scenario:

1. master branch is auto-deploying to the staging heroku environment
2. joe creates a new branch `phones` that adds a phone number column to the users table
3. he opens a PR (lets say 7) for `phones`
3. heroku automatically creates `staging-pr-7`
3. heroku automatically runs the `dev:prime` script and migrates the db, all via `app.json`
3. `staging-pr-7` is good to go and reflects phone numbers on users
3. joe's feature requirements expand to include a cell phone and a work phone number
3. joe adds a second phone number and features to support them
[...]

PR #859, opened 2017-08-05
Auto setup PR environments

toobulkeh:
In replace of #834, I've setup the work required to:
1. Remove requirement of running `bin/setup_review_app` manually on a new PR env
2. Run `rake db:migrate` on subsequent updates to the PR branch

I've manually tested this with Heroku and several scenarios, and it works well.

A couple of notes:
1. I'm not sure why Heroku's release stage doesn't happen on subsequent deploys to PR envs (why rake db:migrate doesn't run automatically) per [their documentation](https://devcenter.heroku.com/articles/release-phase#when-does-the-release-command-run). This seems like a product bug.
2. This replaces the current strategy of copying staging backups to staging-pr-### and runs `rake dev:prime` instead.
3. The build logs [show errors](https://kb.heroku.com/why-am-i-seeing-user-does-not-have-connect-privilege-error-with-heroku-postgres-on-review-apps) on Heroku because `rake dev:prime` first runs `rake db:setup` which first runs `rake db:create` which fails on Heroku -- but the rest of the build completes just fine.

I'll need to add some automated tests for these changes, but I wanted to get the general solution out for review and comment today.

### DIFF
--- a/lib/suspenders/actions.rb
+++ b/lib/suspenders/actions.rb
@@ -15,7 +15,7 @@ module Suspenders
     end
 
     def action_mailer_asset_host(rails_env, host)
-      config = "config.action_mailer.asset_host = { host: #{host} }"
+      config = "config.action_mailer.asset_host = #{host}"
       configure_environment(rails_env, config)
     end
 
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -302,9 +302,9 @@ Rack::Timeout.timeout = (ENV["RACK_TIMEOUT"] || 10).to_i
 
     def configure_action_mailer
       action_mailer_host "development", %{"localhost:3000"}
-      action_mailer_asset_host "development", %{"localhost:3000"}
+      action_mailer_asset_host "development", %{"http://localhost:3000"}
       action_mailer_host "test", %{"www.example.com"}
-      action_mailer_asset_host "test", %{"www.example.com"}
+      action_mailer_asset_host "test", %{"http://www.example.com"}
       action_mailer_host "production", %{ENV.fetch("APPLICATION_HOST")}
       action_mailer_asset_host(
         "production",
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -169,7 +169,7 @@ RSpec.describe "Suspend a new project with default configuration" do
   it "sets action mailer default host and asset host" do
     config_key = 'config\.action_mailer\.asset_host'
     config_value =
-      '{ host: ENV\.fetch\("ASSET_HOST", ENV\.fetch\("APPLICATION_HOST"\)\) }'
+      %q{ENV\.fetch\("ASSET_HOST", ENV\.fetch\("APPLICATION_HOST"\)\)}
     expect(production_config).to match(/#{config_key} = #{config_value}/)
   end
 


templates/dev.rake
```diff
@@ -1,4 +1,4 @@
-if Rails.env.development? || Rails.env.test?
+unless Rails.env.production?
```
tute: Doesn't suspenders run on production for every Heroku environment?

toobulkeh: I.. hope not?
I see that you [started getting rid of staging](https://github.com/thoughtbot/suspenders/pull/743), but it looks like it [was](https://github.com/thoughtbot/suspenders/blob/8c39efccac0fad8e6b74d25ff3a8809dd2f7df00/templates/secrets.yml#L8) [never](https://github.com/thoughtbot/suspenders/blob/master/templates/Gemfile.erb#L49) [completed](https://github.com/thoughtbot/suspenders/blob/a930e6bf9fbc4fb11bf59fbf637349a0e3733fad/templates/postgresql_database.yml.erb#L21).



We still use staging/production environments -- and staging PR environments are tagged as staging at the ENV level as well.

While I agree that [Heroku made that recommendation](https://devcenter.heroku.com/articles/multiple-environments#managing-staging-and-production-configurations), I don't understand it. It's like taking a config option that was built for environment separation and no longer using it.

toobulkeh: The main reason we haven't migrated is because we use Bugsnag, which automatically parses environments from `RAILS_ENV`. I see that suspenders is still configured to use Honeybadger.. how do you override this in Honeybadger (we had the same issue there when investigating this split).

croaky: For Honeybadger, we set `HONEYBADGER_ENV=staging`

croaky: https://github.com/thoughtbot/suspenders/pull/860 to finish cleaning up staging references

toobulkeh: But... why? We want these environments to be handled differently. We already mapped the entire production config to staging for dev/prod parity.

Now we have a custom configuration throughout the app running complex checks instead of just using the Rails convention that was already designed for this exact purpose.

In particular, for this line, we're going to have to use something nasty like:
```ruby
if Rails.env.development? or Rails.env.test? or ENV.fetch("HEROKU_APP_NAME", "").include?("staging-pr-")
  ...
```

While part of me may be resisting change.. the main part just doesn't understand -- despite the Heroku recommendations.

In this same vein, should we get rid of the `test` environment because it's similar to `development`? Why even have `development`? Why not just run everything in `RAILS_ENV=production`

tute: In my experience, the staging/production split is a source of potential mistakes: code doesn't run any place you test as it does in production. I don't have specifics, but remember unpleasant surprises. The idea makes me uncomfortable.

Makes sense to me to configure different services to different keys, and have the code run exactly the same.

In an app I needed to disable some services in staging because of costs, and I added a line similar to `ENV.fetch("HEROKU_APP_NAME", "").include?("staging-pr-")` to `Rails.application.config`. Can also suggest adding a boolean like `DEMO=true` or `STAGING_ENV=true`, that `HONEYBADGER_ENV` can mimic. This still rings bells in my brain, but it's been useful so far.

Development and test ought to run in different databases, use different mailer configurations, and sometimes fakes. I think it makes sense to configure those independently. The differences between the three `config/environments/*` files describe more examples.

Thanks for your discussion. I've been taking this for granted for too long. :)

toobulkeh: Thanks for the discussion! I'll go ahead and change this PR.

Are you suggesting Staging uses the same DB instance as Production? Or were you referring to different types of databases?

So you're suggesting we add a different ENV variable for distinguishing staging from production, and code that depends on it? Any ideas what to call it? `HEROKU_ENV`?

tute: > Thanks for the discussion! I'll go ahead and change this PR.

Thank you too! :)

> Are you suggesting Staging uses the same DB instance as Production? Or were you referring to different types of databases?

I meant database types. In the servers we can configure the different database instances with `DATABASE_URL`, but we know how risky it is to configure it that way in development. Imagine if we had to set it to development after running tests... 😨 😆 

> So you're suggesting we add a different ENV variable for distinguishing staging from production, and code that depends on it? Any ideas what to call it? HEROKU_ENV?

Another env variable can get confusing indeed. Not sure how to even name it. In a project I added a Rails configuration:

```
--- a/config/application.rb
+++ b/config/application.rb
@@ -51,6 +51,10 @@ module PatientCheckIn

+    config.server_env = ENV.
+      fetch("APTIBLE_APPLICATION_HANDLE", Rails.env).
+      gsub("app-name-", "")
```

Don't love it either but it works.

What about adding many ENV variables? `FEATURE_A_ENABLED`, `HONEYBADGER_ENV`, `BUGSNAG`, etc? And we forget about almost-production Rails environments altogether? They can be properly named, and are easy to configure during tests too.

Still thinking!

toobulkeh: I just personally don't like seeing "production" anywhere it's NOT production. From an Ops perspective, I've been burned by that before.

That said, I'd suggest renaming "dev:prime" to "db:prime" and an env flag of `PRIMEABLE_ENV=true` -- but I'll leave it up to you to decide.

toobulkeh:
Here's a potential downside of using the custom `release` command: Heroku currently sends email notifications to everyone on the team if it fails for some reason (in our case, a developer pushed some build-chain failures that didn't pass CI, but Heroku didn't wait for CI to complete on PR branches).

We're investigating it some more and talking with Heroku to see if there are permissions for these things.
screen shot 2017-08-11 at 1 46 22 pm

toobulkeh:
Here's another weird side effect in Heroku's UI
screen shot 2017-08-23 at 9 40 42 am

purinkle:
@toobulkeh, are you still seeing some weird side-effects with this on Heroku?

toobulkeh:
@purinkle Yup -- but it's honestly worth it. Those things can definitely be improved by Heroku later if we feature-request them.

templates/Gemfile.erb
```diff
@@ -41,12 +41,19 @@ group :development, :test do
   gem "bullet"
   gem "bundler-audit", ">= 0.5.0", require: false
   gem "dotenv-rails"
-  gem "factory_girl_rails"
   gem "pry-byebug"
   gem "pry-rails"
   gem "rspec-rails", "~> 3.6"
 end
 
+group :development, :test, :staging do
```
purinkle: Looking at how Suspenders is set up out of the box, the Staging environment does not exist anymore. The Staging application runs under the Production environment. Is there a way you can think of that does not need us to include these gems in Production?

toobulkeh: Suspenders uses FactoryGirl by default for `dev:prime` (renamed to `data:prime` here).
We could skip that requirement... OR we could have a conditional install based on the existence of the ENV variable... yet again another reason to not do ENV level parity between staging and production :)

templates/Gemfile.erb
```diff
@@ -66,5 +59,8 @@ group :test do
 end
 
 group :production do
+  if ENV["PRIMEABLE"]
+    gem "factory_girl_rails"
```
toobulkeh: @purinkle what do you think about this?

toobulkeh:
Official Heroku response about `release` dyno showing up on Dashboard.
screen shot 2017-09-15 at 2 41 33 pm

purinkle:
@toobulkeh, have you thought about setting up a `review` environment that inherits from `production` instead of setting a `PRIMEABLE` environment variable?

toobulkeh:
That's a cool concept -- I have not. I can consider it.
I assumed that with the focus on staging/production deduplication recently, a `review` environment would be a move in the opposite direction.

purinkle:
The argument would be that `staging` and `production` should be _exactly_ the same. Under this circumstance – where we are using the environment to control the configuration – I would argue that adding a new environment configuration is an acceptable solution.

PR #860, opened 2017-08-08
Remove vestigial `staging` references

croaky:
The staging environment is set to production as of
ffa20dbeec382d6a78704be15881d5dd1248d9df in order to:

* Simplify configuration
* Enforce staging/production parity
* Follow Heroku recommendations
  https://devcenter.heroku.com/articles/multiple-environments

For Honeybadger, we set a config variable `HONEYBADGER_ENV=staging`.

Related:

https://github.com/thoughtbot/suspenders/pull/743

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,8 +127,6 @@ This:
 
 * Creates a staging and production Heroku app
 * Sets them as `staging` and `production` Git remotes
-* Configures staging with `RACK_ENV` environment variable set
-  to `staging`
 * Creates a [Heroku Pipeline] for review apps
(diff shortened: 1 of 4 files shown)


toobulkeh:
I'd suggest automating `HONEYBADGER_ENV=staging` on staging env setup. And add it to `app.json` for staging-pr envs.

croaky:
@toobulkeh Good idea, I can do that in a follow-up PR

tute:
Looks good to me.

I'd also remind about automating `HONEYBADGER_ENV` for staging and staging-pr apps, but agree it can be treated as a separate bug.

PR #864, opened 2017-10-03
Remove Database Cleaner

derekprior:
Rails 5.1 shares the existing database connection with the test thread,
which means we can rely on transactions once more. Huzzah. See [Capybara
README][1].

The Circle CI template was also updated to expoicitly run `bin/setup` in
the `development` environment. This allows us to test any development or
database seeds `bin/setup` might end up adding to the database without
polluting the test database that will be used when specs are run later
in the build.

[1]: https://github.com/teamcapybara/capybara#transactions-and-database-setup


derekprior:
That's a 1.0 config. I haven't wrapped my head around 2.0 yet. I think we should open an issue for that.

croaky:
👍  https://github.com/thoughtbot/suspenders/issues/865

PR #873, opened 2017-12-29
Document Bundler::RubyVersionMismatch

mike-burns:
This comes up far too often when popping in to work on this project, so
document what to do when it happens.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,3 +52,11 @@ We may suggest changes.
 
 To update the Ruby version,
 change `.ruby-version` and `.travis.yml`.
+
+If you see this error while running tests:
+
+```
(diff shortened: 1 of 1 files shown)


derekprior:
Is there value in keeping `tmp/` around ever? Should we delete it in `bin/setup` and perhaps also in a `before(:suite)` hook in RSpec?

mike-burns:
There was a time in history when The Thing To Do was to write exactly one `expect` per `it`. That lead to a test suite with a `before(:all)` that generates a suspenders project and quite a few `it` blocks that test against it. So that's why we can't delete it in a `before(:each)`.

Deleting it in a `before(:all)` makes sense but is also something we already do: https://github.com/thoughtbot/suspenders/blob/master/spec/features/new_project_spec.rb#L6 . This project is so difficult to debug that I don't understand why that doesn't work when changing Ruby versions.

PR #882, opened 2018-01-13
Introduce JavaScript driver generator

mike-burns:
This introduces `suspenders:js_driver`, a generator for installing our
JavaScript test runner. This one is capybara-webkit.

---

While here, remove the `configure_ci` build step that was undefined in a
prior commit.


Ru-T:
Looks good! 🚢

lib/suspenders/generators/js_driver_generator.rb
```diff
@@ -0,0 +1,20 @@
...
+  class JsDriverGenerator < Rails::Generators::Base
+    source_root File.expand_path(
+      File.join("..", "..", "..", "templates"),
+      File.dirname(__FILE__),
+    )
+
+    def add_gems
+      gem "capybara-webkit", group: :test
+
+      Bundler.with_clean_env { run "bundle install" }
```
mike-burns: Don't need to run this here.

PR #883, opened 2018-01-13
Introduce a testing generator

mike-burns:
This introduces `suspenders:testing`, a generator that installs our
testing infrastructure. Specifically, rspec and shoulda-matchers.

The `#empty_directory_with_keep_file` method is defined in
`Rails::Generators::AppBase`, but that is too heavy of a dependency;
instead, define the method manually.


lib/suspenders/generators/app_generator.rb
```diff
@@ -95,14 +95,8 @@ def setup_development_environment
 
     def setup_test_environment
       say 'Setting up the test environment'
-      build :generate_rspec
-      build :configure_rspec
       build :configure_background_jobs_for_rspec
```
Ru-T: Out of curiosity, why not move these into the generator as well? Will they be part of a future move?

mike-burns: I was thinking that the `configure_background_jobs_for_rspec` would go into a `suspenders:jobs` generator. `configure_ci` is in `suspenders:ci` and `configure_capybara_webkit` is in `suspenders:js_driver`.

Ru-T: Sounds right to me!

PR #884, opened 2018-01-13
Introduce views generator

mike-burns:
This introduces the `suspenders:views` generator, responsible for
populating the Rails app with some useful base templates.

One of the templates brings in the `analytics` partial, which does not
exist until Segment is installed. When that is extracted we should make
sure to handle this.


Ru-T:
Not sure why `hound`is stuck - might be worth getting it to run again. But apart from that, 🚢

PR #885, opened 2018-01-13
Introduce analytics generator

mike-burns:
This introduces `suspenders:analytics`, a generator for installing app
analytics tracking. We use Segment.

This modifies an existing template to conditionalize the rendering of
the analytics partial. If the `suspenders:views` generator was not run,
do not render the partial, but do copy the partial so that the developer
can do what they please with it.


lib/suspenders/generators/analytics_generator.rb
```diff
@@ -0,0 +1,29 @@
...
+      File.dirname(__FILE__),
+    )
+
+    def install_partial
+      copy_file "_analytics.html.erb",
+        "app/views/application/_analytics.html.erb"
+    end
+
+    def render_partial
+      if File.exist?(js_partial)
```
Ru-T: Curious when this view would not exist. In the prior PR, it was always copied.

```
def create_shared_javascripts
  copy_file "_javascript.html.erb",
    "app/views/application/_javascript.html.erb"
end
```

Is there some conditional I'm missing?

mike-burns: Now that this is an independent task I tried to consider what it'd be like if it ran in isolation from the other tasks. If the `suspenders:views` generator isn't run, the `_analytics.html.erb` template won't exist and this will error.

Ru-T: Makes sense!

PR #887, opened 2018-01-18
Move config of background jobs for rspec to jobs generator.

Ru-T:
This PR introduces `jobs_generator`, a generator that runs a `delayed_job` generation task, in this case specifically for rspec.

It also combines the `initialize_active_job_generator` into this new one


Ru-T:
@mike-burns, yes! Moved gem installation to the generator. Check it out now? 🙏

mike-burns:
So on a flight home I did some looking around. Something seems off in a previous decision.

38b530c0 added and `initialize_active_job` generator. I _think_ it can work separately from the rest of this ... but I could be trivially persuaded that it should be renamed to `suspenders:jobs`. In fact, I might already be persuaded: let's rename `initialize_active_job` to `jobs`.

And then there's the `configure_active_job` function; it sets the app to use `delayed_job`. Anything that adds the `delayed_job` gem should probably also run that `configure_active_job` functionality.

So this could go two ways:

- Rename the existing `initialize_active_job` generator to `jobs`, include the testing setup you have already done here, and the `gem` bit that you added, and also pull in the `configure_active_job` functionality.
- Split the jobs concepts into `jobs` and `job_tests` generators, each of which does the `gem` bit conditionally (like, only if it's not already in the `Gemfile`).

I am unconvinced that `job_tests` can make sense without `jobs`, though. But maybe?

Anyway, what do you think?

Ru-T:
@mike-burns - I agree that they should be one generator. 

Please take a look at this updated PR. I included the other method from `app_builder` that handled `active_job` as well.

Ru-T:
@mike-burns Great. Any reason not to merge? Asking because there are a few open PR's of this nature that have been recently approved.

mike-burns:
No reason not to merge, just haven't had time. Please merge any PR that seems ready!

PR #895, opened 2018-03-02
Introduce an email generator

mike-burns:
This introduces `suspenders:production:email`, a generator that installs
our email settings for production. In this case it sets up SMTP with
variables based on environment variables.

I separated out the RecipientInterceptor configuration, since that is a
development- and testing-related configuration.

### DIFF
--- a/lib/suspenders/generators/production/email_generator.rb
+++ b/lib/suspenders/generators/production/email_generator.rb
@@ -1,0 +1,30 @@
+require "rails/generators"
+
+module Suspenders
+  module Production
+    class EmailGenerator < Rails::Generators::Base
+      source_root File.expand_path(
+        File.join("..", "..", "..", "templates"),
+        File.dirname(__FILE__),
+      )
+
+      def smtp_configuration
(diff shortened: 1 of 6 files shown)


lib/suspenders/generators/production/email_generator.rb
```diff
@@ -0,0 +1,30 @@
...
+        prepend_file "config/environments/production.rb",
+          %{require Rails.root.join("config/smtp")\n\n}
+      end
+
+      def use_smtp
+        config = <<-RUBY
+
+  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.smtp_settings = SMTP_SETTINGS
+        RUBY
```
nickcharlton: Would we be able to use a [squiggly heredoc](https://infinum.co/the-capsized-eight/multiline-strings-ruby-2-3-0-the-squiggly-heredoc) here to avoid the "inner" indentation?

mike-burns: I thought about it, but I actually _do_ want the two spaces of indentation.

PR #897, opened 2018-03-06
Fix normalize.css Sass import

tysongach:
When an `@import` in Sass ends with `.css`, Sass doesn't pull that file
in as a partial and concatenate those styles with the rest of the
manifest. Rather, it outputs a CSS `@import`, which we don't want here.


creuter:
👍

PR #898, opened 2018-03-09
Move enforce_ssl to production:enforce_tls

mike-burns:
Move the `suspenders:enforce_ssl` generator to
`suspenders:production:enforce_tls`. This namespaces it as something
specific to production (useful for those who are looking for the
Heroku-specific functionality), and it renames SSL to TLS, as that is
the abbreviation used in this century.

While here, shuffle some things around, which exposes existing bugs.
And fix those bugs!


lib/suspenders/generators/production/email_generator.rb
```diff
@@ -4,15 +4,15 @@ module Suspenders
...
+        File.join("..", "..", "..", "..", "templates"),
         File.dirname(__FILE__),
       )
 
       def smtp_configuration
         copy_file "smtp.rb", "config/smtp.rb"
 
         prepend_file "config/environments/production.rb",
-          %{require Rails.root.join("config/smtp")\n\n}
+          %{require Rails.root.join("config/smtp")\n}
```
mike-burns: We, unfortunately, have a test that ensures our `production.rb` has no vertical whitespace in it.

jferris:
This makes me twitchy because Rails calls it `enforce_ssl`. That means the generator name doesn't match the setting name. On the other hand, the setting name doesn't match the protocol! What's a dev to do?

mike-burns:
I've been trying to name the generators after what they do and not how it is implemented. For example, the `suspenders:lint` generator configures Hound, but if we want to change from Hound to Code Climate in the future we won't need to rename it. Similarly `suspenders:js_driver` (c-w), `suspenders:testing` (Rspec), `suspenders:analytics` (Segment), etc.

Therefore, maybe `suspenders:force_tls` is shorter and nicer. Eh?

jferris:
That makes sense and I feel better now.

PR #899, opened 2018-03-09
Introduce suspenders:production:timeout

mike-burns:
This introduces `suspenders:production:timeout`, a generator for
configuring an app to cut requests short. Here it uses Heroku's
Rack::Timeout middleware.

This is a production concern in two ways:

- We don't want to hit the timeout locally while developing. Our dev
environment runs slowly, with uncached assets and debugging breakpoints,
so a timeout simply does not make sense.
- We host on Heroku, which handles timeouts by killing the process. By
using Rack::Timeout we can more gracefully handle the timeout ourselves.


mike-burns:
Hey Mike, can we move the gem aspect of this into this diff too?

That's a great idea, Mike! I'll get to that now.

mike-burns:
The stuff that is currently deemed Heroku-specific is at

https://github.com/thoughtbot/suspenders/blob/master/lib/suspenders/generators/app_generator.rb#L124-L136
https://github.com/thoughtbot/suspenders/blob/master/lib/suspenders/generators/app_generator.rb#L224-L228

The flag is more about creating and managing a Heroku app and less about configuring the Rails app to be ready for Heroku. Passing `--heroku` will cal the `heroku` command at various points.

---

One major point of this refactoring is to remove any need for flags on the command-line, so I'm not excited about adding one now. Every generator under the `suspenders:production` namespace is Heroku-specific.

derekprior:
:+1:

PR #900, opened 2018-03-16
Test the suspenders:production:email generator

mike-burns:
This is the first isolated test for a generator in this project. It
introduces a few new test helpers:

- `#with_app` takes a block and runs the block after ensuring a fresh
app.
- `#generate` runs the specified generator.
- `#match_contents` is a matcher that takes a regexp as an argument and
applies in the context for a filename expectation. It reads the file
within the dummy app and returns whether the given regexp matches the
contents of the file.
- `#rails_new` generates a Rails app.
- `#system_rails_bin` is the Rails binary.
- `#project_rails_bin` is the `bin/rails` within an app.

This test runs in isolation in approximately five seconds on my
computer. This is both slow and one of the fastest tests in the repo.

---

While here, remove two race conditions because filesystem race conditions irk me.

### DIFF
--- a/spec/support/match_contents_matcher.rb
+++ b/spec/support/match_contents_matcher.rb
@@ -1,0 +1,6 @@
+RSpec::Matchers.define :match_contents do |regexp|
+  match do |filename|
+    contents = IO.read(File.join(project_path, filename))
+    contents =~ regexp
+  end
+end
--- a/spec/support/suspenders.rb
+++ b/spec/support/suspenders.rb
@@ -11,52 +11,81 @@ module SuspendersTestHelpers
 
   def run_suspenders(arguments = nil)
     arguments = "--path=#{root_path} #{arguments}"
-    Dir.chdir(tmp_path) do
-      Bundler.with_clean_env do
-        add_fakes_to_path
-        with_revision_for_honeybadger do
-          `
-            #{suspenders_bin} #{APP_NAME} #{arguments}
-          `
+    run_in_tmp do
+      add_fakes_to_path
+
+      with_revision_for_honeybadger do
+        `#{suspenders_bin} #{APP_NAME} #{arguments}`
+      end
+
+      Dir.chdir(APP_NAME) do
+        with_env("HOME", tmp_path) do
+          `git add .`
+          `git commit -m 'Initial commit'`
         end
-        Dir.chdir(APP_NAME) do
-          with_env("HOME", tmp_path) do
-            `git add .`
-            `git commit -m 'Initial commit'`
-          end
+      end
+    end
+  end
+
+  def with_app
+    drop_dummy_database
+    remove_project_directory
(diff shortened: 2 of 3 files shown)


spec/support/suspenders.rb
```diff
@@ -29,6 +29,51 @@ def run_suspenders(arguments = nil)
...
+        end
+      end
+    end
+  end
+
+  def generate(generator)
+    Dir.chdir(project_path) do
+      Bundler.with_clean_env do
+        with_revision_for_honeybadger do
+          `bin/spring stop`
```
mike-burns: This is based on lessons learned in 3cf23ed37a94a890f7678a8ec5cd072385f93bb4.

spec/support/suspenders.rb
```diff
@@ -40,23 +85,23 @@ def suspenders_help_command
   end
 
   def setup_app_dependencies
-    if File.exist?(project_path)
-      Dir.chdir(project_path) do
-        Bundler.with_clean_env do
-          `bundle check || bundle install`
-        end
+    Dir.chdir(project_path) do
+      Bundler.with_clean_env do
```
georgebrock: This pattern of `Dir.chdir(project_path) { Bundler.with_clean_env { … } }` seems to be repeated quite a lot. How about a `#run_in_project` method?

spec/support/suspenders.rb
```diff
@@ -29,6 +29,52 @@ def run_suspenders(arguments = nil)
     end
   end
 
+  def with_app(&block)
+    drop_dummy_database
+    remove_project_directory
+    rails_new
+    setup_app_dependencies
+
+    block.call
```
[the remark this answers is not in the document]
georgebrock: I agree with the Hound on this one.

spec/support/match_contents_matcher.rb
```diff
@@ -0,0 +1,6 @@
+RSpec::Matchers.define :match_contents do |regexp|
+  match do |filename|
+    contents = IO.read(File.join(project_path, filename))
```
seanpdoyle: Would [`Pathname(project_path, filename).read`](https://ruby-doc.org/stdlib-2.1.3/libdoc/pathname/rdoc/Pathname.html#method-i-read) yield the same value for `contents`?

seanpdoyle: Ignore that, `Pathname` and `Pathname.new` don't take variable arguments, they both expect a single argument string.

mike-burns: I couldn't find a `Pathname` constructor that takes multiple arguments, unfortunately -- so it'd still be `Pathname(File.join(project_path, filename)).read`. Dunno if that reads any more nicely.

mike-burns: Ah you just said that. Yup!

PR #911, opened 2018-05-01
Run 'bundle install' after adding new gems

### LINKED ISSUE #910: Generators fail if certain gems aren't already installed
If `bullet`, `capybara-webkit`, and `factory_bot_rails` are not already installed before generating a new app, several of the generators fail with bundler errors like `"Could not find gem 'bullet' in any of the gem sources listed in your Gemfile."`

Steps to reproduce:

```
gem uninstall bullet capybara-webkit factory_bot_rails
suspenders test_app
```

composerinteralia:
Fixes #910

Adding gems in a generator without rerunning `bundle install` can cause
problems. If the gems aren't already installed on the system, the
remaining generators will fail with a bundler error.

### DIFF
--- a/lib/suspenders/generators/db_optimizations_generator.rb
+++ b/lib/suspenders/generators/db_optimizations_generator.rb
@@ -4,6 +4,7 @@ module Suspenders
   class DbOptimizationsGenerator < Rails::Generators::Base
     def add_bullet
       gem "bullet", group: %i(development test)
+      Bundler.with_clean_env { run "bundle install" }
     end
 
     def configure_bullet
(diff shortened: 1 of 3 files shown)


composerinteralia:
Eep! PR 911

@mike-burns this might be related to the dev dependencies you added in #909. My impression is that this change would make those dev dependencies unnecessary, but I might be missing something.

I am having some trouble running the test suite (both before and after these changes). I am consistently getting four green dots and then it hangs forever. I will try to dig into it more later this week.

mike-burns:
Confirmed, that fixed it. Thank you!

PR #912, opened 2018-05-02
Update to Rails 5.2.0

composerinteralia:
Rails 5.2.0 will give us ActiveStorage and all sorts of other goodies:
http://edgeguides.rubyonrails.org/5_2_release_notes.html

rails/rails#30016 introduced `Rails::AppBuilder#ruby_version` to create
a .ruby_version file. To avoid getting stuck in an overwrite prompt
when we try to create our own .ruby_version file based on
`Suspenders::RUBY_VERSION`, I overrode `#ruby_version`.

Rails 5.2.0 adds bootsnap to the Gemfile and to config/boot.rb. I added it
to our Gemfile, but we could also `skip_bootsnap` by default if we don't
want to use it.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -162,7 +162,7 @@ config.public_file_server.headers = {
       end
     end
 
-    def set_ruby_to_version_being_used
+    def ruby_version
       create_file '.ruby-version', "#{Suspenders::RUBY_VERSION}\n"
     end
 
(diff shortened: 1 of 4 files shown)


toobulkeh:
Looks like you're missing the addition of a few lines for bootsnap to work:

https://github.com/Shopify/bootsnap#usage

composerinteralia:
@toobulkeh Rails 5.2 adds `require 'bootsnap/setup'` to `config/boot.rb` by default. Is there something else I am missing?

mike-burns:
Merged as d3c2e89f91784f60b2923b947114245a3c427a82, thank you!

PR #913, opened 2018-05-16
Added a generator for ChromeDriver

TheABrad:
This introduces the `suspenders:chromedriver` generator option
for javascript testing if you want to use headless-chrome instead of
capybara-webkit.
It installs capybara-selenium and chromedriver-helper.
I decided to add this optional generator since Rails 5.2 comes with chromedriver-helper on the default setup
and it is easier to set up than capybara-webkit if you want that option.


mike-burns:
This PR is fantastic, thank you.

One thing I'll want to check internally before merging it is whether we want to make this the default going forward. If not, this would be the first generator in Suspenders that is not run as part of any default flow, and I had not considered that possibility.

mike-burns:
And I'm back! The survey results:

sample size: 45
people who foolishly ignored me: 42
people who think this should be the default: 3

@TheABrad how do you feel about replacing `lib/suspenders/generators/js_driver_generator.rb` with this? It can remain `Suspenders::JsDriverGenerator`, but it'd generate a chromiumdriver config instead of capybara-webkit.

mike-burns:
Merged as cf78134766c689dab1d9e5db2f3be454ba81f487, thank you again!

PR #916, opened 2018-05-21
Use HTTPS url for Github Hub script

rickychilcott:
To prevent MITM attacks and to follow general best practices, change the `hub` url from `http://hub.github.com/standalone` to `https://hub.github.com/standalone`


mike-burns:
Thank you!

PR #920, opened 2018-07-05
Switch to ENV configuration of Rack::Timeout

### LINKED ISSUE #918: Breaking change to Rack::Timeout causes Heroku deployment failure
See [here](https://github.com/heroku/rack-timeout/issues/130) for more info. It looks like we can just update [this file](https://github.com/thoughtbot/suspenders/blob/master/lib/suspenders/generators/production/timeout_generator.rb) - I can try to open up a PR later tonight. Thanks!

orenyk:
Resolves #918

Rack::Timeout v0.5.0 removed legacy setter methods used for
configuration (https://github.com/heroku/rack-timeout/pull/125) such
that the configuration added by Suspenders causes a NoMethodError when
initializing Rails in the production environment (e.g. during Heroku
deployment). This PR switches to using environment variable-based
configuration of Rack::Timeout, matching the recommendations in the
README.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,6 @@
+* Bug fix: change production timeouts generator to use working configuration
+  method with latest version of Rack::Timeout
+
 1.47.0 (May 25, 2018)
 
 * Bug fix: normalize.css Sass import is concatenated with other styles now
--- a/lib/suspenders/generators/production/timeout_generator.rb
+++ b/lib/suspenders/generators/production/timeout_generator.rb
@@ -8,13 +8,13 @@ module Suspenders
       end
 
       def configure_rack_timeout
-        append_file "config/environments/production.rb", rack_timeout_config
+        append_file ".env", rack_timeout_config
       end
 
       private
 
       def rack_timeout_config
-        %{Rack::Timeout.timeout = ENV.fetch("RACK_TIMEOUT", 10).to_i}
+        %{RACK_TIMEOUT_SERVICE_TIMEOUT=10}
       end
     end
   end


composerinteralia:
Thanks! I'm going to go ahead and merge this, but I agree we could also get rid of the custom timeout altogether. In https://github.com/thoughtbot/suspenders/commit/059ead51d9977de27c7db9818d8d9e43ed240c7b it looks like we changed it from 5 seconds to 10 seconds because Heroku recommended that number. I'm not sure where we found that recommendation though; it looks like the default has been 15 seconds for a long time:

* https://github.com/heroku/rack-timeout/commit/da43026b0927bd489fa22fb66b3a7732cc10a656
* https://github.com/heroku/rack-timeout/commit/e599a9f13c769166788aa8a49152164335d8452a
* https://github.com/heroku/rack-timeout/commit/d1dd4810cfcd40f0c3d83334ea72da6f65feae6a
* https://github.com/heroku/rack-timeout/commit/9c37718b88307cbc49f473b073cc92ca63b664b0

croaky:
Sorry for lack of detail on 059ead5!

I believe the idea was that there were three levels of timeouts:

* Ruby (controlled by Rack Timeout)
* web server (Unicorn then, probably Puma now)
* Heroku routing mesh (30s)

Adding any timeout was supposed to avoid runaway processes and each lower level timeout was supposed to be lower in value to avoid losing debugging information.

`10s`-`15s`-`30s` might have been in Heroku's Devcenter somewhere at the time but I think the main goal in 059ead5 was to switch from `5s`-`5s`-`30s` to `i`-`i + n < 30`-`30`

PR #923, opened 2018-07-20
Introduce suspenders:json

mike-burns:
Tools for JSON parsing, serialization. So far all it does is install the
`oj` gem.

This comes from our guides, which has recommended the `oj` gem for all
our projects. This kind of information should live in Suspenders instead
of our guides, so lets pull it into here.

Inspired by https://github.com/thoughtbot/guides/pull/509 .


mike-burns:
Good idea. I should check whether any of that README is up-to-date still, first.

composerinteralia:
Is the README the only thing holding us back from merging this? If so let's go ahead and merge and open a separate issue for making sure the README is up to date.

PR #925, opened 2018-07-26
Fix mysql and sqlite3 database option bug

### LINKED ISSUE #893: Unable to create project with sqlite3 database
When running `suspenders foobar --database sqlite3` I get the following error

```
Specified 'sqlite3' for database adapter, but the gem is not loaded. Add `gem 'sqlite3'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord). (Gem::LoadError)
```

edgarlepe:
Fixes #893 and adds support for mysql as well.

If one tried to use a database other than postgres, the following error would occur:
```
Specified 'some_database_other_than_pg' for database adapter, but the gem is not loaded. Add `gem 'some_databse_other_than_pg'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord).
```
This error occurred because the `pg` gem was hard coded into the Gemfile template, so the `pg` gem was always loaded regardless of the `--database` option.


mike-burns:
We do not support databases other than PostgreSQL in Suspenders -- we simply don't have the maintainer-power to handle another permutation of configurations, and none of us use anything besides PG. Additionally, we are moving away from configuration flags and pivoting Suspenders into a collection of generators.

Closing this.

Thank you for the submission!

PR #929, opened 2018-08-05
Create local heroku setup only if heroku option is true

delphaber:
I think that these actions should be run only if `options[:heroku]` is true, but let me know if I'm wrong :)

### DIFF
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -114,10 +114,12 @@ module Suspenders
     end
 
     def create_local_heroku_setup
-      say "Creating local Heroku setup"
-      build :create_review_apps_setup_script
-      build :create_deploy_script
-      build :create_heroku_application_manifest_file
+      if options[:heroku]
+        say "Creating local Heroku setup"
+        build :create_review_apps_setup_script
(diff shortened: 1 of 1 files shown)


mike-burns:
The README says:

> You can optionally create Heroku staging and production apps:
> 
>     suspenders app --heroku true

The build steps you're conditionalizing is for adding scripts and config to your local repo. So strictly speaking, `--heroku` does not affect them.

I'm going to be pedantic about this one because of the direction we're trying to take Suspenders in: instead of command-line flags, we'd like Suspenders to be a collection of Rails generators. You can see an example in [the `generate_production_default` step](https://github.com/thoughtbot/suspenders/blob/master/lib/suspenders/generators/app_generator.rb#L200-L204) just below the `create_local_heroku_setup` step. Ideally `create_local_heroku_setup` would move into generator(s), too, which would mean that you can undo any generated files you don't want.

(Long view: instead of generating all, you'd generate just the features you need in the moment.)

---

Therefore, I'm going to close this. We very much welcome PRs that add generators for the Heroku steps. Be warned that we're still figuring out how we want our generators to work and exist so there will likely be active discussion and churn on such a PR.

Thank you very much for opening this!

PR #931, opened 2018-08-14
Add tzinfo-data on Windows

markspolakovs:
On Windows, the tzinfo-data gem is required, and Rails fails with an error if it is started without it. This PR adds it to the default Gemfile.


mike-burns:
Merged as 2ca2846b996b6288596fa19b57a3c81c85bd2b3b. Thank you for finding and fixing this!

PR #937, opened 2018-09-09
Use Release Phase for database migration

geoffharcourt:
Heroku's [Release Phase](https://devcenter.heroku.com/articles/release-phase) feature provides a much better migration experience for a production deploy of a Rails application.

Without using a release phase task, here's how you migrate your database
in production:

1. Deploy your application via Git or pipeline promotion.
2. With the migration deployed, run `rake db:migrate`.
3. Immediately restart the application so that Rails is now aware of the
schema changes.

This process is awkward and can lengthen downtime in deploys. It results
in the application being restarted twice (once for the deployment of
code, and then again post-migration), and also may push application
logic that depends on the new code into production before the migration
has run or finished, which can lead to unnecessary exceptions.

There are strategies to mitigate these problems, but they generally
involve many small commits and deploys that must be carefully
orchestrated (not pushed all at once) that can be laborious and error
prone.

This commit leverages the Heroku Release Phase task to execute a
smoother migration. Here's how the process works:

1. Start your deploy with Git or a pipeline promotion
2. Heroku builds or promotes your application slug
3. Before deploying the application code to dynos, a run dyno is started
and it executes the release phase task. In this case, it's a shell
script that runs `rake db:migrate` if the application has an
`AUTO_MIGRATE_DB` environment variable set.
4. After the migration finishes, Heroku deploys the application and
restarts. If the release phase failed (migration problem), the deploy does not
proceed.

This process reduces the chance of new application code that depends on
the un-run migration being deployed and reduces the number of
application restarts from two to one. We've been using it at CommonLit
for several months now and noticed a dramatic downturn in
deployment-related downtime.

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -217,15 +217,7 @@ config.public_file_server.headers = {
     end
 
     def configure_automatic_deployment
-      deploy_command = <<-YML.strip_heredoc
-      deployment:
-        staging:
-          branch: master
-          commands:
-            - bin/deploy staging
-      YML
-
-      append_file "circle.yml", deploy_command
+      append_file "Procfile", "release: bin/auto_migrate"
     end
 
     def create_github_repo(repo_name)
--- a/lib/suspenders/generators/production/manifest_generator.rb
+++ b/lib/suspenders/generators/production/manifest_generator.rb
@@ -10,6 +10,7 @@ module Suspenders
           scripts: {},
           env: {
             APPLICATION_HOST: { required: true },
+            AUTO_MIGRATE_DB: { value: "true" },
             EMAIL_RECIPIENTS: { required: true },
             HEROKU_APP_NAME: { required: true },
             HEROKU_PARENT_APP_NAME: { required: true },
--- a/templates/bin_deploy
+++ b/templates/bin_deploy
@@ -8,5 +8,3 @@ branch="$(git symbolic-ref HEAD --short)"
 target="${1:-staging}"
 
 git push "$target" "$branch:master"
-heroku run rails db:migrate --exit-code --remote "$target"
-heroku restart --remote "$target"
--- a/templates/bin_auto_migrate
+++ b/templates/bin_auto_migrate
@@ -1,0 +1,5 @@
+ set -e
+
+ if [ -n "$AUTO_MIGRATE_DB" ]; then
+   bundle exec rake db:migrate
+ fi
--- a/templates/bin_setup_review_app.erb
+++ b/templates/bin_setup_review_app.erb
@@ -17,6 +17,5 @@ heroku pg:backups:capture --app $PARENT_APP_NAME
 URL=`heroku pg:backups public-url --app $PARENT_APP_NAME`
 
 heroku pg:backups restore $URL DATABASE_URL --confirm $APP_NAME --app $APP_NAME
-heroku run rails db:migrate --exit-code --app $APP_NAME
 heroku ps:scale worker=1 --app $APP_NAME
 heroku restart --app $APP_NAME


geoffharcourt:
I just noticed this change is almost identical to #906, with the difference that this script is opt-in for environments that shouldn't run migrations (for example, we don't run migrations on PR apps on every deploy). We could combine this release phase task with the other work in #906.

rikkipitt:
This should have the added benefit of being able to `./bin/deploy` and then allowing the use of `ctrl+c` to detach from the build process.

https://devcenter.heroku.com/articles/git#detaching-from-the-build-process

mike-burns:
Discussed out of band:

- The env var here is great.
- The lack of custom CI script in #906 is great.
- Let's merge these two solutions and get it into Suspenders.

geoffharcourt:
@toobulkeh I think your PR does more than mine does, do you want to crib my ENV var solution into your PR? If you're busy I'm happy to pull some of your work into this PR.

geoffharcourt:
@mike-burns I added a second commit to copy over @toobulkeh's nice CircleCI improvement. I think we should be ready for final review.

templates/app.json.erb
```diff
@@ -0,0 +1,30 @@
+{
```
composerinteralia: How does this template relate to the [manifest generator](https://github.com/thoughtbot/suspenders/blob/5af02dad438ceed099a9b042ad0fc9a0e2feaba0/lib/suspenders/generators/production/manifest_generator.rb#L11)

lib/suspenders/app_builder.rb
```diff
@@ -217,15 +217,7 @@ def create_heroku_apps(flags)
...
-      deploy_command = <<-YML.strip_heredoc
-      deployment:
-        staging:
-          branch: master
-          commands:
-            - bin/deploy staging
-      YML
-
-      append_file "circle.yml", deploy_command
+      append_file "Procfile", "release: bundle exec rails db:migrate"
```
composerinteralia: Should this be:

```suggestion
      append_file "Procfile", "release: bin/auto_migrate"
```

templates/Procfile
```diff
@@ -1,2 +1,3 @@
+release: bin/auto_migrate
```
composerinteralia: The `configure_automatic_deployment` method in app_builder.rb adds also adds a `release` process. My impression is that we don't need both.

templates/bin_deploy
```diff
@@ -8,5 +8,4 @@ branch="$(git symbolic-ref HEAD --short)"
 target="${1:-staging}"
 
 git push "$target" "$branch:master"
-heroku run rails db:migrate --exit-code --remote "$target"
 heroku restart --remote "$target"
```
composerinteralia: If I am understanding all of this correctly we can get rid of the restarts as well.

templates/app.json.erb
geoffharcourt: Oh, yikes, I didn't realize this is where this went. I'll update.

templates/bin_deploy
geoffharcourt: That's correct!

templates/Procfile
geoffharcourt: Yup. Wow, this has drifted so much since I started, thank you so much for carefully reviewing these changes!!!

geoffharcourt: I kept the version in app builder so folks could opt-in/out with generators as needed.

lib/suspenders/generators/production/manifest_generator.rb
```diff
@@ -10,6 +10,7 @@ def render_manifest
           scripts: {},
           env: {
             APPLICATION_HOST: { required: true },
+            AUTO_MIGRATE_DB: { value: "true" },
```
geoffharcourt: Note: this will make PR apps auto-migrate. I'm not sure if this should be here or not, we don't use it at CommonLit and prefer for a migration-carrying PR to have engineers migrate if/when necessary.

geoffharcourt:
@composerinteralia & @mike-burns, I think we're ready for another review.

PR #938, opened 2018-09-28
Fix email generator

rikkipitt:
The email generator attempts to inject SMTP config into `config/environments/production.rb` after the line `config.action_mailer.raise_delivery_errors = false`. This line is not found anymore and thus fails (NB: find reference).

I propose the code is injected after `config.action_mailer.perform_caching = false` instead.

In production, you may experience the following error as action mailer attempts to use the default port 25:

```
Connection refused - connect(2) for "localhost" port 25
/app/vendor/ruby-2.5.1/lib/ruby/2.5.0/net/smtp.rb:539:in `initialize'
/app/vendor/ruby-2.5.1/lib/ruby/2.5.0/net/smtp.rb:539:in `open'
/app/vendor/ruby-2.5.1/lib/ruby/2.5.0/net/smtp.rb:539:in `tcp_socket'
/app/vendor/ruby-2.5.1/lib/ruby/2.5.0/net/smtp.rb:549:in `block in do_start'
/app/vendor/ruby-2.5.1/lib/ruby/2.5.0/timeout.rb:93:in `block in timeout'
/app/vendor/ruby-2.5.1/lib/ruby/2.5.0/timeout.rb:103:in `timeout'
/app/vendor/ruby-2.5.1/lib/ruby/2.5.0/net/smtp.rb:548:in `do_start'
/app/vendor/ruby-2.5.1/lib/ruby/2.5.0/net/smtp.rb:518:in `start'
...
...
```

### DIFF
--- a/lib/suspenders/generators/production/email_generator.rb
+++ b/lib/suspenders/generators/production/email_generator.rb
@@ -26,7 +26,7 @@ module Suspenders
         RUBY
 
         inject_into_file "config/environments/production.rb", config,
-          after: "config.action_mailer.raise_delivery_errors = false"
+          after: "config.action_mailer.perform_caching = false"
       end
 
       def env_vars


rikkipitt:
Further investigation:

`config.action_mailer.perform_caching = false` is in the default Rails `production.rb` config file (commented out), thus works via tests. It seems that the comment removal task is being run before the email_generator, hence the problem I'm finding and why the email_generator_spec works. An ordering issue?

Explicitly adding:

```
it "configs email delivery method for production" do
    expect(production_config).
      to match(/^ +config.action_mailer.delivery_method = :smtp$/)
end
```

To the new_project_spec.rb shows the expected failure...

composerinteralia:
Thanks! I ran into this problem this week.

https://github.com/thoughtbot/suspenders/blob/master/spec/features/production/email_spec.rb runs the email generator directly without going through the rest of the suspenders setup, so indeed the comments are still in the file for that test.

Another option would be to add the configuration at the end, like we do here: https://github.com/thoughtbot/suspenders/blob/285c002cb1095a0031444840b8ee1c31324264e8/lib/suspenders/actions.rb#L26.

rikkipitt:
Thanks for merging in. I was wondering about the tests, I came back to it to add in a case for it but didn't have enough time to fully figure it out...

composerinteralia:
Thanks for the PR! We ended up with a failing test once I rebased on master 🤷‍♂️ , so I added https://github.com/thoughtbot/suspenders/commit/476dabb614cd210a91eb1c293b03b7deb9452056. Hopefully that commit will future proof this a bit.

PR #939, opened 2018-10-01
We need to install JS dependencies in setup

danbee:
This gets run when first creating a Suspenders project but not on setup. It should be.

### DIFF
--- a/templates/bin_setup
+++ b/templates/bin_setup
@@ -10,6 +10,9 @@ set -e
 gem install bundler --conservative
 bundle check || bundle install
 
+# Install JavaScript dependencies
+bin/yarn
+
 # Set up database and add any development seed data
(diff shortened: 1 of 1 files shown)


templates/bin_setup
```diff
@@ -10,6 +10,9 @@ set -e
 gem install bundler --conservative
 bundle check || bundle install
 
+# Install JavaScript dependencies
+bin/yarn
```
composerinteralia: Will there always be a `bin/yarn`?

composerinteralia: Based on our conversation IRL, it seems like the answer is yes

danbee: Yes, it gets run on `suspenders new` to install Normalize.css.

PR #947, opened 2018-10-19
Strip extra newline from the production config

composerinteralia:
With e09a56d we are getting a failing spec because of an extra newline
in the production config. Removing comments later will take care of the
extra newlines, and I think may have made e09a56d unnecessary.


mike-burns:
Huh.

Yeah sure LGTM.

PR #951, opened 2018-11-06
Move the review app check before asset host

danbee:
The ASSET_HOST environment variable needs to be set correctly on Heroku review apps. Before this change the review app check was happening *after* the setting of the asset host and consequently assets weren't loading.

This moves the review app check to the top of the config file so the ASSET_HOST gets set correctly on review apps.


composerinteralia:
I just noticed https://github.com/thoughtbot/suspenders/pull/928, which seems related

PR #953, opened 2018-12-07
Bump to Ruby 2.5.3

mike-burns:


mjankowski:
Might need to also change? https://github.com/thoughtbot/suspenders/blob/master/.travis.yml#L2

PR #956, opened 2018-12-28
Introduce suspenders:production:deployment

mike-burns:
This moves the Heroku deployment script generation into a one-off
generator. Now `create_local_heroku_setup` only does one thing, bringing
us one step closer to pushing that method entirely into generators.

This generator copies in a `bin/deploy` executable and updates the
project's README with a mention of its existence.

---

Introduces a `be_executable` RSpec matcher, parameterized over a file
path string. It puts the file path into the project root and determines
whether the file is executable.

---

Introduces a `:generator` spec type. This type is used to indicate that
the test is at a smaller "unit", instead of a full system test. For
example, testing a generator.

We then use this hook to ensure that the dummy project is cleared before
running a generator example.

---

Paired with @dimanyc.


lib/suspenders/generators/production/deployment.rb
```diff
@@ -0,0 +1,32 @@
+require "rails/generators"
+
+module Suspenders
+  module Production
+    class DeploymentGenerator < Rails::Generators::Base
+      source_root File.expand_path(
+        File.join("..", "..", "..", "..", "templates"),
```
composerinteralia: Would it be helpful to have a `Suspenders::Generators::Base` that inherits from `Rails::Generators::Base` and knows where all the templates live?

spec/support/be_executable_matcher.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+RSpec::Matchers.define :be_executable do
+  match do |filename|
+    File.new(File.join(project_path, filename)).stat.executable?
```
composerinteralia: Is this the same thing?

```suggestion
    File.stat(File.join(project_path, filename)).executable?
```

spec/features/production/deployment_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect("README.md").to match_contents(%r{bin/deploy})
+  end
+
+  it "destroys the configuration for a production deployment" do
+    touch "bin/deploy"
+
+    with_app { destroy("suspenders:production:deployment") }
+
+    expect("bin/deploy").not_to exist_as_a_file
+    expect("README.md").not_to match_contents(%r{bin/deploy})
```
composerinteralia: Does this spec have any value if we don't write to the README first?

lib/suspenders/generators/production/deployment.rb
mike-burns: Believe it or not, I was just thinking that about while reviewing #957. I'm going to do that as a separate commit in which I refactor everything.

spec/support/be_executable_matcher.rb
mike-burns: Word.

spec/features/production/deployment_spec.rb
mike-burns: I've no idea. In the least it's documentation, and at worst it makes sure we don't slip something about `bin/deploy` into the README unless it's through this generator.

Actually writing to the README first would be kinda tedious so I'm leaning toward not doing it.

PR #957, opened 2018-12-28
Introduce suspenders:production:manifest

mike-burns:
This generator copies in the `app.json` needed for Heroku, defining our
required environment variables and the app name.

It uses the RSpec type `:generator`, defined in a different branch.
Hopefully by the time this is merged, it will all just work.

---

Paired with @dimanyc.


lib/suspenders/generators/production/manifest_generator.rb
```diff
@@ -0,0 +1,22 @@
...
+        File.dirname(__FILE__),
+      )
+
+      def render_manifest
+        template "app.json.erb", "app.json"
+      end
+
+      private
+
+      def app_name
```
mike-burns: In a future PR I'm going to extract `app_name` and the `source_root` to a common base class.

PR #936, opened 2018-09-07
Changed sass-rails to sassc-rails

theabrad:
The Ruby Sass gem will be deprecated by March 26, 2019
http://sass.logdown.com/posts/7081811 so the suggestion is to use the
`sassc` gem. This adds the `sassc-rails` gem to the Gemfile and removes
`sass-rails`.

PR #959, opened 2019-01-08
Use sassc-compatible versions of neat and bourbon

composerinteralia:
To merge #936, we need to make sure we have a version of bourbon with
https://github.com/thoughtbot/bourbon/commit/8257c683f8a250c06407f66d077dfbe6ce6595b0
and a version of neat with
https://github.com/thoughtbot/neat/commit/a092ff79a5f81be5a01328f9daa5e15f97b4a58a

I also got rid of the pessimistic version constraint so a fresh
suspenders run will always give us the latest versions of these gems.

### DIFF
--- a/lib/suspenders/generators/stylesheet_base_generator.rb
+++ b/lib/suspenders/generators/stylesheet_base_generator.rb
@@ -7,8 +7,8 @@ module Suspenders
       File.dirname(__FILE__))
 
     def add_stylesheet_gems
-      gem "bourbon", "~> 5.0"
-      gem "neat", "~> 2.1"
+      gem "bourbon", ">= 5.0.1"
+      gem "neat", ">= 3.0.1"
       Bundler.with_clean_env { run "bundle install" }
     end
 


SViccari:
Good morning 🌤 I'm uncertain how these versions of [bourbon](https://github.com/thoughtbot/bourbon/blob/5b4ae4418386006a12d02d8aba5df51e8097252d/bourbon.gemspec#L11) and [neat](https://github.com/thoughtbot/neat/blob/92a70ff596a5d2932dda108967508f3ba8d3a602/neat.gemspec#L10) are `sassc-compatible` since both versions still list `sass` as a dependency. Can `bourbon` and `neat` be `sassc-compatible` while still depending on `sass`? Thanks!

tysongach:
@SViccari There is [an open issue on Bourbon](https://github.com/thoughtbot/bourbon/issues/1064) which speaks to removing the `sass` (Ruby Sass) runtime dependency. We should do the same over on Neat. I don’t think having the Ruby Sass dependency breaks SassC, but it does mean it always gets installed, which is a bummer.

I’d be happy to review a PR, if you’re up for it!

SViccari:
@tysongach 
> I don’t think having the Ruby Sass dependency breaks SassC, but it does mean it always gets installed, which is a bummer.

Ah, okay. That makes sense, thanks! 

> I’d be happy to review a PR, if you’re up for it!
sounds good! I may not have time to get to this today but 🤞

PR #961, opened 2019-01-11
Suspenders::Generators::Base base class

mike-burns:
We keep repeating the same code; time to extract it into a shared
object. Since Thor pulls us into inheritance, use a superclass.

This superclass includes our custom actions, defines an app name, and
sets the source root. I could not figure out how to use the
`source_root` class method; since we're inheriting from the
Rails::Generators::Base we can override the `default_source_root` class
method instead to achieve the same effect.

### DIFF
--- a/lib/suspenders/generators/base.rb
+++ b/lib/suspenders/generators/base.rb
@@ -1,0 +1,20 @@
+require "rails/generators"
+require_relative "../actions"
+
+module Suspenders
+  module Generators
+    class Base < Rails::Generators::Base
+      include Suspenders::Actions
+
+      def self.default_source_root
+        File.expand_path(File.join("..", "..", "..", "templates"), __dir__)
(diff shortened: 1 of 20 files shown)


lib/suspenders/generators/base.rb
```diff
@@ -0,0 +1,20 @@
...
+      include Suspenders::Actions
+
+      def self.default_source_root
+        File.expand_path(File.join("..", "..", "..", "templates"), __dir__)
+      end
+
+      private
+
+      def app_name
+        Rails.application.class.parent_name.demodulize.underscore.dasherize
```
composerinteralia: It looks like `Rails.application` returns `app_class.instance`, so we could skip a step:

```suggestion
        Rails.app_class.parent_name.demodulize.underscore.dasherize
```

PR #969, opened 2019-02-15
Introduce suspenders:production:compression

mike-burns:
This generator adds `Rack::Deflater` as a middleware in the production
environment.


composerinteralia:
CI now failing because of https://github.com/thoughtbot/suspenders/pull/976. If you rebase I expect it will pass.

lib/suspenders/generators/production/compression_generator.rb
```diff
@@ -0,0 +1,15 @@
+require_relative "../base"
+
+module Suspenders
+  module Production
+    class CompressionGenerator < Generators::Base
+      def add_rack_deflater
+        inject_into_file(
```
composerinteralia: https://github.com/thoughtbot/suspenders/blob/cf78134766c689dab1d9e5db2f3be454ba81f487/lib/suspenders/actions.rb#L22-L28

```suggestion
        configure_environment(:production, "config.middleware.use Rack::Deflater")
```

mike-burns:
This is merged but GH doesn't like admitting as much.

PR #972, opened 2019-03-05
Update ruby to 2.6.1

delphaber:
It seems they are already testing against ruby 2.6: https://github.com/rails/rails/blob/master/.travis.yml :)


composerinteralia:
Thanks!

PR #974, opened 2019-03-08
Add stylelint generator

composerinteralia:
As mentioned in https://github.com/thoughtbot/guides/pull/532, we ran a
successful research experiment with stylelint.

This PR changes the hound.yml to use stylelint intead of scss-lint. It
also adds a generator to set up stylelint to use [thoughtbot's
stylelint config](https://github.com/thoughtbot/stylelint-config/)

Co-authored-by: Tyson Gach <tyson@tysongach.com>


spec/support/suspenders.rb
```diff
@@ -45,9 +45,7 @@ def rails_new
       end
 
       Dir.chdir(APP_NAME) do
-        File.open("Gemfile", "a") do |file|
-          file.puts %{gem "suspenders", path: #{root_path.inspect}}
-        end
+        append_dependencies_to_gemfile
```
composerinteralia: I didn't mean to include this. Will get rid of it before merging.

templates/hound.yml
```diff
@@ -10,5 +10,6 @@ ruby:
   # config_file: .ruby-style.yml
   enabled: true
 scss:
-  # config_file: .scss-style.yml
+  enabled: false
+stylelint:
```
composerinteralia: We will need to add the `config_file` key in the stylelint generator. One approach is to have a comment here, the way we do with the others, and then have the generator uncomment it.

lib/suspenders/generators/stylelint_generator.rb
```diff
@@ -0,0 +1,22 @@
...
+      copy_file "stylelintrc.json", ".stylelintrc.json"
+    end
+
+    def install_dependencies
+      case self.behavior
+      when :invoke
+        run "bin/yarn add stylelint --dev"
+        run "bin/yarn add @thoughtbot/stylelint-config --dev"
+      when :revoke
+        self.behavior = :invoke
```
composerinteralia: This seems silly, but I don't know if there is a better way. The run command doesn't do anything when the behavior is `:revoke`: https://github.com/erikhuda/thor/blob/d3762d0b89a80da488285d640b052f5276fac6bb/lib/thor/actions.rb#L245-L246

composerinteralia:
Hm, I'm seeing:

```
warning " > @thoughtbot/stylelint-config@0.2.0" has unmet peer dependency "stylelint@^9.x".                                                                                     
warning "@thoughtbot/stylelint-config > stylelint-order@2.1.0" has unmet peer dependency "stylelint@^9.10.1".                                                                   
warning "@thoughtbot/stylelint-config > stylelint-scss@3.5.4" has unmet peer dependency "stylelint@^8.0.0 || ^9.0.0".    
```

lib/suspenders/generators/stylelint_generator.rb
composerinteralia: I had to do something similar for the hound commenting. Maybe I need some sort of custom action here?

templates/hound.yml
composerinteralia: Added

tysongach:
@composerinteralia When you see that warning, what are the contents of your `package.json`?

composerinteralia:
It was because we were removing `stylelint` before `@thoughtbot/stylelint-config` on `destroy`.

composerinteralia:
CI failing because of https://github.com/thoughtbot/suspenders/pull/976

lib/suspenders/generators/stylelint_generator.rb
```diff
@@ -0,0 +1,74 @@
...
+    class YarnInstall
+      def initialize(base, dependencies, flags)
+        @base = base
+        @dependencies = dependencies
+        @flags = flags
+      end
+
+      def invoke!
+        @dependencies.each do |dependency|
+          @base.run "bin/yarn add #{dependency} #{@flags}"
```
seanpdoyle: Have you considered running `yarn add` with multiple package names?

```rb
dependencies_argument = @dependencies.join(" ")

@base.run "bin/yarn add #{dependencies_argument} #{@flags}"
```

spec/features/stylelint_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      with_app { generate("suspenders:stylelint") }
+
+      expect(".hound.yml").to(
+        match_contents(/^  config_file: \.stylelintrc\.json/),
+      )
+    end
+  end
+
+  context "destroy" do
+    it "creates .stylelintrc.json" do
```
seanpdoyle: Was this docstring copied from elsewhere?

Based on this block's expectation, would `it "removes  .stylelintrc.json"` be more descriptive?

spec/features/stylelint_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      with_app do
+        generate("suspenders:stylelint")
+        destroy("suspenders:stylelint")
+      end
+
+      expect(".stylelintrc.json").not_to exist_as_a_file
+    end
+
+    it "adds stylelint and @thoughtbot/stylelint-config to the package.json" do
```
seanpdoyle: Should this mention removing the dependencies?

lib/suspenders/generators/stylelint_generator.rb
composerinteralia: Yeah, that seems reasonable. Maybe I will also be able to get rid of the `reverse` if I do it that way.

spec/features/stylelint_spec.rb
composerinteralia: Good catch

spec/features/stylelint_spec.rb
composerinteralia: Eep!. Bad copying from other examples.

tysongach:
🎉 Yay! Thank you all for the effort here.

PR #976, opened 2019-04-20
Avoid installing autoprefixer in api mode

composerinteralia:
spec/features/api_spec.rb is failing on CI because Rails 5.2.3 in API
mode isn't playing well with autoprefixer-rails. It doesn't really make
sense to have a gem for CSS in API mode.

### DIFF
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -7,7 +7,9 @@ end
 
 ruby "<%= Suspenders::RUBY_VERSION %>"
 
+<% unless options[:api] %>
 gem "autoprefixer-rails"
+<% end %>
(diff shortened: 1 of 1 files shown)


templates/Gemfile.erb
```diff
@@ -7,7 +7,10 @@ end
 
 ruby "<%= Suspenders::RUBY_VERSION %>"
 
+<% unless options[:api] %>
 gem "autoprefixer-rails"
+<% end %>
+
```
tysongach: Maybe cut this empty line, so that on output there isn't an extra line between gems? We seem to do this on Line 27, as well, with the webpack option.

Issue #872, opened 2017-12-06
Heroku CI fails because DATABASE_URL is set


johannesschirrmeister:
Hi,

I just enabled Heroku CI for a project I recently created using suspenders. Unfortunately, I couldn't run the tests successfully due to the following line: `abort("DATABASE_URL environment variable is set") if ENV["DATABASE_URL"]` in https://github.com/thoughtbot/suspenders/blob/master/templates/rails_helper.rb#L4

It seems that the heroku-postgresql add-on is added to the Heroku CI app, causing the DATABASE_URL variable to be set. I'm not aware of any configuration changes I have made so I'm assuming the fact that this add-on is installed is a result of "normal" suspenders usage.

I'm puzzled that this hasn't been reported before as this line has been there for a while. Why shouldn't I use a DB provided by Heroku in my test environment? And why was it configured automatically anyways?

Cheers,
Johannes

PS: If nothing else comes out of this, I will propose a change to the abort message. It was very hard to spot "DATABASE_URL is set" as the error cause.

mike-burns:
From the PR that added it: https://github.com/thoughtbot/suspenders/pull/601#issuecomment-131845392

> Rails 5 is going to address this issue, but in the meantime we could raise if DATABASE_URL is set at all.

@derekprior can you speak more to how Rails 5 addresses this? Can we remove this hack?

derekprior:
I believe the comment was mistaken, unfortunately. 

Rails 5 added limited environment protection so you can't run `rake db:drop` or other destructive rake tasks without something equivalent to `ENABLE_FOOT_GUN=1 rake db:drop` (not the actual env variable, but you get the point).

My hope at the time of that comment was that the solution introduced in Rails 5 would also cover database cleaner, but it does not.

We could maintain the spirit of this check by asserting `ActiveRecord::InternalMetadata[:environment] == 'test'` instead of checking `database_url`, I think. it's also worth considering if we would need this protection at all if we got rid of database cleaner?

mike-burns:
Maybe we should revisit the standard `rails_helper.rb` from RSpec, e.g. https://github.com/rspec/rspec-rails/pull/1383

gerryhd:
So... was this fixed? Will this be fixed? Does anyone have a workaround, or is it safe to just comment out that line? It's odd to claim this works with Heroku out of the box but make CI tests fail due to that line in the rails_helper.rb file.

nickcharlton:
[...]

PR #979, opened 2019-05-29
Check for the environment, not DATABASE_URL

nickcharlton:
Suspenders was checking for the presence of `DATABASE_URL` to see if it
was running in production. This isn't so reliable in practice as
`DATABASE_URL` can be used in other environments. The spirit of this was
originally to stop you truncating a production database, which checking
for the Rails environment acheives.

https://github.com/thoughtbot/suspenders/issues/872
https://github.com/rspec/rspec-rails/pull/1383

### DIFF
--- a/templates/rails_helper.rb
+++ b/templates/rails_helper.rb
@@ -1,7 +1,10 @@
 ENV["RACK_ENV"] = "test"
 
 require File.expand_path("../../config/environment", __FILE__)
-abort("DATABASE_URL environment variable is set") if ENV["DATABASE_URL"]
+
+if Rails.env.production?
+  abort("The Rails environment is running in production mode!")
+end
 
 require "rspec/rails"
 


tute:
Hi @nickcharlton! 👋 

Would we stay protected from the potential risk of running tests (maybe with `guard`), and setting `DATABASE_URL` for other purposes, mistakenly wiping out production data? The previos fix in 2c069e1e took that situation into account (which I've seen happen 😱).

nickcharlton:
👋 

With the previous implementation, if `DATABASE_URL` was defined at all it would `abort`. This is what pushed me to change to the `rspec-rails` implementation on the project I was referring to: running on CI we were overriding `DATABASE_URL` to have the correct test database referenced.

If you'd set `DATABASE_URL` to your production database, we'd lose that protection with this change 😞.

The previous `abort` was a bit crude, so maybe instead we could adjust the messaging? I can't think of a way to mitigate both scenarios (not being able to use `DATABASE_URL` at all, which is at worst confusing and the problem where you could overwrite production which is dangerous).

Travis CI had this exact problem not long ago, too: https://blog.travis-ci.com/2018-04-03-incident-post-mortem

jdbann:
I'm in favour of the environment check, it's what rspec-rails includes in the generator:

https://github.com/rspec/rspec-rails/blob/30820f431526225d8d222b2c95b7ad9a691ff7b2/lib/generators/rspec/install/templates/spec/rails_helper.rb#L9-L10

The `DATABASE_URL` check seems to be more specific to a particular implementation.

PR #983, opened 2019-06-19
Set a bigger window for `save_and_open_screenshot`

gabebw:
1680x1050 is the size of a 15" retina Macbook Pro.

Without this setting, `save_and_open_screenshot` captures perhaps 75% of the screen, which makes it much less useful for debugging.

I'm sure some people have bigger screens, and could use an even larger window size, but this feels like a good default.

### DIFF
--- a/templates/chromedriver.rb
+++ b/templates/chromedriver.rb
@@ -7,6 +7,7 @@ end
 Capybara.register_driver :headless_chrome do |app|
   options = ::Selenium::WebDriver::Chrome::Options.new
   options.headless!
+  options.add_argument "--window-size=1680,1050"
 
   Capybara::Selenium::Driver.new app,
     browser: :chrome,


templates/chromedriver.rb
```diff
@@ -5,12 +5,18 @@
 end
 
 Capybara.register_driver :headless_chrome do |app|
+  options = ::Selenium::WebDriver::Chrome::Options.new
+  # Set a big (15" MBP retina resolution) window size so that
```
composerinteralia: suspenders is fairly [opinionated about removing configuration comments](https://github.com/thoughtbot/suspenders/blob/996fe4aa0d22931b6c3f5b01ab575ffd826c8746/lib/suspenders/app_builder.rb#L266-L292). How would you feel about removing this comment?

gabebw: Makes sense. Done in 598aa91f08874a7f9da32b25ae649b8fd1fefa2f.

composerinteralia:
Thanks!

PR #987, opened 2019-08-02
Update bitters to 2.0, bourbon to 6.0

tysongach:
- [Bitters' change log](https://github.com/thoughtbot/bitters/blob/master/CHANGELOG.md)
- [Bourbon's change log](https://github.com/thoughtbot/bourbon/blob/master/CHANGELOG.md#600---2019-07-10)

### DIFF
--- a/lib/suspenders/generators/stylesheet_base_generator.rb
+++ b/lib/suspenders/generators/stylesheet_base_generator.rb
@@ -3,7 +3,7 @@ require_relative "base"
 module Suspenders
   class StylesheetBaseGenerator < Generators::Base
     def add_stylesheet_gems
-      gem "bourbon", ">= 5.0.1"
+      gem "bourbon", ">= 6.0.0"
       Bundler.with_clean_env { run "bundle install" }
     end
 
--- a/suspenders.gemspec
+++ b/suspenders.gemspec
@@ -28,7 +28,7 @@ rush to build something amazing; don't use it if you like missing deadlines.
   s.test_files = `git ls-files -- {test,spec,features}/*`.split("\n")
   s.version = Suspenders::VERSION
 
-  s.add_dependency 'bitters', '~> 1.7'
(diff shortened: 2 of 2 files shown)


composerinteralia:
Seeing this on CI:

```
Bundler::GemNotFound: You have requested:
  bourbon >= 6.0.0

The bundle currently has bourbon locked at 5.1.0.

Try running `bundle update bourbon`

If you are updating multiple gems in your Gemfile at once,
try passing them all to `bundle update`
```

Adding `bourbon >= 6.0.0` to the project Gemfile might make that go away.

tysongach:
@composerinteralia Interesting! So, add it to `suspenders.gemspec` like we do with Bitters?

https://github.com/thoughtbot/suspenders/blob/dc79cf034b1dd992dea96557b922a3e95192e30b/suspenders.gemspec#L31

composerinteralia:
🤷‍♂ it is worth a try. I don't fully understand why, but similar things have worked on the factory_bot_rails test suite.

composerinteralia:
@tysongach I rebased this branch on master to try out the fix I was thinking of. It didn't work because bitters has a [pessimistic version constraint for bourbon](https://github.com/thoughtbot/bitters/blob/master/bitters.gemspec#L9). We will need to bump that version and release a new version of bitters to get bourbon 6 working with suspenders. I would recommend a non-pessimistic version constraint instead (">=" instead of "~>") so this doesn't happen again in the future.

We also have a [pessimistic version constraint on bitters](https://github.com/thoughtbot/suspenders/blob/master/suspenders.gemspec#L31) here in suspenders. I will change that in a separate PR.

tysongach:
@composerinteralia Ah, good catch! Thanks for those details. I can help move this forward, perhaps next week. Bourbon 6 is not a huge release, so I'm in no hurry here.

PR #989, opened 2019-08-19
Bump Rails version to 6.0.0

composerinteralia:
In addition to bumping the Rails version, this commit makes a couple JS
changes to support Rails using webpacker by default.

The remaining changes were to get CI to pass:

- Explicitly adding the travis postgresql service might not have solved
the test failures I was having, but it did make some warnings go away.
- I don't fully understand why we need to add rack_timeout as a
development dependency and not any of the other gems, but it made Travis
happy
- I also don't fully understand why I needed to move the
`suspenders:views` generator last. The tests passed locally without
that change, and I was able to generate a working application regardless
of the order. But on CI, the generators that ran just after
`suspenders:views` were failing with errors about `flashes_helper.rb`
not existing. There is something fishy going on there, and I spent a lot
of time debugging without success. I am suspicious of bootsnap and
zeitwerk, both of which appeared in the stack trace. Since the order of
this generator doesn't matter much, I would like to go forward with this
workaround. I may come back and debug more later to see if there is a
bug somewhere.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,5 @@ before_install:
 install: bundle install
 notifications:
   email: false
+services:
+  - postgresql
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -185,7 +185,6 @@ module Suspenders
       generate("suspenders:lint")
       generate("suspenders:jobs")
       generate("suspenders:analytics")
-      generate("suspenders:views")
     end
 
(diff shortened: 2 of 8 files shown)


composerinteralia:
Hm, for some reason I am hitting `ForceTlsGenerator#config_enforce_ssl` locally, but not on CI

composerinteralia:
It boots on my machine, but I hadn't tried pushing to Heroku. I will do that before merging.

PR #998, opened 2019-08-25
Use spring-watcher-listen gem

### LINKED ISSUE #950: Missing spring-watcher-listen?
I noticed that Suspenders doesn't include [spring-watcher-listen](https://github.com/jonleighton/spring-watcher-listen), which has been part of the default app Gemfile since Rails 5. Is that intentional?

composerinteralia:
Closes #950

Back in 2014 the [spring listen watcher was extracted into a gem][b614].
Since that time, the listen gem has been in our Gemfile not doing
anything. Both spring and listen are dependencies of
spring-watcher-listen, so we can replace them both.

[b614]: https://github.com/rails/spring/commit/b614d94e83d863ffcb3eec06c6b2bb9250d5b5a5

### DIFF
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -25,9 +25,8 @@ gem "tzinfo-data", platforms: [:mingw, :x64_mingw, :mswin, :jruby]
 gem "webpacker"
 
 group :development do
-  gem "listen"
   gem "rack-mini-profiler", require: false
-  gem "spring"
+  gem "spring-watcher-listen"
   gem "web-console"
 end
 


templates/Gemfile.erb
```diff
@@ -25,9 +25,8 @@ gem "tzinfo-data", platforms: [:mingw, :x64_mingw, :mswin, :jruby]
 gem "webpacker"
 
 group :development do
-  gem "listen"
   gem "rack-mini-profiler", require: false
-  gem "spring"
```
andyw8: I think the `spring` entry shouldn't have been removed here.

If I created a project using Suspenders, and then chose to remove `spring-watcher-listen` from my Gemfile, it would would cause Spring to removed, even thought that wasn't the intention.

composerinteralia: That makes sense to me. We can add it back.

PR #1001, opened 2019-10-17
Remove unnecessary dev dependency

delphaber:
Travis was complaining about missing `rack-timeout` gem because, even if
we defined the `rack-timeout` as a `production` only gem, when we run
`bundle install` we install all of the gems, regardless of the
environment.

I added the `bundle install` command in the timeout generator, so this
should remove the need of a dev dependency on `rack-timeout`


mike-burns:
Nice find with that `bundle install`. Thank you!

PR #1004, opened 2019-10-20
Copy auto migrate into project

delphaber:
I noticed that `bin_auto_migrate` was not copied into project directory. I think it must be present so heroku will call it in its `release` script


mike-burns:
Seems right, thanks for catching this.

PR #1007, opened 2019-11-03
Keep using sprockets v3

delphaber:
This commit should fix the CI on master.

I don't know if in thoughtbot you already use sprockets v4. We do not, so we fixed the version of sprockets to v3. Even Rails 6 does not support sprockets v4 right now 🤔 

From a plain new `rails new blog` application with rails 6

```
    sass-rails (5.1.0)
      railties (>= 5.2.0)
      sass (~> 3.1)
      sprockets (>= 2.8, < 4.0)
      sprockets-rails (>= 2.0, < 4.0)
      tilt (>= 1.1, < 3)
```

### DIFF
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -19,7 +19,7 @@ gem "rails", "<%= Suspenders::RAILS_VERSION %>"
 gem "recipient_interceptor"
 gem "sassc-rails"
 gem "skylight"
-gem "sprockets", ">= 3.0.0"
+gem "sprockets", "< 4"
 gem "title"
 gem "tzinfo-data", platforms: [:mingw, :x64_mingw, :mswin, :jruby]
 gem "webpacker"


composerinteralia:
Thanks! I am OK merging this, but I wonder if the sprockets version is something we will end up fighting with again later on.

It looks like Rails 6.0.1 will [default to using sass-rails > 6](https://github.com/rails/rails/blob/2a3f759eef10352bedce5f13b12dbdda30aacab2/railties/lib/rails/generators/app_base.rb#L309). sass-rails 6 [no longer depends on Sprockets at all](https://github.com/rails/sass-rails/commit/706526d411d2d78e578c7944e873ece28e65ed42#diff-4066928b9297aa27eaa7097204205e9fL18-L19).  Instead, it is simply a wrapper over `sassc-rails`. I would be curious to see what the Gemfile looks with `rails new blog` on 6.0.1.rc1.

We have sassc-rails in our suspenders Gemfile (instead of the sass-rails wrapper over it), and the latest version has a [runtime dependency on Sprockets > 3](https://github.com/sass/sassc-rails/blob/master/sassc-rails.gemspec#L30). Ideally we wouldn't need to have an entry for Sprockets in our suspenders Gemfile at all.

composerinteralia:
Update: Rails 6.0.1 is out and the default `rails new` Gemfile.lock does indeed end up with a `> 3` Sprockets dependency:

```rb
    sassc-rails (2.1.2)
      railties (>= 4.0.0)
      sassc (>= 2.0)
      sprockets (> 3.0)
      sprockets-rails
      tilt
```

So I don't locking Sprockets to a lower version is a great solution here. It will make it more difficult to upgrade to Rails 6.0.1.

delphaber:
Interesting 🤔  Then I should investigate more on the problem without locking the sprockets version

composerinteralia:
I'm going to merge this to unblock you on your other open PRs, and also open an issue to address the Sprockets version.

PR #1011, opened 2019-11-25
Fix failing tests

mike-burns:
We have two variety of fixes here: Heroku migration, and JSON subset
matcher.

The Heroku migration improvements from 5b68a176cbd40456e80ea9b91a98597cd91b4d8d
modified the `circle.yml` and `bin/deploy` files, but three tests had
not caught up.The `circle.yml` checks can be removed; that's no longer
done. Any reference to `heroku run rails db:migrate` can be removed;
that's handled for us automatically.

The `contain_json` matcher used `Hash#<=` to find the sub-hash; however,
this does not recur. Instead we can write our own recursive `subhash?`
helper method.

### DIFF
--- a/spec/features/heroku_spec.rb
+++ b/spec/features/heroku_spec.rb
@@ -36,17 +36,6 @@ RSpec.describe "Heroku" do
 
       expect(readme).to include("./bin/deploy staging")
       expect(readme).to include("./bin/deploy production")
-
-      circle_yml_path = "#{project_path}/circle.yml"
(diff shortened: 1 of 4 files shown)


seanpdoyle:
Should we close https://github.com/thoughtbot/suspenders/pull/1002#issuecomment-557616509 in favor this branch?

mike-burns:
Oh shoot, I should have looked before diving in! I'll give them co-authorship credit.

PR #1018, opened 2019-12-03
Introduce suspenders:profiler

mike-burns:
This adds the `suspenders:profiler` generator, which is a
re-implementation of #678 (005e73f2) via generator instead of hardcoded.
While here, update the generated README as suggested by Sean elsewhere.

Let's talk about this:

```ruby
def augment_default_env
  append_to_file ".env", "RACK_MINI_PROFILER=0"
rescue Errno::ENOENT
    create_file ".env", "RACK_MINI_PROFILER=0"
rescue Thor::Error => e
  if e.message =~ /does not appear to exist/
    create_file ".env", "RACK_MINI_PROFILER=0"
  else
    raise
  end
end
```

Currently Suspenders will install a `.env` in every project (this was
done in #563). We'd like this `.env` file to set `RACK_MINI_PROFILER`.
However, this generator could potentially be run on a fresh Rails app,
which encourages a `.sample.env` instead of `.env`.

On `rails generate suspenders:profiler`:

1. Attempt to append to `.env`.
2. If that file does not exist, [Thor will raise a `Thor::Error`] with
   the message `The file .env does not appear to exist`.
3. On that specific failure, create the file instead.
4. On any other failure, raise.

[Thor will raise a `Thor::Error`]: https://github.com/erikhuda/thor/blob/8d15101e3d68f929ec2f4441ac2dfdf7ca11eeea/lib/thor/actions/inject_into_file.rb#L67

On `rails destroy suspenders:profiler`:

1. Attempt to remove from `.env`.
2. If that fails, [`File.read` will raise] `Errno::ENOENT`.
3. On that failure, delete the file instead.

[`File.read` will raise]: https://github.com/erikhuda/thor/blob/8d15101e3d68f929ec2f4441ac2dfdf7ca11eeea/lib/thor/actions/inject_into_file.rb#L110

We could check for the file explicitly, but that would introduce a
classic race.


mike-burns:
That's a complicated question!

It is true that this generator itself is opt in. However, Suspenders as a whole is unchanged: when you create an app using Suspenders, it will run this generator for you, along with all the others. So in a sense, nothing has changed.

If you'd like, an additional argument for leaving it at `0` is because not every team member -- designers, for example -- will care about the DB performance of every page, at all times.

Switching it to default to `1` would be implying that we need to care more about performance, and that running with this profiler always on will help with that. Is that an accurate statement?

PR #1019, opened 2019-12-04
Introduce suspenders:runner

mike-burns:
This introduces the `suspenders:runner` generator. It does everything
needed to make the app into something you can run locally.

This affects four files:

- `Procfile`. Our Procfile runs Puma for the Web and ActiveJob for the
workers.
- `README.md`. This appends documentation on how to configure global
settings and how to run the app locally.
- `.sample.env`. This is where defaults go, which should be copied to
`.env`.
- `bin/setup`. Offer to copy the `.sample.env` to `.env` as part of the
setup, but not on CI.

Drop dotenv in place of Foreman. We always need to run both the server
and the workers, so depend on Foreman for this behavior. If someone
really needs to do otherwise they can `source .env && rails s`.

 ## `Procfile`

This does not intentionally do anything about running the app in
production. It depends on the `config/puma.rb` that ships with Rails,
which is sufficient. Handling the `puma.rb` that is appropriate for
production should be done in a Heroku-specific
`suspenders:production:...` generator.

 ## `README.md`

The text for the README is adapted from an existing project which,
ironically, does not use Heroku. It recommends `heroku local`, which is
implemented in terms of node-foreman, but a regular `foreman start`
would also work. Since we remove dotenv, `rails s` will no longer work.

 ## `.sample.env`

Switch from `.env` to `.sample.env`, which is a reverse of what was
introduced in #563 (8739344046cfbca5b74dcb4d9aba3c6e9c805463). We no
longer use dotenv, instead depending on Foreman or node-foreman (via
`heroku local`). The node-forman documentation [explictly says] not to
check `.env` into version control.

[explictly says]: https://github.com/strongloop/node-foreman#best-practices

In addition, many coworkers, new and old, had assumed that we always use
a `.sample.env` that we copy to `.env` and were surprised to discover
that Suspenders does it differently.

 ## `bin/setup`

Thus the `bin/setup` script copies `.sample.env` to `.env`. It copies
interactively (`-i`) to avoid overwriting an existing `.env` file.

Because this is a stand-alone generator it must now work for both the
default Rails setup script and the Suspenders one. We do this naïvely:
if it's Ruby, assume it's the Rails script; if it's shell and it
mentions CI, assume it's ours; otherwise shove it on the end.


lib/suspenders/generators/runner_generator.rb
```diff
@@ -0,0 +1,75 @@
...
+        insert_into_file "bin/setup", %{  cp -i .sample.env .env\n},
+          after: %{if [ -z "$CI" ]; then\n}
+      else
+        append_to_file "bin/setup",
+          %{\nif [ -z "$CI" ]; then\n  cp -i .sample.env .env\nfi}
+      end
+    end
+
+    def update_readme
+      configure = <<~README
```
seanpdoyle: I _think_ GitHub will syntax highlight Ruby heredocs delimited by language names.

Are you willing to try that out? 

I think that would involve using `MD` or `MARKDOWN` instead of `README`

lib/suspenders/generators/runner_generator.rb
```diff
@@ -0,0 +1,75 @@
...
+Use the `heroku local` runner to run the app locally as it would run on Heroku.
+This uses the node-forman runner, which reads from the `Procfile` file.
+
+```sh
+heroku local
+```
+
+Once the server is started the application is reachable at
+`http://localhost:3000`.
+      README
```
jsilasbailey: This is not necessarily for the scope of this change, just a question I had while reading it. Would it be valuable to have a way to pull these sections from markdown "partial" files in the repo instead of having to type out long heredocs?

mike-burns: You are definitely correct, that is the right thing to do here.

We have seven generators (including this one) with heredoc in it. I'll make a separate PR for moving them into the `templates/` directory.

lib/suspenders/generators/runner_generator.rb
mike-burns: Tried `MARKDOWN`, but no dice. That said, it's a better name.

lib/suspenders/generators/runner_generator.rb
seanpdoyle: That's a great idea! It addresses my underlying concerns that brought about my suggestion to delimit with `MARKDOWN` instead of `README`.

PR #1029, opened 2020-03-13
Document the generator goals

mike-burns:
This is something we've been working toward for half a decade now, so
it's best to document our intentions here for other contributors and our
own direction.

### DIFF
--- a/GOALS.md
+++ b/GOALS.md
@@ -1,0 +1,65 @@
+# Goals
+
+Suspenders is an opinionated platform for adding functionality to a Rails app.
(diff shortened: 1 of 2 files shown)


GOALS.md
```diff
@@ -0,0 +1,65 @@
+# Goals
+
+Suspenders is a opinionated platform for adding functionality to a Rails app.
```
thorncp:
```suggestion
Suspenders is an opinionated platform for adding functionality to a Rails app.
```

PR #1031, opened 2020-04-06
Add suspenders to rails app earlier

composerinteralia:
We are seeing failures on CI for `Bundler could not find compatible
versions for gem "thor"`.

When our tests run `rails new`, that bundle installs inside the new
Rails application with thor constrained to ">= 0.20.3", "<2.0".
This started resolving to thor 1.0.0 when that version came out last
December (which is when our test suite started failing).
2019.

After running rails new, we add suspenders to the Gemfile of the new
rails application. suspenders depends on bitters, which depends on thor
"~> 0.19". This is incompatible with thor 1.0.0, which by this point is
already in the Gemfile.lock

This commit gets around the problem by adding suspenders to the Gemfile
via a template so it is there before the `bundle install`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 rvm: 2.6.5
 cache: bundler
 before_install:
-  - nvm install 8.16.0
+  - nvm install 13.12.0
   - git config --global user.name 'Travis CI'
   - git config --global user.email 'travis-ci@example.com'
   - gem update --system
--- a/spec/support/rails_template.rb
+++ b/spec/support/rails_template.rb
@@ -1,0 +1,1 @@
+gem "suspenders", path: File.expand_path("../..", __dir__)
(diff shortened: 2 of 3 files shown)


composerinteralia:
Progress! It looks like the one remaining failure is because we are getting a node version that is too old for stylelint.

mike-burns:
Does this mean any user of command-line suspenders will also need this template?

composerinteralia:
Running `suspenders my_app` currently works fine.

Running `rails new my_app`, then adding suspenders to the Gemfile and running `bundle install` currently fails, but the user would be able to resolve that by running `bundle update thor`. Once we cut a release of bitters they won't need to run `bundle update thor` anymore.

I wouldn't expect people to use this template. I thought for our tests it made sense to use the template rather than adding an extra step to `bundle update`, since it is a little quicker.

mike-burns:
Aha, great solution.

Does this mean that we cant get rid of this diff after bitters is fixed?

composerinteralia:
We could (except the node update), but I don't know if we should. Keeping this change may make our tests less flaky. Someday when thor 2 comes out this is likely to happen again, and I don't think it makes sense to get tests failures here in suspenders because of a dependency problem in bitters.

mike-burns:
Oooh! I see what you did. This is a great solution, thank you!

PR #1034, opened 2020-04-26
Remove Gemfile constraint on sprockets

### LINKED ISSUE #1009: Remove '< 3' constraint on Sprockets
We locked Sprockets to `< 3` in https://github.com/thoughtbot/suspenders/pull/1007 to get around some CI failures, but this constraint may make it more difficult to upgrade Rails.

composerinteralia:
Closes #1009

We had locked the version of Sprockets to get around some test failures.
If those test failures are no longer around, we should not lock this
version.


composerinteralia:
Hm, these seem to be the same failures and I am not planning to debug at the moment.

PR #1036, opened 2020-05-01
Finish .sample.env migration

mike-burns:
In a056e1a85398e7140ce60b7c14721f9149cfa2b6 we migrated from `.env` to
`.sample.env`. Some stray `.env` files were left behind; fix those.

### DIFF
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -139,6 +139,7 @@ module Suspenders
 
     def generate_default
       run("spring stop")
+      generate("suspenders:runner")
       generate("suspenders:profiler")
       generate("suspenders:json")
       generate("suspenders:static")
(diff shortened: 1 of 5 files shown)


mike-burns:
Yup, this is why CI is failing.

mike-burns:
Fixing this is now at the bottom of my to-do stack (at the top was #1022, but now that's fixed). Very clearly something in the test run is depending on that `.env` file existing.

composerinteralia:
I think I know why this is failing. With this change the `ProfilerGenerator` now creates the `.sample.env`. `RunnerGenerator` (which comes after `ProfileGenerator`) then gets hung up waiting for confirmation whether to overwrite the existing `.sample.env`.

Perhaps running the `RunnerGenerator` before the `ProfilerGenerator` will fix this.

mike-burns:
Is there a flag we can pass when we run the tests that is like the "always overwrite" flag?

composerinteralia:
We could run generators with a `--force` or `--skip` flag, but that might hide real issues. In this case the ordering of the generators didn't really make sense, and the test failures helped us fix that. What I would like is a way to run the generators and fail if there are any conflicting files, but that option doesn't seem to exist.

mike-burns:
Yeah makes sense. LGTM.

PR #1043, opened 2020-06-24
Add standard for linting

composerinteralia:
Add standard for linting
===

https://github.com/testdouble/standard

thoughtbot is moving toward standard instead of Hound + Rubocop:
https://github.com/thoughtbot/guides/pull/606

This commit adds the standard gem and adds standard to the default rake
task

Apply standard fixes
===

This commit applies the changes after running `rake standard:fix`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
-source 'https://rubygems.org'
+source "https://rubygems.org"
 
 gemspec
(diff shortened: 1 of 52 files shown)


composerinteralia:
Yeah, I knew Hound would be upset about this one. We will need to disable Hound when this PR gets merged.

PR #1044, opened 2020-06-26
Fix ActionMailer asset host in production.

georgebrock:
This change was [originally made in a suspenders-based app to fix a problem
we were having with images not rendering in emails][1]. This commit replicates
the fix up-stream for future suspenders users.

The change fixes image URLs in emails by adding a protocol to the
ActionMailer asset_host setting.

While the setting is called "host", based on the examples in the Rails
documentation, it could be more accurately described as a "URL-prefix" --
all of the examples include a protocol as well as a host.

Prior to this change, image URLs were generated in emails from production
environments with a protocol-relative URL (e.g. //example.com/image.png),
and were not rendering correctly in the various mail clients we tested.

[1]: https://github.com/thoughtbot/albaik-web/pull/609

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -170,7 +170,7 @@ module Suspenders
       action_mailer_host "production", %{ENV.fetch("APPLICATION_HOST")}
       action_mailer_asset_host(
         "production",
-        %{ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))}
+        %q{"https://#{ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))}"}
       )
     end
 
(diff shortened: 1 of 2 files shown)


lib/suspenders/app_builder.rb
```diff
@@ -170,7 +170,7 @@ def configure_action_mailer
       action_mailer_host "production", %{ENV.fetch("APPLICATION_HOST")}
       action_mailer_asset_host(
         "production",
-        %{ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))},
+        %q{"https://#{ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))}"}
```
georgebrock, line 173: I think TLS in production is a reasonable assumption in 2020. I couldn't see anywhere else where we were explicitly specifying a protocol in the generated app code.

thiagoa:
@georgebrock Thank you! :pray: Sorry that it took us so long to merge your PR.

PR #1048, opened 2020-07-10
Add standard to lint generator

composerinteralia:
This commit adds the [standard] gem, and requires the standard rake task
in the Rakefile.

It also disables RuboCop on Hound, since standard is meant to replace
RuboCop.

[standard]: https://github.com/testdouble/standard

Co-authored-by: Daniel Colson <danieljamescolson@gmail.com>

### DIFF
--- a/spec/features/lint_spec.rb
+++ b/spec/features/lint_spec.rb
@@ -1,0 +1,26 @@
+require "spec_helper"
+
+RSpec.describe "suspenders:lint", type: :generator do
+  it "sets up standard" do
+    with_app { generate("suspenders:lint") }
+
+    expect("Gemfile").to match_contents(/standard/)
+
+    run_in_project do
+      expect(`rake -T`).to include("rake standard")
+    end
+  end
+
+  it "removes standard" do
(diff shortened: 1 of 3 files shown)


spec/features/lint_spec.rb
```diff
@@ -0,0 +1,26 @@
+require "spec_helper"
+
+RSpec.describe "suspenders:lint", type: :generator do
+  it "sets up standard" do
+    with_app { generate("suspenders:lint") }
+
+    expect("Gemfile").to match_contents(/standard/)
+
+    run_in_project do
+      expect(`rake -T`).to include("rake standard")
```
mike-burns, line 10: Oh fancy!

mike-burns:
Hey @mike-burns of the past, you were supposed to notice that this diff doesn't run `bundle install` after modifying the `Gemfile`! Some reviewer you were ...

Apologies to @thorncp and @composerinteralia for not noticing that. It's a real subtle thing that only impacts people who run this generator in isolation (which I just did).

PR #1052, opened 2020-08-09
upgrade ruby to 2.7.1

dorianmariefr:
that's the latest stable version


composerinteralia:
Thanks for the PR. We had been holding off on this because of the Ruby 2.7 kwarg deprecation warnings. The situation is much better now so I think it is a good time to revisit.

composerinteralia:
Merged as https://github.com/thoughtbot/suspenders/commit/c82a3c994e02b59ace14f0d6e768e1acd94cb823. Thank you!

Issue #1053, opened 2020-08-09
"trueconfig" error in config/environments/development.rb


dorianmariefr:
when generating a new project, `config/environments/development.rb` has:

```ruby
  config.active_storage.service = :local
  config.action_mailer.raise_delivery_errors = trueconfig.after_initialize do
  Bullet.enable = true
  Bullet.bullet_logger = true
  Bullet.rails_logger = true
end
  config.action_mailer.delivery_method = :file
```

error is:

```
undefined local variable or method `trueconfig' for #<Hivebrite::Application:0x00007febed6c92e8> (NameError)
```

i think a diff for the fix might be something like:

```diff
--- a/lib/suspenders/generators/db_optimizations_generator.rb
+++ b/lib/suspenders/generators/db_optimizations_generator.rb
@@ -11,7 +11,7 @@ module Suspenders
       inject_template_into_file(
         "config/environments/development.rb",
         "partials/db_optimizations_configuration.rb",
-        after: /config.action_mailer.raise_delivery_errors = .*/
+        after: /config.action_mailer.raise_delivery_errors = .*\n/m
       )
     end
   end
```

composerinteralia:
Thanks for the report. What version of suspenders are you on? I was under the impression we had fixed this in https://github.com/thoughtbot/suspenders/commit/a8173cd767656c4a240edd9fd1d780f239789c54.

kaka-ruto:
I just got the same error too. I am on suspenders `v1.55.0`

kaka-ruto:
@composerinteralia, @dorianmariefr ,have a quick work around for this?

composerinteralia:
The workaround for now is to manually go in and fix the spacing in the file (i.e. add a newline between `true` and `config`).

kaka-ruto:
Thanks @composerinteralia . And then run `bin/setup` to continue installation? Sorry, it's my first time using suspenders :)

composerinteralia:
Yeah, running bin/setup makes sense as a next step. Also, if any of the default generators failed you may need to re-run some of them. You can get the list of generators with `bin/rails generate | grep suspenders` and get more info about a specific generator with `bin/rails generate suspenders:<generator name> -h`

kaka-ruto:
Thanks, yup, some of the generators failed. I'll follow along to find the failed generators, these will be helpful

dorianmariefr:
[...]

PR #1054, opened 2020-08-12
fix broken development config

dorianmariefr:
i'm still getting a bootsnap related error when running locally and a puma compile error when running the tests locally so i'm making a pull requests to see if the tests pass on travis

see #1053

### DIFF
--- a/spec/features/new_project_spec.rb
+++ b/spec/features/new_project_spec.rb
@@ -190,6 +190,11 @@ RSpec.describe "Suspend a new project with default configuration" do
     )
   end
 
+  it "does not make invalid config because of bullet config in development" do
+    expect(development_config).to match(/^ +Bullet.enable = true$/)
+    expect(development_config).to_not match(/trueconfig/)
+  end
+
   it "configs bullet gem in development" do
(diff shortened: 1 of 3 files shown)


spec/features/new_project_spec.rb
```diff
@@ -190,6 +190,11 @@
     )
   end
 
+  it "does not make invalid config because of bullet config in development" do
+    expect(development_config).to match(/^ +Bullet.enable = true$/)
+    expect(development_config).to_not match(/trueconfig/)
```
composerinteralia, line 195: Could we add this expectation into the spec below rather than introducing a new one? Normally I would prefer it as a separate spec the way you have done it, but these tests are really slow because they run the whole suspenders command as setup.

composerinteralia:
I would be OK adding a [standard.yml](https://github.com/testdouble/standard#what-you-might-do-if-youre-clever) and ignoring this one file

spec/features/new_project_spec.rb
dorianmariefr: ok, i merged the two tests

dorianmariefr:
good idea for the .standard.yml file, i though standard had no config file (should have checked)

Issue #1055, opened 2020-09-13
Replace `git` command in gemspec


utkarsh2102:
Hello @composerinteralia 👋🏻 

Thanks for your great work here and elsewhere! ❤️ 

Whilst maintaining this in Debian, I found that this library relies on `git` to produce a list of files.
https://github.com/thoughtbot/suspenders/blob/f16e7439f42acc508e297b32e79f551b11e995c2/suspenders.gemspec#L20

Using `git` in gemspec is a bit problematic and it's best if avoided. More about this can be found here:
https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#gemspec-git-rationale

I'd be happy to submit a PR if you concur with this?

composerinteralia:
Thanks for the kind words. Yeah, I'd be open to moving away from `git ls-file`. We use `Dir.glob` over on [factory_bot](https://github.com/thoughtbot/factory_bot/blob/master/factory_bot.gemspec#L13-L15), so I assume we could do something similar here.

PR #1057, opened 2020-09-26
Drop git in gemspec

utkarsh2102:
Hello @composerinteralia 👋🏻 

Once again, thanks for your work! 🚀 
As discussed in #1055, here's a PR fixing the same! This also helps us in maintaining `suspenders` in Debian! 😄

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,18 @@
+require:
+  - rubocop-packaging
+
+AllCops:
+  TargetRubyVersion: 2.4
+  DisabledByDefault: true
+
+Packaging/BundlerSetupInTests:
+  Enabled: true
+
+Packaging/GemspecGit:
+  Enabled: true
+
+Packaging/RequireHardcodingLib:
+  Enabled: true
+
+Packaging/RequireRelativeHardcodingLib:
+  Enabled: true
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,18 @@
+require:
```
composerinteralia, line 1: We switched from RuboCop to [standard](https://github.com/testdouble/standard) recently. Given the no-configuration philosophy of standard, I would prefer to leave this out. I think the gemspec is unlikely to change frequently anyway, so leaving off the rule seems OK.

suspenders.gemspec
```diff
@@ -17,19 +17,20 @@ Gem::Specification.new do |s|
...
   s.summary = "Generate a Rails app using thoughtbot's best practices."
-  s.test_files = `git ls-files -- {test,spec,features}/*`.split("\n")
+  s.test_files = Dir["spec/**/*"]
   s.version = Suspenders::VERSION
 
   s.add_dependency "bitters", ">= 2.0.4"
   s.add_dependency "rails", Suspenders::RAILS_VERSION
 
   s.add_development_dependency "rspec", "~> 3.2"
+  s.add_development_dependency "rubocop-packaging", "~> 0.5"
```
composerinteralia, line 34: We would want to remove this as well.

thiagoa:
@utkarsh2102 Thank you for all the work you've done here!

Because this PR is stale I will close it in favor of https://github.com/thoughtbot/suspenders/pull/1092

PR #1058, opened 2020-10-23
Add Dockerfile

jferris:
This adds a Dockerfile that can serve as a reference point for
Suspenders-like applications which will be deployed to OCI platforms
like Docker, Kubernetes, and ECS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN gem install bundler \
 
 COPY . /app
 
-# The SECRET_KEY_BASE here isn't used. Precomiling assets doesn't use your
+# The SECRET_KEY_BASE here isn't used. Precompiling assets doesn't use your
 # secret key, but Rails will fail to initialize if it isn't set.
 RUN RAILS_ENV=production \
   PRECOMPILE=true \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,65 @@
+FROM ruby:2.7.1-slim AS builder
+
+ENV LANG en_US.UTF-8
+
+RUN apt-get update -qq \
+  && apt-get install -y \
+  apt-transport-https \
+  build-essential \
+  ca-certificates \
+  curl \
+  git \
+  libpq-dev \
+  && curl -sL https://deb.nodesource.com/setup_10.x | bash \
+  && apt-get update -qq \
+  && apt-get install -y nodejs \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN mkdir /app
+WORKDIR /app
+
+COPY Gemfile* /app/
+RUN gem install bundler \
+  && bundle config --global frozen 1 \
+  && bundle config --global without "development test" \
+  && bundle install -j4 --retry 3 \
+  && rm -rf /usr/local/bundle/cache/*.gem \
+  && find /usr/local/bundle/gems/ -name "*.c" -delete \
+  && find /usr/local/bundle/gems/ -name "*.o" -delete
+
+COPY . /app
+
+# The SECRET_KEY_BASE here isn't used. Precomiling assets doesn't use your
+# secret key, but Rails will fail to initialize if it isn't set.
+RUN RAILS_ENV=production \
+  PRECOMPILE=true \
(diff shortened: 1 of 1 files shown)


mike-burns:
Should this be in a generator? I don't actually understand what this file is doing.

jferris:
> Should this be in a generator?

I'm not sure where it should be, but I'm guessing that if it were in a generator, nobody would ever notice it again.

> I don't actually understand what this file is doing.

Having this file present means that running `docker build` or pushing to a pipeline that supports Docker will result in having a Docker image. It's similar to a Heroku buildpack, which tells Heroku how to make your application into a slug, whereas this tells a Docker-compatible builder how to make your application into a container image.

I think my preference would be to have it new repositories, similar to having `.tool-versions`. If you don't use ASDF or anything, that file does nothing. A `Dockerfile` is the same.

More than anything, I want to have a place where we document and collaborate on a good stock `Dockerfile` for Suspenders projects, and Suspenders seems like a good place to start.

mike-burns:
Oooh. I read the PR description as if you were adding this as a sample Dockerfile for projects, which is confusing because that's the point of Suspenders.

But you're adding this Dockerfile so you can use Docker with Suspenders? That makes much more sense.

mike-burns:
Anyway, my feelings on Docker is: does this have to live in the repo? Can we have a separate repo of Dockerfiles for the people who do use Docker?

Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+  && bundle config --global frozen 1 \
+  && bundle config --global without "development test" \
+  && bundle install -j4 --retry 3 \
+  && rm -rf /usr/local/bundle/cache/*.gem \
+  && find /usr/local/bundle/gems/ -name "*.c" -delete \
+  && find /usr/local/bundle/gems/ -name "*.o" -delete
+
+COPY . /app
+
+# The SECRET_KEY_BASE here isn't used. Precomiling assets doesn't use your
```
johnpaulashenfelter, line 33: Type: "... Precompiling"

jferris: Fixed in 1876ed3.

thiagoa:
I like this idea. As @mike-burns noted, this could become part of some deployment generator: we'd move the Heroku logic into a separate generator and the `Dockerfile` would be part of another generator, yet to be named. That would allow us to choose Heroku or something else if that's what we want.

However, as the `Dockerfile` is at the root of `suspenders`, how would it get copied to actual applications? Should we place this file in `templates` and add a build instruction to copy it over to the root of the generated application? The template could have an `.erb` extension that would allow us to inject the Ruby version used by the client (probably the `RUBY_VERSION` constant), but then there's the question: if we inject the Ruby version, would a Docker image always exist for it? Currently, the Ruby version is hardcoded in Suspenders, but I'm working on making it flexible: https://github.com/thoughtbot/suspenders/issues/1073 

Since we don't have a defined idea of what generator we will use for the `Dockerfile`, I think we could simply go with this older-school solution. Being in a generator would still allow us to run that generator by default when running `bin/suspenders`.

stevepolitodesign:
Would [MRSK](https://github.com/mrsked/mrsk) solve for this?

jferris:
> Would [MRSK](https://github.com/mrsked/mrsk) solve for this?

No, I don't believe so. I haven't used MRSK but it looks like an orchestration platform for Docker containers. You'd still need a Dockerfile (or something else) to containerize the application.

stevepolitodesign:
Actually, it looks like Rails [will ship with a Dockerfile](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt) in the next release?

jferris:
If that's the case, will Suspenders automatically get that when we update it for that version of Rails?

mike-burns:
Yup!

mike-burns:
That said, if we want a different one, we should add a `production:deploy` generator that generates the Dockerfile we'd prefer.

jferris:
I looked at the template for the Dockerfile in Rails and it seems like a good starting place for most applications. We might have to make changes for specific deployment platforms, but it's enough to get us going.

I'm going to close this pull request.

PR #1062, opened 2021-02-12
Use GitHub Actions for CI

mike-burns:
Two commits here: the first one fixes some thing around Ruby and Rails versions. This gets the test suite passing on my computer.

But secondly: switch to GitHub Actions for CI.

This GitHub Action [runs] when a pull request is opened, re-opened, or
[synchronized], [ignoring] any PR that only touches docs. It also can be
run [manually].

It uses the [setup-ruby] action provided by Ruby, tracking releases. The
only interesting bit here is that it [caches] bundler artifacts.

[caches]: https://github.com/ruby/setup-ruby/blob/master/README.md#caching-bundle-install-automatically
[setup-ruby]: https://github.com/ruby/setup-ruby
[ignoring]: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths
[manually]: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
[runs]: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
[synchronized]: https://github.community/t/what-is-a-pull-request-synchronize-event/14784/2


mike-burns:
I'm not a doctor, but I think I need to merge this for it to take effect. Going to merge it.

purinkle:
What was the thinking in choosing GitHub Actions over CircleCI?

I know that we would normally default to CircleCI.

I'm interested to know why we wouldn't do the same here.

mike-burns:
I'm inspired by discussions I had seen in #dev about GitHub Actions and its performance, but the real reason is that I wanted to try it out.

It's still broken, though: https://github.com/thoughtbot/suspenders/pull/1063.

PR #1069, opened 2021-08-14
Add support for generator tests

thiagoa:
Relates to #1070 

## Goals

The goal of this epic is to:

1. Speed up Suspender's test suite, composed mostly of end-to-end tests that hit the network via `bundler`, `yarn`, `rails new`, and what have you. A single example may take one or more minutes to finish;
2. Provide basic support to run "generator tests" through mini-integration tests, rather than running the full-blown `suspenders` command every time;
3. Pave the way to further extract new generators from `AppGenerator` (more information about that [here](https://github.com/thoughtbot/suspenders/blob/main/GOALS.md)). Waiting too long to run generator assertions over a tiny subset of the functionality is tiresome and not very practical;
4. Improve the developer experience.

The goal of this PR is to:

- Provide basic support to run generator tests (the hard stuff);
- Convert 7 feature tests over to generator tests (example: `spec/features/ci_spec` to `spec/suspenders/generators/ci_generator_spec`). I don't want the diff to be lengthy or overwhelming, that's why I'm only converting a few tests for now.

## Rationale

`spec/support/suspenders.rb` provides the `SuspendersTestHelpers` module, which is included globally in RSpec to run the `suspenders` command and generate a project in `tmp/dummy_app`. That directory is deleted through a `before` hook and generated from scratch on each example. Let's call that kind of test "feature test".

In this PR, the newly introduced "generator tests" rely on a similar approach, but they copy over a fake project directory containing just a few files (the bare minimum), from `spec/support/fixtures/dummy_app` to `tmp/dummy_app`. Generator tests are not supposed to hit the network, but they can interact with the filesystem and run fake bin commands (or even real bin commands if they run fast, don't hit the network, or have nasty side-effects).

The advantage of `suspenders/tmp/dummy_app` over a random directory in `/tmp` is that it's possible to rely on a deterministic location for debugging in case something goes wrong. The files are not deleted _after_ a test, so they are still available for debugging.

To make generator tests possible, I had to extract functionality from `SuspendersTestHelpers`:

- `TestPaths` - A module containing common filesystem paths to template files, fixtures, dummy app, fake app, and so on.
- `FileOperations` - Provide common file operations for both feature and generator specs.
- `EnvPath` - Temporarily modify `ENV["PATH"]` to run our fake binary files.

The new `GeneratorTestHelper` module I've created reuses those modules. Now we have two main helper modules:

- `GeneratorTestHelper` - For generator tests
- `FeatureTestHelper` (previously `SuspendersTestHelpers`) - For end-to-end tests

## Results so far

In CI, I'm seeing an improvement of around 7-10 minutes, from 33-40 minutes down to 26 minutes - afforded by 7 tests. As we convert new specs to the new format, that time will continue to go down.


spec/expand_json_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "spec_helper"
 
 RSpec.describe Suspenders::Actions::ExpandJson do
-  let(:destination_root) { File.join(root_path, "tmp") }
+  let(:destination_root) { TestPaths.root_path.join("tmp") }
```
thiagoa, line 4: This spec does not have explicit type metadata, so we don't include any helper modules in it by default. That's why we're explicitly calling `TestPaths.root_path` instead of just `root_path`.

thiagoa, line 4: This spec does not have explicit type metadata, so by default, we don't include any helper modules in it. That's why we're explicitly calling `TestPaths.root_path` instead of just `root_path`.

spec/features/api_spec.rb
```diff
@@ -1,9 +1,9 @@
 require "spec_helper"
 
-RSpec.describe "Suspend a new project with --api flag" do
+RSpec.describe "Suspend a new project with --api flag", type: :feature, autoclean: false do
```
thiagoa, line 3: The `spec/features` directory had two types of test:

1. `type: :generator`
2. Tests with no type, like this one

However,:

- All tests from this directory are actually feature tests;
- `type: :generator` tests were actually full-blown integration tests and not the more isolated generator tests we're striving for here.

What I'm doing here is tagging every test inside `spec/features` with `type: :feature`.

`autoclean: false` is temporary and will be removed in a future PR. It should be used for tests that use a `before(:all)` hook to setup the project directory, which will be a characteristic of all feature tests in the future.

Feature tests that don't use `before(:all)` (former "generator tests") are still relying on `autoclean` to keep backward compatibility, but they will be converted to more unitary "generator tests" as the work progresses.

spec/features/api_spec.rb
```diff
@@ -1,9 +1,9 @@
...
   before(:all) do
     drop_dummy_database
-    remove_project_directory
```
thiagoa, line 6: Instead of removing just the tmp project directory, the new overall approach is to clear the whole `tmp` directory. That is much more reliable and less error-prone. The consequence is that we will no longer run individual methods like `FakeGithub.clear!`, `FakeHeroku.clear!`, `remove_project_directory`, etc.

spec/features/api_spec.rb
```diff
@@ -1,9 +1,9 @@
...
+    clear_tmp_path
```
thiagoa, line 6: Instead of removing just the tmp project directory, the new overall approach is to clear the whole `tmp` directory. That is much more reliable and less error-prone. The consequence is that we will no longer run individual methods like `FakeGithub.clear!`, `FakeHeroku.clear!`, `remove_project_directory`, etc.

spec/features/production/deployment_spec.rb
```diff
@@ -1,8 +1,8 @@
 require "spec_helper"
 
-RSpec.describe "suspenders:production:deployment", type: :generator do
+RSpec.describe "suspenders:production:deployment", type: :feature do
   it "generates the configuration for a production deployment" do
-    rm "bin/deploy"
+    rm_file "bin/deploy"
```
thiagoa, line 5: I'm taking the opportunity to make these method names more consistent. They should all be suffixed with `_file`.

spec/spec_helper.rb
```diff
@@ -4,18 +4,19 @@
...
 RSpec.configure do |config|
-  config.include SuspendersTestHelpers
-  config.include ProjectFiles
+  config.include FeatureTestHelpers, type: :feature
+  config.include GeneratorTestHelpers, type: :generator
 
   config.before(:all) do
-    add_fakes_to_path
-    create_tmp_directory
+    FileOperations.create_tmp_path
```
thiagoa, line 14: Since we are now including helper modules selectively depending on the "type" metadata, it is more reliable to call helper methods with the fully qualified name at this point.

spec/spec_helper.rb
```diff
@@ -4,18 +4,19 @@
 
 require (Pathname.new(__FILE__).dirname + "../lib/suspenders").expand_path
 
-Dir["./spec/support/**/*.rb"].sort.each { |file| require file }
+Dir["./spec/support/*.rb"].sort.each { |file| require file }
```
thiagoa, line 7: The `spec/support` folder now has the following structure:

- `spec/support/*` - Normal spec support files
- `spec/support/fixtures` - Fake project template directory and other fixtures
- `spec/support/fakes` - Binary stubs and Ruby file stubs

spec/support/fake_operations.rb
```diff
@@ -0,0 +1,15 @@
+module FakeOperations
+  extend self
```
thiagoa, line 2: I really think `module_function` is more appropriate for this, but I'm sticking with `extend self` because RSpec is unable to call helper methods from within custom expectations if they are `module_functions`

spec/support/fakes/bundler/audit/task.rb
```diff
@@ -0,0 +1,11 @@
+require "pathname"
```
thiagoa, line 1: As you will see down below, we are sticking `spec/support/fakes/bundler` to `LOAD_PATH` like this:

```ruby
expect(`RAILS_ENV=development rake -I#{fakes_path} -rfake_rails -T`)
```

The bundler audit rake task that we're testing runs the following code:

```ruby
if Rails.env.development? || Rails.env.test?
  require "bundler/audit/task"
  Bundler::Audit::Task.new
end
```

That `require` will actually hit this Ruby file, and on `Bundler::Audit::Task.new` it will output `Fake task loaded`, which is a side-effect we can use in our assertion.

spec/support/test_paths.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module TestPaths
+  extend self
+
+  APP_NAME = "dummy_app"
```
thiagoa, line 6: I'd probably use another name, but this is for backward compatibility with current feature specs.

suspenders.gemspec
```diff
@@ -6,7 +6,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= #{Suspenders::RUBY_VERSION}"
   s.required_rubygems_version = ">= 2.7.4"
   s.authors = ["thoughtbot"]
-  s.date = Date.today.strftime("%Y-%m-%d")
```
thiagoa, line 9: The new version of `standardrb` linted against this line, so we can safely remove it.

suspenders.gemspec
```diff
@@ -30,6 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency "bitters", ">= 2.0.4"
   s.add_dependency "rails", Suspenders::RAILS_VERSION
 
+  s.add_development_dependency "pry"
```
thiagoa, line 32: Why not? :smile:

spec/support/fakes/bin/heroku
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env ruby
+
+require "pathname"
+require Pathname(__dir__) / ".." / ".." / "fake_heroku"
```
thiagoa, line 4: `heroku` and `hub` fakes already existed before. I think GH is showing a green diff because I refactored them to use `Pathname` instead of the old `File` API.

PR #1072, opened 2021-08-20
Remove preloader generator

### LINKED ISSUE #1071: Remove the spring generator
Rails already ships with Spring, so we can safely remove `SpringGenerator`.

thiagoa:
Closes #1071


templates/descriptions/testing.md
```diff
@@ -6,4 +6,4 @@ Set up the project for an in-depth test-driven development workflow.
 - Prepare for system tests in `spec/system`.
 - Integrate Formulaic for easier form testing.
 - RSpec infers the file type based on the directory name.
-- Install and configure RSpec and shoulda-matchers, with Spring integration.
+- Install and configure RSpec and shoulda-matchers
```
louis-antonopoulos, line 9:
```suggestion
- Install and configure RSpec and shoulda-matchers.
```

louis-antonopoulos (approved):
This looks nicely scoped! I wouldn't trust me to be the only reviewer, but it looks good!

templates/descriptions/testing.md
thiagoa: Ooops! Good call!

lib/suspenders/generators/testing_generator.rb
```diff
@@ -3,7 +3,6 @@
 module Suspenders
   class TestingGenerator < Generators::Base
     def add_testing_gems
-      gem "spring-commands-rspec", group: :development
```
mike-burns, line 6: As someone who is anti-Spring: yeah! Go away, Spring!

But don't Spring users actually want this gem?

suspenders.gemspec
```diff
@@ -6,7 +6,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= #{Suspenders::RUBY_VERSION}"
   s.required_rubygems_version = ">= 2.7.4"
   s.authors = ["thoughtbot"]
-  s.date = Date.today.strftime("%Y-%m-%d")
```
mike-burns, line 9: How do we want to set the `date` for the gem? Can you say more about this line and the ramifications of it?

templates/spring.rb
```diff
@@ -1,6 +0,0 @@
-Spring.watch(
```
mike-burns, line 1: Does Rails add this automatically? Where's that come from?

mike-burns (changes requested):
Can you describe the Rails internals and the motivations more in your PRs? This helps reviewers, yourself in the future when you're debugging something, and also the next Suspenders maintainers.

suspenders.gemspec
thiagoa: @mike-burns Sure! The new version of `standardrb` (it is unversioned in our gemspec, so it always gets the latest version) linted against this line saying it is not necessary because it's already the default behavior. In other words, the date will be set to the current date by default.

templates/spring.rb
thiagoa: Yes, Rails automatically adds a spring initializer that is exactly the same as this one. However, AFAIK Rails will remove Spring from the default `rails new` template.

lib/suspenders/generators/testing_generator.rb
thiagoa: I think so. Do you think we should detect if the app is using spring and if so add this line?

I see how that could be useful, but it would only be useful with `rails generate suspenders:testing`. If we generate a new app with the `suspenders` command (the classic approach), it's not that useful.

suspenders.gemspec
mike-burns: Aha! Nice.

I guess it's been like that for [nearly two decades](https://github.com/rubygems/rubygems/commit/3864a56b1054f0916c91ff8547653f16ce7cb52b)! Glad we found it, finally.

lib/suspenders/generators/preloader_generator.rb
```diff
@@ -1,122 +0,0 @@
...
-    def spring_config
-      template "spring.rb", "config/spring.rb", force: false, skip: true
-    end
-
-    def spring_gem
-      always_gsub_file("Gemfile", /# Spring speeds up development.*/, "")
-
-      gem_group :development do
-        gem "spring"
-        gem "spring-watcher-listen", "~> 2.0.0"
```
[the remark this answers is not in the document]
mike-burns: It [doesn't seem like it's in the default](https://github.com/rails/rails/search?q=spring-watcher-listen), but it [gives a potential performance boost](https://github.com/rails/spring#watching-files-and-directories):

> By default Spring polls the filesystem for changes once every 0.2 seconds. This method requires zero configuration, but if you find that it's using too much CPU, then you can use event-based file system listening by installing the spring-watcher-listen gem.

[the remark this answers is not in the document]
thiagoa: In my opinion, the issue is:

- If we're using the classic `suspenders` command, the default will be _not to have_ spring installed, so we do not need `spring-watcher-listen`
- If we're using `rails generate suspenders:TASK`, we will not care about Spring either - the user will either adopt Rails defaults, which includes spring, or will setup Spring themselves, or if using Rails 7 will explicitly ask for `rails new` to include Spring, which probably already includes `spring-watcher-listen`. So I don't think we need `spring-watcher-listen` in that case as well.

Does that make sense to you?

lib/suspenders/generators/testing_generator.rb
thiagoa: @mike-burns Because with this PR, the `suspenders` command will ignore Spring and will generate an application *without* spring.

However, from a pure generator standpoint it makes sense! We could detect if the user is using spring, and if so add `spring-commands-rspec` when generating `suspenders:testing`. Do you think that makes sense?

lib/suspenders/generators/preloader_generator.rb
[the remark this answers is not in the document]
mike-burns: > If we're using the classic suspenders command, the default will be not to have spring installed

I confess that I haven't kept up but now I've learned that [Spring is gone]. Finally!

[Spring is gone]: https://github.com/rails/rails/pull/42997

OK yes, get rid of all the spring related gems! Begone!

mike-burns (approved):

PR #1075, opened 2021-09-03
Convert remaining generator tests

thiagoa:
## Related pull requests:

- [Add support for generator tests](https://github.com/thoughtbot/suspenders/pull/1069)
- [Refactor generator test smells](https://github.com/thoughtbot/suspenders/pull/1076)

## What?

In this PR, we are converting the remaining feature tests disguised as generator tests to actual generator tests that run reasonably fast!

We are using the same structure introduced in [Refactor generator test smells](https://github.com/thoughtbot/suspenders/pull/1069).

## Results

In CI, the test suite used to run in ~36 minutes. Now it runs in ~12 minutes, so it's around 3 times faster. That time should continue to drop as we extract new generators from `AppGenerator`.


spec/support/fixtures/dummy_app/bin/yarn
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env ruby
+
+require "pathname"
```
thiagoa, line 3: Differently from the fake `bundler` binary, this binary is located in the dummy app's fixtures. The reason is that we call `yarn` from the root of the Rails application, like this:  `bin/yarn` -- so `yarn` does not need to be in the `PATH`.

spec/support/generator_test_helpers.rb
```diff
@@ -5,14 +5,20 @@ module GeneratorTestHelpers
   include TestPaths
   include FileOperations
 
-  def invoke!(klass, *args, **kwargs)
+  def invoke!(klass, *args, **kwargs, &block)
```
thiagoa, line 8: Changes in this file overlap with the same changes [in this PR](https://github.com/thoughtbot/suspenders/pull/1076). It will be fixed during merge.

spec/support/fixtures/dummy_app/config/environments/production.rb
```diff
@@ -0,0 +1,24 @@
+# This file is part of our fake Rails app, for testing purposes
+Rails.application.configure do
```
thiagoa, line 2: One of the generators relies on the existence of this file, which is not present in the `suspenders/templates` directory. For that reason, we are creating it as part of the dummy app fixtures.

spec/spec_helper.rb
```diff
@@ -14,9 +14,4 @@
     FileOperations.create_tmp_directory
     EnvPath.prepend_env_path!(TestPaths.fake_bin_path)
   end
-
-  config.before(:each, type: :feature) do |spec|
```
thiagoa, line 18: This hook was needed for the end-to-end generator tests (the ones we just converted to pure generator tests) to properly clean up the `tmp` directory. Now that all tests are converted, we no longer need it! We can also remove `autoclean: false` from the remaining (real) feature specs.

spec/features/api_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "spec_helper"
 
-RSpec.describe "Suspend a new project with --api flag", type: :feature, autoclean: false do
```
thiagoa, line 3: The reason why we are doing this is explained [here](https://github.com/thoughtbot/suspenders/pull/1075/files#r702303397).

spec/support/fake_yarn_matchers.rb
```diff
@@ -0,0 +1,25 @@
+# FakeYarn records yarn information through a fake "bin/yarn"
+# binary. This RSpec matcher relies on that functionality.
+RSpec::Matchers.define :have_yarned do |actual_yarn_args = ""|
```
mike-burns, line 3: Hah! Good name.

spec/support/fake_yarn_matchers.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    @yarn_ran = fake_yarn.ran?
+    @given_yarn_args = fake_yarn.arguments
+
+    @yarn_ran && @given_yarn_args == actual_yarn_args
+  end
+
+  failure_message do
+    if !@yarn_ran
+      "The generator did not call yarn!"
```
mike-burns, line 17: Do you mind dropping the `!` from these strings? Pedantic, maybe, but I don't like shouting when I'm trying to debug a problem.

mike-burns (approved):
Amazing work.

spec/support/rails_stub.rb
```diff
@@ -0,0 +1,17 @@
+module RailsStub
+  extend RSpec::Mocks::ExampleMethods
+
+  module_function
```
gnfisher, line 4: TIL you can not pass arguments and the susequently defined methods become module functions!

gnfisher (approved):
Awesome :+1:

spec/support/fake_yarn.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def arguments
+    @arguments ||= File.read(TAPE)
+  rescue Errno::ENOENT
+    warn <<~MSG
+      Could not find #{TAPE.basename} file!
+      Fake yarn did not run.
+    MSG
+
+    ""
```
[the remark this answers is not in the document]
thiagoa: Good question! That was my gut reaction as well when I saw we were storing temp files within the project folder, but there are actually two reasons for that:

1. That's how it works in feature tests, so I thought I'd keep backward compatibility;
2. It's useful to rely on a deterministic place to debug if something goes wrong. Suspenders' tmp folder is cleaned up before running a test, so we can still peek at the folder to debug after a test runs. `Tempfile` with a random location would not work as well, but `/tmp/fixed_location` would; however, as soon as the location is fixed, it doesn't make much of a difference to use `suspenders/tmp` or `/tmp/fixed_location` IMO.

spec/support/fake_yarn_matchers.rb
thiagoa: Not at all! I feel you.

spec/support/rails_stub.rb
thiagoa: Hum, can't we do `module_function :my_method` as well (if that's what you mean)?

I think it works much like the `private` modifier.

gnfisher: Right! I didnt't know it worked like private, thanks!

PR #1077, opened 2021-09-04
Fix bundler RSpec matcher error messages

thiagoa:
The error message when bundler is given unexpected arguments is switching actual with expected arguments. This PR fixes that problem.

### DIFF
--- a/spec/support/fake_bundler_matchers.rb
+++ b/spec/support/fake_bundler_matchers.rb
@@ -43,7 +43,7 @@ RSpec::Matchers.define :have_bundled do |actual_bundler_args = ""|
 
       <<~MSG
         Bundler was given unexpected arguments.
-        Expected #{given} but got #{actual}
+        Expected #{actual} but got #{given}
       MSG
     elsif @gemfile_matching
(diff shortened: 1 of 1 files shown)


spec/support/fake_bundler_matchers.rb
```diff
@@ -43,7 +43,7 @@
 
       <<~MSG
         Bundler was given unexpected arguments.
-        Expected #{given} but got #{actual}
+        Expected #{actual} but got #{given}
```
emilford, line 46: I _think_ you're right that these backwards. Nice catch. I still find this a little confusing, however. Would it be clearer to rename `actual` to `expected` and `given` to `actual` throughout to match the RSpec naming convention? Maybe as a separate PR?

emilford (approved):

spec/support/fake_bundler_matchers.rb
thiagoa: Thank you, I agree that `expected` and `actual` is much more clear! Since the changes are few, I'm reusing this PR if you don't mind.

PR #1078, opened 2021-09-04
Introduce a clear failure message for the match_contents RSpec matcher

thiagoa:
The `match_contents` RSpec matcher is used to match on the contents of files manipulated by a generator.  When a failure happens, however, the reason why it happens is often not clear.

This PR improves the error message and includes the contents of the target file in it to allow for better debugging.

### DIFF
--- a/spec/support/match_contents_matcher.rb
+++ b/spec/support/match_contents_matcher.rb
@@ -1,17 +1,18 @@
 RSpec::Matchers.define :match_contents do |regexp|
-  match do |filename|
-    @filename = filename
-    @contents = TestPaths.app_path.join(filename).read
-
-    @contents =~ regexp
+  def file_contents(filename)
+    TestPaths.app_path.join(filename).read
   end
 
-  failure_message do
-    <<~MESSAGE
-      #{@filename} does not match the regex #{regexp}.
-      Contents for #{@filename} are:
+  match do |filename|
+    file_contents(filename) =~ regexp
+  end
 
-      #{@contents}
+  failure_message do |filename|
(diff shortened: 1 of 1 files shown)


spec/support/match_contents_matcher.rb
```diff
@@ -1,7 +1,18 @@
...
-    contents =~ regexp
+    @filename = filename
+    @contents = TestPaths.app_path.join(filename).read
+
+    @contents =~ regexp
+  end
+
+  failure_message do
+    <<~MESSAGE
+      #{@filename} does not match the regex #{regexp}.
```
emilford, line 11: `filename` is passed to `failure_message` if we want to get rid of the instance variable.

thiagoa: Good call! Will do.

spec/support/match_contents_matcher.rb
```diff
@@ -1,7 +1,18 @@
 RSpec::Matchers.define :match_contents do |regexp|
   match do |filename|
-    contents = TestPaths.app_path.join(filename).read
-    contents =~ regexp
+    @filename = filename
+    @contents = TestPaths.app_path.join(filename).read
```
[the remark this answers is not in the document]
thiagoa: I like that! The only (minor) drawback is that on test failure it would run the helper twice, and we'd have to pass it the `filename` param. Otherwise, we could memoize it and that would not be an issue.

But as I said, I think that's a very minor issue on test environment and I even prefer the version with the helper method, so I'm doing it. Thanks!

PR #1079, opened 2021-09-04
Remove remaining spring references

thiagoa:
In this PR, we are removing the `spring stop` command from `AppGenerator`.

I thought it would be useful to run `spring stop` for individual generators, but within the context where `spring stop` is being removed we generate a full application with the `suspenders` command, so there's no risk of spring running or interfering with anything.


emilford (approved):
Thanks for the explanation. Sounds good to me.

Issue #1073, opened 2021-08-20
Make suspenders command smarter regarding the Ruby version


thiagoa:
If I run Suspenders with a Ruby version other than the one declared in `Gemfile`, I'm still able to generate my application but I get the following message over and over:

> Your Ruby version is X.X.X, but your Gemfile specified 2.6.6

 Here's some sample output:

```
Your Ruby version is 2.7.2, but your Gemfile specified 2.6.6
    generate  suspenders:advisories
       rails  generate suspenders:advisories
Your Ruby version is 2.7.2, but your Gemfile specified 2.6.6
    generate  suspenders:preloader
       rails  generate suspenders:preloader
Your Ruby version is 2.7.2, but your Gemfile specified 2.6.6
    generate  suspenders:staging:pull_requests
       rails  generate suspenders:staging:pull_requests
Your Ruby version is 2.7.2, but your Gemfile specified 2.6.6
    generate  suspenders:production:single_redirect
       rails  generate suspenders:production:single_redirect
Your Ruby version is 2.7.2, but your Gemfile specified 2.6.6
    generate  suspenders:production:compression
       rails  generate suspenders:production:compression
Your Ruby version is 2.7.2, but your Gemfile specified 2.6.6
    generate  suspenders:production:force_tls
```

There are some options to deal with this problem. The following ones crossed my mind:

1. Declare the minimum and maximum Ruby versions, and configure `Gemfile` with whatever version the user is using. If the user's version is less than the minimum or the maximum, fail before running anything.
2. Fail if the user's Ruby version is different than `Suspenders::RUBY_VERSION`

thiagoa:
Personally, I think option 1 is less strict and more appropriate. With option 2, we are forcing the user to go with a particular Ruby version; for instance, if I `gem install suspenders` with a newer Ruby version, I would still get the same errors.

Declare a minimum Ruby version and letting the user choose which version they want to run may be more sensible. It's also worth noting that option 2 is harder to maintain because it requires constant updates to the codebase. 

That all said, there are some drawbacks to option 1. One of them is that we should not use any Ruby features not present in `MINIMUM_RUBY_VERSION`, and that would be something hard to put under test coverage.

stevepolitodesign:
Just noting that after merging in #1087, we are no longer are able to generate a new app if the versions do not match. Instead, the script will abort.

[...]

PR #1085, opened 2021-11-20
Make removing comments from config files work reliably

thiagoa:
## What this PR does

Reliably remove comments from Rails config files when generating Suspenders apps. With the [`parser`](https://github.com/whitequark/parser) gem, we're now removing comments by processing their surrounding code using AST (abstract syntax tree) processors.

We were using a regex to do that, which is text-based and thereby subject to edge cases that only a proper AST solution can handle.

## First motivation

[This pull request](https://github.com/thoughtbot/suspenders/pull/1044) made changes to some config files to fix the Action Mailer asset host.

While working on merging that PR, the tests were not passing because they looked for an output that was missing in one of the changed config files:

```ruby
config.action_mailer.asset_host = %q{"https://#{ENV\.fetch\("ASSET_HOST", ENV\.fetch\("APPLICATION_HOST"\)\)}}
```

The PR code changes are kind of straightforward and that config option should have been present in the config file but it was not.

The reason was that Suspenders was removing code comments with an overly permissive regex : (`/^.*#.*$/`), which removed every line containing a `#` character, no matter where they were in the line - so it was also removing the new Action Mailer config setting.

That regex was also silently messing up with other config lines, for example:

```ruby
config.public_file_server.headers = {
  'Cache-Control' => "public, max-age=#{2.days.to_i}"
}
```

As a quick fix to merge that PR, I've changed the regex to `/^\s*#.*$/` but it was not an ideal solution - which is what I'm striving for with this PR. With the old text-based regex solution, there were still nasty edge cases to deal with, and we may never know when they'd occur. Also, the previous regex did not work with inline comments.

## Second motivation

The final aesthetics of the config files could be improved in my opinion. This was how they looked like:

```ruby
  config.cache_classes = false
  config.eager_load = false
  config.consider_all_requests_local = true
  if Rails.root.join('tmp', 'caching-dev.txt').exist?
    config.action_controller.perform_caching = true
    config.action_controller.enable_fragment_cache_logging = true
    config.cache_store = :memory_store
    config.public_file_server.headers = {
      'Cache-Control' => "public, max-age=#{2.days.to_i}"
    }
  else
    config.action_controller.perform_caching = false
    config.cache_store = :null_store
  end
  config.active_storage.service = :local
  config.action_mailer.raise_delivery_errors = true
  config.after_initialize do
    Bullet.enable = true
    Bullet.bullet_logger = true
    Bullet.rails_logger = true
  end

  # ...
end
```

And with the newest solution they look like this:

```ruby
  config.cache_classes = false
  config.eager_load = false
  config.consider_all_requests_local = true

  if Rails.root.join('tmp', 'caching-dev.txt').exist?
    config.action_controller.perform_caching = true
    config.action_controller.enable_fragment_cache_logging = true

    config.cache_store = :memory_store
    config.public_file_server.headers = {
      'Cache-Control' => "public, max-age=#{2.days.to_i}"
    }
  else
    config.action_controller.perform_caching = false
    config.cache_store = :null_store
  end

  config.action_mailer.raise_delivery_errors = true

  config.after_initialize do
    Bullet.enable = true
    Bullet.bullet_logger = true
    Bullet.rails_logger = true
  end

  # ...
```

Which preserves newlines where appropriate, so that's easier to read.

## Third motivation

This pull request uses the `parser` gem, and I feel like introducing that gem in the project may be a good move over the long term.

For example, we need to finish up the [gem action](https://github.com/thoughtbot/suspenders/pull/1066) to sort `gem` declarations alphabetically when adding gems to the `Gemfile`, and the most reliable way to do that is with AST parsing. 

## Other related changes

1. Make the indentation of some templates match the Rails environment's
config files'. This goes along with the aesthetics introduced in this
PR.

2. Because of (1), I've also found a bug that was concatenating two
lines of codes in `config/environments/production.rb` with no blank
line between them (in email_generator.rb), so I'm taking the
opportunity to fix it

3. standardrb fixes after updating to the latest standardrb version


.ruby-version
```diff
@@ -1 +1 @@
-2.7.2
+2.7.4
```
thiagoa, line 1: The `parser` gem emits a warning if we're using Ruby 2.7.x but < `2.7.4`. I think it's OK to bump it in Suspenders since we're dealing with minor versions.

lib/suspenders/actions/strip_comments_action.rb
```diff
@@ -0,0 +1,203 @@
...
+      class StripLeadingScopeNewlines
+        def self.call(source, parser)
+          buffer = Parser::Source::Buffer.new(nil, source: source)
+          ast = parser.parse(buffer)
+
+          LeadingNewlineStripRewriter.new.rewrite(buffer, ast)
+        end
+
+        class LeadingNewlineStripRewriter < Parser::TreeRewriter
+          def on_module(node)
```
thiagoa, line 148: There are probably more cases to deal with here if we'd want this to be more generic. Currently, we're dealing with modules, classes, and blocks, which may be enough to cleanly delete comments from the config files.

mike-burns:
+1 to using `parser`. Great idea. Love it.

What do we think of leaving the comments there instead?

thiagoa:
@mike-burns In my opinion, removing comments is a nice feature so I would vote for keeping it, as most default comments are noise. Another option is to have a dedicated generator where the user would explicitly opt-in for this feature.

 Personally, I would maybe extract this functionality to a gem in the future while making the necessary changes for it to be as generic as possible.

mike-burns (approved):
LGTM. This is an exciting new land for Suspenders!

I'm a little worried about that Ruby version change. Is this going to cause us to chase Ruby versions more than anything else?

lib/suspenders/actions/strip_comments_action.rb
```diff
@@ -0,0 +1,203 @@
...
+
+module Suspenders
+  module Actions
+    class StripCommentsAction
+      class << self
+        def call(source)
+          parser = Parser::CurrentRuby.new
+
+          source
+            .yield_self { |s| strip_comments(s, parser) }
```
MatheusRich, line 11: Do we need to support old Ruby versions? If we don't, what about using `then` here. I love how it reads
```suggestion
            .then { |src| strip_comments(src, parser) }
```

thiagoa:
@mike-burns We were already chasing Ruby versions, so that should not be an issue :cry: 

I wrote up the following ticket to deal with this Ruby version problem: https://github.com/thoughtbot/suspenders/issues/1073

Hopefully, it will address the concerns of "chaining Ruby versions".

lib/suspenders/actions/strip_comments_action.rb
thiagoa: @MatheusRich Good call! `then` was introduced in Ruby 2.6 so we should be safe here. We are not supposed to support old Ruby versions. Thank you for that, I'll replace it with `then`.

thiagoa: About the variable name, IMO `s` is perfectly good and succinct because it can be inferred from the context that we're dealing with `source`.

Issue #1080, opened 2021-10-01
Make Suspenders abort when something goes wrong


thiagoa:
## Description of the problem

Sometimes the suspenders command fails in the middle of execution, for example:

```
Installing listen 3.7.0
Using method_source 1.0.0
Fetching pg 1.2.3
Installing pg 1.2.3 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

....
```

But it does not abort. It continues to run the remaining setup steps and generators indefinitely.

Errors like that may trigger a cascade of other errors because later generators would fail to bundle, which would result in an incomplete application with missing pieces. That is [confusing to users](https://github.com/thoughtbot/suspenders/issues/1046) and only observable by wading through the output log, if observable at all.

Many non-obvious errors can happen because the program does not abort:

- [Wrong Ruby version](https://github.com/thoughtbot/suspenders/issues/1073) - In that case, it's not obvious that every generator would fail to run, as the suspenders command would just keep running. 
- [Problems with Bundler setup](https://github.com/thoughtbot/suspenders/issues/1047)

We must fail as early as possible to make it clear something's wrong with the user's setup, or even Suspenders itself. There are also some measures we can take on our side to make Suspenders more lenient (and flexible) to different versions of software.

## A possible solution

Currently, most setup steps executed by the suspenders command rely on subprocesses (including running individual generators,) which means we can check their exit status and abort with a meaningful message if something goes awry, for example:

```ruby
def customize_gemfile
  build :replace_gemfile, options[:path]
  bundle_command "install"

  if $?.exitstatus.nonzero?
    abort "Some failure message"
  end
end
```

But before going that route we should consider the following question:

- Does Rails/Thor provide cleaner ways to abort?

PR #1087, opened 2022-01-29
Make Suspenders abort when something goes wrong

### LINKED ISSUE #1073: Make suspenders command smarter regarding the Ruby version

thiagoa:
Closes [#1080](https://github.com/thoughtbot/suspenders/issues/1080)

This task is part of our "Rescue" board: https://github.com/thoughtbot/suspenders/projects/1

## What changed?

`bin/suspenders` is now halting on error. Not halting on error was a huge source of confusion for Suspenders users. Some examples of issues we are resolving here:

- If using the wrong Ruby version, Suspenders will fail loudly. Currently, Suspenders is limited to a single Ruby version (we'll make the Ruby version flexible in another PR.) Example:

```
      create  storage/.keep
      create  tmp/storage
      create  tmp/storage/.keep
      remove  config/initializers/cors.rb
      remove  config/initializers/new_framework_defaults_6_1.rb
       force  Gemfile
         run  bundle install
Your Ruby version is 2.7.5, but your Gemfile specified 2.7.4
```

No more output. The user will instantly know something is wrong and will attempt to fix the error.

- If Postgres is not running, Suspenders will fail when trying to create the database
- If a generator fails because of some unmet requirement, Suspenders will halt as well (recall that every individual generator runs within a subprocess.)

## Changelog

- Override thor's `exit_on_failure?` to `true` on `AppGenerator`. `AppGenerator` is the class that extends `Rails::Generators::AppGenerator` and runs our own version of `rails new` through `bin/suspenders`.
- Override thor's `exit_on_failure?` to `true` on `Suspenders::Generators::Base` (the base class of our individual generators.)
- Make `Rails::Generators::AppGenerator#bundle_command` fail on error. Unfortunately, it is not hooked up to thor's default error handling mechanism, so it won't leverage Thor's `exit_on_failure` setting.
- Override the default value of `skip_spring` to `true`. Rails will still try to generate spring binstubs when finishing up the application, which would make `bin/suspenders` fail given our newly introduced `exit_on_failure` setting.
- Avoid the `spring stop` command from halting `bin/suspenders` when `spring` is not installed or when it calls a binstub that can't find the real `spring`.
- Add unit tests for the new settings. `exit_on_failure` is a `:nodoc:` in Thor, and it's unfortunate they don't provide an official setting for that. If it ever breaks, we'll need to know.
- Don't run the stylesheets generator when `--api` is set. That avoids the stylesheet generator (and thus `bin/suspenders`) from failing with our new settings.
- Rename `DEBUG` => `SHOW_DEBUG` because `DEBUG` is already taken by Ruby and prints out too many symbols.

### DIFF
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -3,6 +3,8 @@ require "rails/generators/rails/app/app_generator"
 
 module Suspenders
   class AppGenerator < Rails::Generators::AppGenerator
+    include ExitOnFailure
+
     hide!
 
     class_option :database, type: :string, aliases: "-d", default: "postgresql",
@@ -34,6 +36,9 @@ module Suspenders
 
     class_option :skip_turbolinks,
       type: :boolean, default: true, desc: "Skip turbolinks gem"
+
+    class_option :skip_spring, type: :boolean, default: true,
+      desc: class_options[:skip_spring].description
 
     def finish_template
       invoke :suspenders_customization
@@ -138,18 +143,16 @@ module Suspenders
     end
 
     def generate_default
-      run("spring stop > /dev/null 2>&1")
+      run("spring stop > /dev/null 2>&1 || true")
       generate("suspenders:runner")
       generate("suspenders:profiler")
       generate("suspenders:json")
       generate("suspenders:static")
-      generate("suspenders:stylesheet_base")
+      generate("suspenders:stylesheet_base") unless options[:api]
       generate("suspenders:testing")
       generate("suspenders:ci")
       generate("suspenders:js_driver")
-      unless options[:api]
-        generate("suspenders:forms")
-      end
+      generate("suspenders:forms") unless options[:api]
       generate("suspenders:db_optimizations")
       generate("suspenders:factories")
       generate("suspenders:lint")
--- a/spec/support/feature_test_helpers.rb
+++ b/spec/support/feature_test_helpers.rb
@@ -1,3 +1,4 @@
+require "English"
 require_relative "env_path"
 require_relative "file_operations"
 require_relative "test_paths"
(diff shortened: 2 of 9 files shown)


lib/suspenders/generators/app_generator.rb
```diff
@@ -138,18 +143,16 @@ def remove_routes_comment_lines
...
 
     def generate_default
-      run("spring stop > /dev/null 2>&1")
+      run("spring stop > /dev/null 2>&1 || true")
       generate("suspenders:runner")
       generate("suspenders:profiler")
       generate("suspenders:json")
       generate("suspenders:static")
-      generate("suspenders:stylesheet_base")
+      generate("suspenders:stylesheet_base") unless options[:api]
```
thiagoa, line 151: This prevents `api_spec.rb` from failing because it tries to run `yarn` but there's no such command in API mode. I'm sure there are other generators we'll want to skip in that case, and we'll probably want to address them in a future PR.

spec/support/feature_test_helpers.rb
```diff
@@ -158,7 +177,7 @@ def run_in_project
   end
 
   def debug(output)
-    if ENV["DEBUG"]
```
thiagoa, line 161: It looks like `DEBUG` is already taken by Ruby and it outputs a ton of garbage.

mike-burns (approved):
Nice! Thank you for doing this.

PR #1091, opened 2022-04-08
Updates default configuration to use requests specs

### LINKED ISSUE #1090: After scaffolding a model, one of the generated tests fails
The generated scaffolded tests include controller specs which don't render views by default.

rspec-rails is not compatible with the rails 6.0 because of a wrong number of arguments

Solution:

Upgrade rspec-rails (will default to generate requests specs which render views)

dorianmariefr:
We were trying out suspenders, e.g. `suspenders projectname`

Then did a scaffold `rails g scaffold Book title`

Then when running the specs (`rspec`) we got an error related to the views not rendering.

A good fix is to use requests specs instead of controller specs, they render the views so they test more code

Fixes #1090
 
Co-Authored-By: Dorian Marié <dorian@dorianmarie.fr>

### DIFF
--- a/lib/suspenders/generators/testing_generator.rb
+++ b/lib/suspenders/generators/testing_generator.rb
@@ -3,7 +3,7 @@ require_relative "base"
 module Suspenders
   class TestingGenerator < Generators::Base
     def add_testing_gems
-      gem "rspec-rails", "~> 3.6", group: %i[development test]
+      gem "rspec-rails", "~> 5.1", group: %i[development test]
       gem "shoulda-matchers", group: :test
 
       Bundler.with_unbundled_env { run "bundle install" }
--- a/lib/suspenders/version.rb
+++ b/lib/suspenders/version.rb
@@ -4,5 +4,5 @@ module Suspenders
     .read("#{File.dirname(__FILE__)}/../../.ruby-version")
     .strip
     .freeze
-  VERSION = "1.55.1".freeze
+  VERSION = "1.55.2".freeze
 end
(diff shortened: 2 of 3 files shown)


iftheshoefritz:
Controller specs are out of fashion with Rails anyway, this change seems like a win to me.

lib/suspenders/generators/testing_generator.rb
```diff
@@ -3,7 +3,7 @@
 module Suspenders
   class TestingGenerator < Generators::Base
     def add_testing_gems
-      gem "rspec-rails", "~> 3.6", group: %i[development test]
+      gem "rspec-rails", "~> 5.1", group: %i[development test]
```
iftheshoefritz, line 6: this might not have been necessary but it seems like a good idea to upgrade in any case.

lib/suspenders/version.rb
```diff
@@ -4,5 +4,5 @@ module Suspenders
     .read("#{File.dirname(__FILE__)}/../../.ruby-version")
     .strip
     .freeze
-  VERSION = "1.55.1".freeze
+  VERSION = "1.55.2".freeze
```
thiagoa, line 7: Can we undo this version bump? There are other unreleased changes in mainline and we may want to version them differently. Thank you!

thiagoa (commented):
This is generally looking good, thank you for taking this on!

The tests are failing, and it looks like you've forgotten to change the generator template here:

https://github.com/thoughtbot/suspenders/blob/3afb005c601213f6f0f76aba8bf002c380b0ec42/lib/suspenders/app_builder.rb#L84

I suspect changing `false` to `true` will make it pass, is that correct?

After fixing the test failure and undoing the version bump, looks good to merge!

dorianmariefr:
Note that I needed to run `bin/webpack-dev-server` with a `.node-version` of `16.15.0`.

dorianmariefr:
This pull request is superseded by https://github.com/thoughtbot/suspenders/pull/1094 because I use my fork

PR #1092, opened 2022-05-06
Drop use of git in gemspec

thiagoa:
Closes https://github.com/thoughtbot/suspenders/issues/1055

Changes `suspenders.gemspec` to not depend on git. Additionally, we can safely remove `s.test_files` since [its use is no longer recommended](https://github.com/rubygems/guides/issues/90) and was introduced in the Suspenders codebase many years ago.


thiagoa:
I'm merging this PR since it's basically the feedback that was asked in https://github.com/thoughtbot/suspenders/pull/1057

PR #1098, opened 2022-06-24
Run database migrations as the last step of bin/suspenders 

### LINKED ISSUE #1095: Running ./bin/setup fails to run
## Steps to reproduce

1. `$ suspenders name_of_app`
2. `$ cd name_of_app`
3. `$ ./bin/setup`

```
Database 'demo_app_development' already exists
Database 'demo_app_test' already exists
You have 1 pending migration:
  20220610190223 CreateDelayedJobs
Run `rails db:migrate` to update your database then try again.
```

stevepolitodesign:
This commit ensures that `bundle exec rails db:migrate` is run before the app is generated. Prior to this commit, running `./bin/setup` in the app generated by suspenders would have resulted in a failure due to a pending migration.

Issues
------
Closes #1095

Co-authored-by: Thiago Silva <thiagoaraujos@gmail.com>

### DIFF
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -135,6 +135,10 @@ module Suspenders
       bundle_command "exec rails db:create db:migrate"
     end
 
+    def run_database_migrations
+      bundle_command "exec rails db:migrate"
+    end
+
     def replace_gemfile(path)
       template "Gemfile.erb", "Gemfile", force: true do |content|
         if path
--- a/lib/suspenders/generators/app_generator.rb
+++ b/lib/suspenders/generators/app_generator.rb
@@ -59,6 +59,7 @@ module Suspenders
       invoke :generate_deployment_default
       invoke :remove_config_comment_lines
       invoke :remove_routes_comment_lines
+      invoke :run_database_migrations
       invoke :outro
     end
 
@@ -140,6 +141,10 @@ module Suspenders
 
     def remove_routes_comment_lines
       build :remove_routes_comment_lines
+    end
+
+    def run_database_migrations
+      build :run_database_migrations
     end
 
     def generate_default


templates/bin_setup
```diff
@@ -13,7 +13,10 @@ bundle check || bundle install
 # Install JavaScript dependencies
 bin/yarn
 
-# Set up database and add any development seed data
+# Set up database and run migrations
+bin/rails db:reset || bin/rails db:migrate
```
stevepolitodesign, line 17: This is inspired by [David Copeland's Rails template](https://github.com/davetron5000/rails-app-template-sustainable/blob/main/bin/setup.tt#L20-L23).

Basically, we want to be able to confidently run migrations without needing to know if the database exists first.

stevepolitodesign (commented):
Some thoughts:

- There might be some room for optimizations since running `bin/rails db:reset` drop and creates the database, and then `bin/rails dev:prime` does the same thing by invoking `db:setup`. The challenge is that it's common to run `bin/rails dev:prime` isolation, which means it should always invoke `db:setup`.
- Is there a way to write a unit test to ensure the newly added script exists?

---

Output when the command is run for the first time on a new application:

```
$ ./bin/setup
The Gemfile's dependencies are satisfied
yarn install v1.22.17
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.15s.
Dropped database 'demo_1_development'
Dropped database 'demo_1_test'
Created database 'demo_1_development'
Created database 'demo_1_test'
You have 1 pending migration:
  20220624140826 CreateDelayedJobs
Run `rails db:migrate` to update your database then try again.
== 20220624140826 CreateDelayedJobs: migrating ================================
-- create_table(:delayed_jobs)
   -> 0.0088s
-- add_index(:delayed_jobs, [:priority, :run_at], {:name=>"delayed_jobs_priority"})
   -> 0.0044s
== 20220624140826 CreateDelayedJobs: migrated (0.0133s) =======================

Database 'demo_1_development' already exists
Database 'demo_1_test' already exists
```

Output when the command is run a subsequent time on an existing application:

```
$ ./bin/setup              
The Gemfile's dependencies are satisfied
yarn install v1.22.17
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.16s.
Dropped database 'demo_1_development'
Dropped database 'demo_1_test'
Created database 'demo_1_development'
Created database 'demo_1_test'
Database 'demo_1_development' already exists
Database 'demo_1_test' already exists
overwrite .env? (y/n [n]) y
```

templates/bin_setup
thiagoa: Is there a situation where `bin/rails db:reset` would fail? If I do `bin/rails db:drop` and then `bin/rails db:reset` there's no failure -- if the database does not exist, it still displays the message "Dropped database ....".

I wonder if that was an issue in older Rails versions. If `db:reset` consistently works everytime, I'd drop the `|| bin/rails db:migrate`. What do you think?

thiagoa: Never mind, I think I understand the purpose. I will delete this comment and write another one :)

thiagoa: @stevepolitodesign Maybe `db:reset` would be too destructive? I would not expect `bin/setup` to destroy and recreate my database, but rather to be idempotent. What do you think of:

```sh
bin/rails db:setup || bin/rails db:migrate
```

`db:setup` doesn't run `db:drop`, so it may be a better option.

If `db:setup` fails, `db:migrate` will be run as a fallback. `db:reset`, on the other hand, will always drop the database so it would probably never fallback to `db:migrate`.

> There might be some room for optimizations since running bin/rails db:reset drop and creates the database, and then bin/rails dev:prime does the same thing by invoking db:setup

Maybe you could update `dev.rake` with something like this to DRY it up:

```ruby
if Rails.env.development? || Rails.env.test?
  require "factory_bot"

  namespace :dev do
    task sync_db: :environment do
      `bin/rails db:setup || bin/rails db:migrate`
    end

    desc "Sample data for local development environment"
    task prime: "dev:sync_db" do
      include FactoryBot::Syntax::Methods

      # create(:user, email: "user@example.com", password: "password")
    end
  end
end
```

The `sync_db` task would be hidden because there's no `desc`, and it wouldn't appear in `rake -T`.

thiagoa:
@stevepolitodesign It looks like some generators generate migration files that are not run by `bin/suspenders`. When `bin/suspenders` finishes, the expectation is that the app is 100% ready to go but instead we get a pending migration:

```ruby
You have 1 pending migration:                                                                                                  │
  20220704184101 CreateDelayedJobs                                                                                             │
Run `rails db:migrate` to update your database then try again.                                                                 │
```

In addition to the `bin/setup` change, I wonder if it would be interesting to also add a `db:migrate` as the last step of `bin/suspenders`? What do you think? I think it would fit [here](https://github.com/thoughtbot/suspenders/blob/f3ce221d553e611d1c8e32a22cafd9fa214999f8/lib/suspenders/generators/app_generator.rb#L60). You can call `db:migrate` [similarly to this](https://github.com/thoughtbot/suspenders/blob/f3ce221d553e611d1c8e32a22cafd9fa214999f8/lib/suspenders/app_builder.rb#L134-L136).

thiagoa (approved):
LGTM, that fixes the issue!

stevepolitodesign (commented):
Output when the command is run for the first time on a new application:

```
% ./bin/setup
The Gemfile's dependencies are satisfied
yarn install v1.22.17
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.15s.
Database 'name_of_app_development' already exists
Database 'name_of_app_test' already exists
```

Output when the command is run a subsequent time on an existing application:

```
% ./bin/setup
The Gemfile's dependencies are satisfied
yarn install v1.22.17
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.15s.
Database 'name_of_app_development' already exists
Database 'name_of_app_test' already exists
overwrite .env? (y/n [n]) y
```

Output when the command is run for when the database has been dropped:

```
% ./bin/setup            
The Gemfile's dependencies are satisfied
yarn install v1.22.17
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.15s.
warning: parser/current is loading parser/ruby27, which recognizes2.7.6-compliant syntax, but you are running 2.7.4.
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
Created database 'name_of_app_development'
Created database 'name_of_app_test'
overwrite .env? (y/n [n]) y
```
Output of database migrations:

```
% rails db:migrate:status


database: name_of_app_development

 Status   Migration ID    Migration Name
--------------------------------------------------
   up     20220715191344  Create delayed jobs
```

PR #1099, opened 2022-07-15
Automate generator lookup in the development environment

thiagoa:
Problem
-------

Some facts

- Suspenders includes itself in the generated application's `Gemfile`

- Suspenders generators run in the context of the generated
applications

- Generated applications include Suspenders as a development dependency

Given the last item, if we're running Suspenders in development, we'd
end up with two copies of it: the development copy and a gem copy. The
problem is that:

- The development copy will start up Suspenders

- Generators, since they run in the context of an application, will be
picked up from the gem

Therefore, running `bin/suspenders` in development may put us in a
situation where the development copy has changes that the gem copy
does not, which means our code changes would not fully apply.

To prevent such a mismatch, the existing solution is to run Suspenders
with `--path`, which inserts a `path` option in the Gemfile:

```ruby
gem "suspenders", path: "..."
```

To activate this behavior, running `bin/suspenders` with a `--path`
option is necessary:

```sh
bin/suspenders myapp --path=$HOME/suspenders
```

That is not an intuitive solution and may trip up developers working
on Suspenders, which won't always remember to use `--path`. What if we
could infer when to use a `path` option in the Gemfile by detecting if
we're running in development? We can certainly do that.

`bin/suspenders` should just work regardless of the environment, and
we shouldn't remember to pass any extra options for it to work.

Solution
--------

- Delete the `--path` option, including tests

- Infer the current environment by checking if the
`suspenders.gemspec` file exists. If so, it certainly means we're in
the development environment.

### DIFF
--- a/spec/support/feature_test_helpers.rb
+++ b/spec/support/feature_test_helpers.rb
@@ -12,10 +12,9 @@ module FeatureTestHelpers
   end
 
   def run_suspenders(arguments = nil)
-    arguments = "--path=#{root_path} #{arguments}"
     run_in_tmp do
       EnvPath.prepend_env_path!(fake_bin_path)
-      command = "#{suspenders_bin} _#{rails_version}_ #{APP_NAME} #{arguments}"
+      command = "#{suspenders_bin} #{APP_NAME} #{arguments}"
 
       with_revision_for_honeybadger do
         result = `#{command}`
@@ -58,7 +57,7 @@ module FeatureTestHelpers
       EnvPath.prepend_env_path!(fake_bin_path)
 
       with_revision_for_honeybadger do
-        debug `#{system_rails_bin} _#{rails_version}_ new #{APP_NAME} --skip-spring -d postgresql -m #{rails_template_path}`
+        debug `#{system_rails_bin} new #{APP_NAME} --skip-spring -d postgresql -m #{rails_template_path}`
       end
 
       Dir.chdir(APP_NAME) do
@@ -125,10 +124,6 @@ module FeatureTestHelpers
 
   def rails_template_path
     root_path.join("spec", "support", "rails_template.rb")
-  end
-
-  def rails_version
-    Rails::VERSION::STRING
   end
 
   def commit_all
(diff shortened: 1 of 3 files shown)


thiagoa:
This issue is documented here: https://github.com/thoughtbot/suspenders/issues/1100

There's a solution for it, but it's not very intuitive. This PR can be reused to improve the current solution.

templates/Gemfile.erb
```diff
@@ -42,4 +42,8 @@ group :test do
   gem "webmock"
 end
 
+<% if SUSPENDERS_PATH.join("suspenders.gemspec").exist? %>
+gem "suspenders", path: "<%= SUSPENDERS_PATH %>", group: [:development, :test]
+<% else %>
```
stevepolitodesign, line 47: I wonder if there's any value in adding a comment that documents why we're doing this 🤔

```suggestion
<% if SUSPENDERS_PATH.join("suspenders.gemspec").exist? %>
# Use local install during develpoment
gem "suspenders", path: "<%= SUSPENDERS_PATH %>", group: [:development, :test]
<% else %>
```

stevepolitodesign (commented):
I think this looks good, I just had one small note regarding the `--path` argument.

spec/support/feature_test_helpers.rb
```diff
@@ -12,10 +12,9 @@ def remove_project_directory
   end
 
   def run_suspenders(arguments = nil)
-    arguments = "--path=#{root_path} #{arguments}"
     run_in_tmp do
       EnvPath.prepend_env_path!(fake_bin_path)
-      command = "#{suspenders_bin} _#{rails_version}_ #{APP_NAME} #{arguments}"
```
thiagoa, line 18: We the [latest changes](https://github.com/thoughtbot/suspenders/pull/1102), we no longer need to be explicit about the Rails version. It will just work.

thiagoa:
I got tripped up with `--path` at least three times, lol. No more lost debugging time.

mike-burns (approved):
Hey great solution!

PR #1103, opened 2022-07-22
Replace Delayed::Job with Sidekiq

### LINKED ISSUE #1074: Replace Delayed::Job with Sidekiq
`Delayed::Job` was [a choice we made back in 2013](https://github.com/thoughtbot/suspenders/pull/166) and it made sense at the time.

Nowadays, Sidekiq is a more mature solution, and the argument of `Delayed::Job` not requiring a new dependency to run no longer makes sense because it's easy to add Redis to any platform, and not particularly expensive.

Sidekiq is not perfect, but it has better performance (scales better) and more momentum in the Ruby community.

Also, it's harder to switch from `Delayed::Job` to `Sidekiq` later on when in need of a full-featured solution.

caiohsramos:
Closes #1074 

All jobs are enqueued to "default" queue. In Rails 6.1 we won't need the application.rb tweak for action_mailer, action_mailbox and active_storage (https://github.com/mperham/sidekiq/wiki/Active-Job#queues).

Using Sidekiq through ActiveJob is significantly slower than using pure Sidekiq workers (https://github.com/mperham/sidekiq/wiki/Active-Job#performance), maybe a migration would be nice in the future

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ It includes application gems like:
 * [Autoprefixer Rails](https://github.com/ai/autoprefixer-rails) for CSS vendor prefixes
 * [Bourbon](https://github.com/thoughtbot/bourbon) for Sass mixins
 * [Bitters](https://github.com/thoughtbot/bitters) for scaffold application styles
-* [Delayed Job](https://github.com/collectiveidea/delayed_job) for background
+* [Sidekiq](https://github.com/mperham/sidekiq) for background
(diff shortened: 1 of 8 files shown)


caiohsramos:
@thiagoa here is the PR :slightly_smiling_face:

thiagoa:
@caiohsramos Thanks a lot!

PR #1104, opened 2022-08-05
Support Rails 6.1.6.1

### LINKED ISSUE #1060: Update for Rails 6.1.0
Suspenders currently locks rails to `~> 6.0.0` on the latest version. It would be nice if there was a new version that allowed users of suspenders to update to the latest version of Rails.

stevepolitodesign:
This commit does the bare minimum to support Rails 6.1.6.1

Because the `gem` command is no longer reversible, we decided to override the method by borrowing as much from the [existing method](https://github.com/rails/rails/blob/04972d9b9ef60796dc8f0917817b5392d61fcf09/railties/lib/rails/generators/actions.rb#L22). The main difference being that we call `append_file` instead of `gsub_file`, so that we can preserve reversibility. However, https://github.com/thoughtbot/suspenders/pull/1066 provides a more elegant solution.

Closes https://github.com/thoughtbot/suspenders/issues/1060

### DIFF
--- a/lib/suspenders/actions.rb
+++ b/lib/suspenders/actions.rb
@@ -31,6 +31,27 @@ module Suspenders
       action ExpandJson.new(destination_root, file, data)
     end
 
+    def gem(*args)
+      options = args.extract_options!
+      name, *versions = args
+
+      parts = [quote(name)]
+
+      if (versions = versions.any? ? versions : options.delete(:version))
+        versions = Array(versions)
+        versions.each do |version|
+          parts << quote(version)
+        end
+      end
+
+      parts << quote(options) unless options.empty?
+
+      str = []
+      str << indentation
+      str << "gem #{parts.join(", ")}"
+      append_file "Gemfile", %(\n#{str.join}\n), verbose: false
+    end
+
     class ExpandJson
       def initialize(destination_root, file, data)
         @destination_root = destination_root
--- a/lib/suspenders/generators/jobs_generator.rb
+++ b/lib/suspenders/generators/jobs_generator.rb
@@ -3,7 +3,7 @@ require_relative "base"
 module Suspenders
   class JobsGenerator < Generators::Base
     def add_jobs_gem
-      append_file "Gemfile", %(\ngem "sidekiq"\n)
+      gem "sidekiq"
       Bundler.with_unbundled_env { run "bundle install" }
     end
 
--- a/templates/active_job.rb
+++ b/templates/active_job.rb
@@ -1,4 +1,5 @@
 require "active_job/logging"
+require "active_job/log_subscriber"
 
 ActiveSupport::Notifications.unsubscribe("enqueue.active_job")
 
--- a/lib/suspenders/app_builder.rb
+++ b/lib/suspenders/app_builder.rb
@@ -100,8 +100,8 @@ module Suspenders
 
     def setup_asset_host
       replace_in_file "config/environments/production.rb",
-        "# config.action_controller.asset_host = 'http://assets.example.com'",
-        'config.action_controller.asset_host = ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))'
+        "# config.asset_host = 'http://assets.example.com'",
+        'config.asset_host = ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))'
 
       if File.exist?("config/initializers/assets.rb")
         replace_in_file "config/initializers/assets.rb",
@@ -250,7 +250,7 @@ module Suspenders
     end
 
     def raise_on_missing_translations_in(environment)
-      config = "config.action_view.raise_on_missing_translations = true"
+      config = "config.i18n.raise_on_missing_translations = true"
 
       uncomment_lines("config/environments/#{environment}.rb", config)
     end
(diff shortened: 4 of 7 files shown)


mike-burns (commented):
What do you think of adding a `#gem` method to `Generators::Base`?

```ruby
def gem(name, **args)
  append_file "Gemfile", %(\n#{name}, #{args.inspect}\n"
end
```

(Or something like that.)

stevepolitodesign:
@mike-burns that works for me!

I ended up adding the method to the `Actions` module, since I _think_ any method definition in the `Base` class cannot accept any arguments since it's seen as a generator method (at least that's what the errors were implying).

I ended up borrowing a lot from the [existing method](https://github.com/rails/rails/blob/04972d9b9ef60796dc8f0917817b5392d61fcf09/railties/lib/rails/generators/actions.rb#L22), but let me know if this is overkill.

7f0329b729cbfba7434fc0ca090fe4f0394aec86

lib/suspenders/generators/jobs_generator.rb
```diff
@@ -3,7 +3,7 @@
 module Suspenders
   class JobsGenerator < Generators::Base
     def add_jobs_gem
-      append_file "Gemfile", %(\ngem "sidekiq"\n)
+      gem "sidekiq"
```
mike-burns, line 6: Beautiful!

mike-burns (approved):
Nice nice nice.

lib/suspenders/generators/jobs_generator.rb
stevepolitodesign: That was my fault. I was accidentally on `main` when I pushed [that up](https://github.com/thoughtbot/suspenders/commit/e6e07b0db3ef6ffd0bfd1c3f70628dc8cd144c31) 🤦.

Is it worth removing that commit before I merge this in? 

Also, I wonder if we should enable a [branch protection rule](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule) to prevent this from happening again?

mike-burns: Your choice!

lib/suspenders/actions.rb
```diff
@@ -31,6 +31,28 @@ def expand_json(file, data)
...
+      name, *versions = args
+
+      parts, message = [quote(name)], name.dup
+
+      if (versions = versions.any? ? versions : options.delete(:version))
+        versions = Array(versions)
+        versions.each do |version|
+          parts << quote(version)
+        end
+        message << " (#{versions.join(", ")})"
```
thiagoa, line 45: It looks like this variable is not being used. In the [original method](https://github.com/rails/rails/blob/6-1-stable/railties/lib/rails/generators/actions.rb#L21-L47), it is used for logging on this line:

```
log :gemfile, message
```

Do you think logging it would be valuable? Otherwise, we can delete it.

lib/suspenders/actions.rb
```diff
@@ -31,6 +31,28 @@ def expand_json(file, data)
...
+      end
+
+      parts << quote(options) unless options.empty?
+
+      str = []
+      str << indentation
```
thiagoa, line 51: Not a big deal, but I prefer how it reads in the original method:

```ruby
    str = "gem #{parts.join(", ")}"
    str = indentation + str
```

Also, the original method uses `in_root,` but I suppose that is not necessary here.

templates/active_job.rb
```diff
@@ -1,4 +1,5 @@
 require "active_job/logging"
+require "active_job/log_subscriber"
```
thiagoa, line 2: Is this also necessary for Rails 6.1?

thiagoa (approved):
Nice job, I'm really excited that we're getting closer to the latest Rails!

lib/suspenders/generators/jobs_generator.rb
thiagoa: I think branch protection would be great! As for the commit, it's small and harmless, so I think we can leave it as-is. This PR makes a change to the same line, and also improves it, so that's good.

templates/active_job.rb
stevepolitodesign: Oddly enough, yes. Without it, we get `uninitialized constant ActiveJob::Logging::LogSubscriber`.

lib/suspenders/actions.rb
```diff
@@ -31,6 +31,28 @@ def expand_json(file, data)
...
+      str << "gem #{parts.join(", ")}"
+      append_file "Gemfile", %(\n#{str.join}\n), verbose: false
```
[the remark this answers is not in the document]
stevepolitodesign: `append_file_with_newline` [calls gsub_file](https://github.com/rails/rails/blob/3da5f09d0e0f319e2a345e0e16d0e1e8bf63231c/railties/lib/rails/generators/actions.rb#L364) which unfortunately is not reversible.

lib/suspenders/actions.rb
stevepolitodesign: Good catch. Updated here: d55520a70085aa71632f708cdebd4269a09c3edd

lib/suspenders/actions.rb
stevepolitodesign: Oddly enough, I was getting failures when I used that implementation, but I'm not totally sure why 🤔

I think this might have also been the result of having run `standardrb --fix`.

PR #1105, opened 2022-08-19
Introduce Accessibility tooling Generator

seanpdoyle:
Create the `Suspenders::AccessibilityGenerator` to install
[capybara_accessibility_audit][] and [capybara_accessible_selectors][].

[capybara_accessibility_audit]: https://github.com/thoughtbot/capybara_accessibility_audit
[capybara_accessible_selectors]: https://github.com/citizensadvice/capybara_accessible_selectors

### DIFF
--- a/lib/suspenders/generators/accessibility_generator.rb
+++ b/lib/suspenders/generators/accessibility_generator.rb
@@ -1,0 +1,12 @@
+require_relative "base"
+
+module Suspenders
+  class AccessibilityGenerator < Generators::Base
+    def capybara_gems
+      gem "capybara_accessibility_audit", group: [:test]
+      gem "capybara_accessible_selectors", group: [:test],
+        github: "citizensadvice/capybara_accessible_selectors"
+      Bundler.with_unbundled_env { run "bundle install" }
+    end
+  end
+end
--- a/spec/suspenders/generators/accessibility_generator_spec.rb
+++ b/spec/suspenders/generators/accessibility_generator_spec.rb
@@ -1,0 +1,38 @@
+require "spec_helper"
+
+RSpec.describe Suspenders::AccessibilityGenerator, type: :generator do
+  describe "invoke" do
+    it "bundles the capybara_accessibility_audit gem" do
+      with_fake_app do
+        invoke! Suspenders::AccessibilityGenerator
+
+        expect("Gemfile")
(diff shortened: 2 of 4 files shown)


spec/suspenders/generators/accessibility_generator_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+    end
+
+    it "bundles the citizensadvice/capybara_accessible_selectors GitHub repository gem" do
+      with_fake_app do
+        invoke! Suspenders::AccessibilityGenerator
+
+        expect("Gemfile").to have_bundled("install").matching(/capybara_accessible_selectors/)
+      end
+    end
```
stevepolitodesign, line 19: [Minor] How would you feel about [testing for syntax errors](https://github.com/thoughtbot/suspenders/blob/099493117c6eaf2f74bec958477c50cb63de96b1/spec/support/generator_matchers.rb#L27) in case there's a regression? We do that in several of our existing generator tests.

https://github.com/thoughtbot/suspenders/blob/099493117c6eaf2f74bec958477c50cb63de96b1/spec/suspenders/generators/inline_svg_generator_spec.rb#L8-L10

stevepolitodesign (approved):
Thank you for adding this! I added a comment for you to consider, but nothing blocking.

lib/suspenders/generators/accessibility_generator.rb
```diff
@@ -0,0 +1,12 @@
+require_relative "base"
+
+module Suspenders
+  class AccessibilityGenerator < Generators::Base
```
thiagoa, line 4: @seanpdoyle Should we hook up this generator to `app_generator.rb`? https://github.com/thoughtbot/suspenders/blob/b9d702007f544ada5f26a02d63e989bc6815b40f/lib/suspenders/generators/app_generator.rb#L147 ?

seanpdoyle: Sure! I think that's a good idea. Where could we add test coverage for that behavior?

thiagoa: This should probably be tested at the system [here](https://github.com/thoughtbot/suspenders/blob/b9d702007f544ada5f26a02d63e989bc6815b40f/spec/features/new_project_spec.rb). This spec file works with a `before(:all)` hook, so it's not that expensive to add a new example.

PR #1106, opened 2022-08-20
Support Rails 7 and Ruby 3

stevepolitodesign:
# 🚀 Support Rails 7 and Ruby 3

This is a major upgrade that aims to make it so generated Rails applications with suspenders are running on Rails version 7 and Ruby version 3. We took the approach of making the least amount of changes possible.

Special thanks to @emilford, @stefannibrasil and @DoodlingDev for their help!

## 🏗 What changed?

- We removed obsolete and deprecated tooling around CSS #1112 
- We favor [cssbundling-rails](https://github.com/rails/cssbundling-rails) and [jsbundling-rails](https://github.com/rails/jsbundling-rails) over [importmap-rails](https://github.com/rails/importmap-rails) [in order to support more options around CSS frameworks](https://guides.rubyonrails.org/working_with_javascript_in_rails.html#choosing-between-import-maps-and-a-javascript-bundler).

## ❓What next?

1. We'll keep this pull-request open long enough to capture initial feedback. Feel free to add a `to-do` to the list below.
2. Once approved, we'll squash everything down into a single commit, since all these changes are related and can't easily be grouped.
3. Bump the version and update the README. 

## 🙏 How to contribute and review

This helps us determine if suspenders is generating an error free application.

1. Checkout the [support_rails_7](https://github.com/thoughtbot/suspenders/tree/support_rails_7) branch.
5. Once you are in the `suspenders` directory, run `./bin/suspenders some_app_name` to generate a Rails app from the current state of suspenders.
6. Change directories into `some_app_name`.
7. Run `rails g scaffold some_model some_attribute` followed by `rails db:migrate`.
8. Run `rails s`

Open up the app and ensure that it is loaded correctly. Try filing out the forms generated from the scaffold. If you run into an error, please follow the steps below.

1. Open a [new issue](https://github.com/thoughtbot/suspenders/issues/new).
2. Tag it as [Rails 7](https://github.com/thoughtbot/suspenders/labels/Rails%207)
4. Add it to the to-do section below. 

### ✅ To-do

- [x] Investigate `bin/setup` error. (`bin/setup: line 14: bin/yarn: No such file or directory`). Looks like it relates to #939 
- [x] Add normalize via PostCSS
- [x] Update CHANGELOG
- [x] Update README
- [x] Bump version
- [x] #1112 which is handled by #1113
- [x] #1116
- [x] #1111 
- [x] #1108
- [x] #1109


DoodlingDev:
I think for sure ruby 3, can't let the underlying runtime stagnate.

Generally I'd expect something like suspenders to have strong default opinions, with options if a project needs to deviate.

So probably... import maps and esbuild? webpacker's deprecated, and if we're moving to rails 7 it's not getting more attention than security issues.

stevehanson:
Thanks for getting this started! I feel like Suspenders being outdated is a big barrier for usage, so I lean toward finding simple solutions now to get Suspenders back up-to-date before spending too much time tackling larger decisions. With that in mind, what is the simplest path toward Rails 7 support?

Which JS engine is simplest to support for the sake of this PR? Rails defaults to `importmap` as the JavaScript engine but the `rails new` command allows passing a `--javascript` option, which can be one of `importmap (default), webpack, esbuild, rollup`. It'd be nice if Suspenders could support options like this eventually, since individual app requirements might make one option better than others in certain scenarios, though that's probably outside the scope of this PR.

Would it make sense to default to import maps for now just to match Rails? Then we can revisit this in a separate issue/PR? If apps need to use esbuild, it should be simple enough for now to follow the [installation steps](https://github.com/rails/jsbundling-rails#installation) after initializing the app.

I agree on defaulting to Ruby 3.

thiagoa:
I agree with @stevehanson that we should follow the path of least resistance to get this upgrade working. We could tackle the bigger decisions later. Maybe it would make sense to default to Webpack for some backward compatibility and avoiding more drastic changes?

As a future default, I vote on Hotwire/importmaps, and on updating the "API" option with newer defaults for React/Rails projects.

stevepolitodesign:
I was able to get a passing test suite when [upgrading to Ruby 3.0.0](https://github.com/thoughtbot/suspenders/pull/1106/commits/fabea62148e0f62e4d1bda8b8e79ee5dca269fa6), but [CI fails](https://github.com/thoughtbot/suspenders/actions/runs/3418310707/jobs/5691466001#step:7:30) on [3.1.2](https://github.com/thoughtbot/suspenders/pull/1106/commits/07e9c181744f09fd8f1b0f4ec639ddd19065d940).

I wonder if we just support 3.0.0 for how, and add support for the latest version later? I didn't have too much time to dig into this, but I think t[he error is related to the version of bundler and/or gems we're using](https://stackoverflow.com/a/72229541).

stefannibrasil:
Just a heads up:

```
/Users/stefannibrasil/.asdf/installs/ruby/3.0.0/lib/ruby/3.0.0/rubygems/dependency.rb:307:in `to_specs': Could not find 'rails' (~> 7.0.0) among 85 total gem(s) (Gem::MissingSpecError)
Checked in 'GEM_PATH=/Users/stefannibrasil/.gem/ruby/3.0.0:/Users/stefannibrasil/.asdf/installs/ruby/3.0.0/lib/ruby/gems/3.0.0' , execute `gem env` for more information
	from /Users/stefannibrasil/.asdf/installs/ruby/3.0.0/lib/ruby/3.0.0/rubygems/dependency.rb:319:in `to_spec'
	from /Users/stefannibrasil/.asdf/installs/ruby/3.0.0/lib/ruby/3.0.0/rubygems/specification.rb:974:in `find_by_name'
```

I then remembered I probably didn't have Rails 7 installed locally. After running `bundle install`, I was able to generate a Rails 7 app with suspenders.

.ruby-version
```diff
@@ -1 +1 @@
-2.7.4
+3.0.0
```
stevepolitodesign, line 1: We tried to support the latest version of Ruby, but ran into an issue c15cb3c9b3f924af3bbb7aca56ffd7f40dc836ee

lib/suspenders/generators/app_generator.rb
```diff
@@ -31,11 +31,8 @@ class AppGenerator < Rails::Generators::AppGenerator
     class_option :skip_system_test,
       type: :boolean, default: true, desc: "Skip system test files"
 
-    class_option :skip_turbolinks,
-      type: :boolean, default: true, desc: "Skip turbolinks gem"
-
-    class_option :skip_spring, type: :boolean, default: true,
-      desc: class_options[:skip_spring].description
+    class_option :css,
+      type: :string, default: "postcss", aliases: "-c", desc: "Choose CSS processor"
```
stevepolitodesign, line 35: I wonder if we need to update the description to load all options, similar to what we do with the database option.


https://github.com/thoughtbot/suspenders/blob/5624ca5903f2d1cf13074de668524a4ef05a7780/lib/suspenders/generators/app_generator.rb#L10-L11

stevepolitodesign: I also wonder if we should test the build by passing in an unsupported option.


```
suspenders some_app --css=no_supported
```

lib/suspenders/version.rb
```diff
@@ -1,5 +1,5 @@
 module Suspenders
-  RAILS_VERSION = "~> 6.1.6.1".freeze
+  RAILS_VERSION = "~> 7.0.0".freeze
```
stevepolitodesign, line 2: I wonder if we should the _latest_ version?

stevepolitodesign (commented):
Just a few things to consider.

lib/suspenders/version.rb
stevepolitodesign: We're good until `7.1.0`  because of `~> `

stevepolitodesign:
We'll need to bump the Gem version. Is this a good opportunity to bump to v2?

suspenders.gemspec
```diff
@@ -4,7 +4,7 @@ require "date"
 
 Gem::Specification.new do |s|
   s.required_ruby_version = ">= #{Suspenders::RUBY_VERSION}"
-  s.required_rubygems_version = ">= 3.0.0"
+  s.required_rubygems_version = ">= 3.1.2"
```
stevepolitodesign: This was a mistake. I thought I was bumping Ruby.

lib/suspenders/generators/app_generator.rb
stevepolitodesign: We tested this with `rails new` and Rails will throw an error.

suspenders.gemspec
stevepolitodesign: Since we're using `>=` my change is still acceptable, it just means that the minimum supported version has changed.

stevepolitodesign:
I think we can remove this line


https://github.com/thoughtbot/suspenders/blob/5624ca5903f2d1cf13074de668524a4ef05a7780/lib/suspenders/generators/app_generator.rb#L153

stefannibrasil (approved):
Alt Text

mike-burns:
> We'll need to bump the Gem version. Is this a good opportunity to bump to v2?

Is that a good time to move to calver instead of semver?

Since this project is a compilation of other projects, every release is a potential breaking change. Therefore, something like [calver](https://calver.org) might make more sense.

stevepolitodesign:
> > We'll need to bump the Gem version. Is this a good opportunity to bump to v2?
> 
> Is that a good time to move to calver instead of semver?
> 
> Since this project is a compilation of other projects, every release is a potential breaking change. Therefore, something like [calver](https://calver.org) might make more sense.

@mike-burns I'm open to this if you think it's a good recommendation. If we go with this approach, do we need to add any additional documentation noting the change in semantics?

mike-burns:
> If we go with this approach, do we need to add any additional documentation noting the change in semantics?

It should transparently work for current users, but it might be worth announcing via the blog.

lib/suspenders/generators/stylesheet_base_generator.rb
```diff
@@ -23,9 +23,8 @@ def install_bitters
       run "bitters install --path app/assets/stylesheets"
     end
 
-    # TODO: This is breaking the test suite
     def install_normalize_css
-      run "bin/yarn add normalize.css"
+      run "yarn add normalize.css"
```
stevepolitodesign: Note: This was working because we were using [Webpacker](https://github.com/rails/webpacker/blob/master/lib/install/bin/yarn).

stevepolitodesign:
@mike-burns how would I use CalVar to version this latest release? If we were to release on January 13, 2023 would I simply version it as `23.01.13`?

mike-burns:
Personally I'd version it as `20230113`. If anyone asks, that's the major, and the minor and patch are `0`.

lib/suspenders/version.rb
```diff
@@ -1,8 +1,8 @@
...
   RUBY_VERSION = IO
     .read("#{File.dirname(__FILE__)}/../../.ruby-version")
     .strip
     .freeze
-  VERSION = "1.56.1".freeze
+  VERSION = "20230113".freeze
```
stevepolitodesign, line 7: @mike-burns this good?

mike-burns: I think so. Honestly I haven't tried calver in a Ruby gem yet so I don't know what it'll do. But I hope this will work.

stevepolitodesign: Just noting, we opted for `yyymmdd.minor_version` to account for cases where we needed to release patches on the same day.

PR #1123, opened 2023-03-10
Upgrade Ruby version to 3.2.1

remy727:
### Why is this PR introduced?
Still using `3.0.5`. This upgrades Ruby version to `3.2.1`


thiagoa:
Side note: we are working on making the Ruby version range more flexible:

https://github.com/thoughtbot/suspenders/issues/1073

But this should provide a quick fix in the meantime.

stevepolitodesign:
Closed by #1135. We now run `3.3.1`

PR #1127, opened 2023-04-28
Fix brittle test

stevepolitodesign:
I was wondering why our Gemfile.lock file was ignored, so I looked at
the [commit][1] that introduced that change. It mentioned that:

> suspenders as a gem relies on version ranges for its dependencies
defined in its gemspec

This made me realize that the one failure we get over and over involves
standard.

It makes sense that CI was passing in the past, because standard was at
a different version.

This commit **originally** first locked the version to `1.9.0`, which is the
version we were using during the last successful build. From there, I stepped
through each [release][2] until I got a failure, which was at `1.26.0`. This
lead to me pessimistically lock the value so it will resolved to `1.25.5`, which
is the latest version to **not** break CI.

The fact that my tests were passing locally was probably because my
`Gemfile.lock` reflected an older standard version, but since deleting
`Gemfile.lock` I'm getting failures.

Instead of locking standard to a specific version to avoid test
failures, we update the test to account for the new command provided by
a recent [release][3]

[1]: https://github.com/thoughtbot/suspenders/commit/539b5c6e7fe68ee6ccad168ea6e267c7f12ae9bd
[2]: https://github.com/standardrb/standard/releases
[3]: https://github.com/standardrb/standard/releases/tag/v1.26.0


stefannibrasil (approved):
🎉

PR #1136, opened 2023-10-27
Introduce `suspenders:factories` generator

stevepolitodesign:
Maintains functionally with the [existing generator][] while adding
support for the [default Rails test suite]. With this change, the
generator can be invoked on a Rails application that use RSpec and the
default Rails test suite.

Adds generator which adds a test to lint all Factories in an effort to
improve developer experience.

Additionally, we remove the generation of the `dev:prime` task as we
felt that should be the responsibly of another generator.

## How to review this pull request

Review is easier when compared with the [existing generator][].

Additionally, if you want to manually test these changes, spin up a new rails app and add the following:

```ruby
group :development, :test do
  gem "suspenders", git: "https://github.com/thoughtbot/suspenders", branch: "factory-bot-generator"
end
```

[existing generator]: https://github.com/thoughtbot/suspenders/blob/main/lib/suspenders/generators/factories_generator.rb
[default Rails test suite]: https://guides.rubyonrails.org/testing.html

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ Unreleased
 * Introduce `suspenders:accessibility` generator
 * Introduce `Suspenders::Generators::APIAppUnsupported` module and concern
 * Introduce `suspenders:inline_svg` generator
+* Introduce `suspenders:factories` generator
 
 20230113.0 (January, 13, 2023)
 
--- a/lib/generators/suspenders/factories_generator.rb
+++ b/lib/generators/suspenders/factories_generator.rb
@@ -1,0 +1,66 @@
+module Suspenders
+  module Generators
+    class FactoriesGenerator < Rails::Generators::Base
+      include Suspenders::Generators::Helpers
+
+      source_root File.expand_path("../../templates/factories", __FILE__)
+      desc <<~TEXT
+        Build test data with clarity and ease.
+
+        This uses FactoryBot to help you define dummy and test data for your test
+        suite. The `create`, `build`, and `build_stubbed` class methods are directly
+        available to all tests.
+
+        We recommend putting FactoryBot definitions in one `spec/factories.rb` (or
+        `test/factories`) file, at least until it grows unwieldy. This helps reduce
+        confusion around circular dependencies and makes it easy to jump between
+        definitions.
+
+        Supports the default test suite and RSpec.
+      TEXT
+
+      def add_factory_bot
+        gem_group :development, :test do
+          gem "factory_bot_rails"
+        end
+
+        Bundler.with_unbundled_env { run "bundle install" }
+      end
+
+      def set_up_factory_bot
+        if default_test_helper_present?
+          insert_into_file Rails.root.join("test/test_helper.rb"), after: "class TestCase" do
+            "\n    include FactoryBot::Syntax::Methods"
+          end
+        elsif rspec_test_helper_present?
+          copy_file "factory_bot_rspec.rb", "spec/support/factory_bot.rb"
+          insert_into_file Rails.root.join("spec/rails_helper.rb") do
+            %(Dir[Rails.root.join("spec/support/**/*.rb")].sort.each { |file| require file })
+          end
+        end
+      end
+
+      def generate_empty_factories_file
+        if default_test_suite?
+          copy_file "factories.rb", "test/factories.rb"
+        elsif rspec_test_suite?
+          copy_file "factories.rb", "spec/factories.rb"
+        end
+      end
+
+      def remove_fixture_definitions
+        if default_test_helper_present?
+          comment_lines "test/test_helper.rb", /fixtures :all/
+        end
+      end
+
+      def create_linting_test
+        if default_test_suite?
+          copy_file "factories_test.rb", "test/factory_bots/factories_test.rb"
+        elsif rspec_test_suite?
+          copy_file "factories_spec.rb", "spec/factory_bots/factories_spec.rb"
+        end
+      end
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -30,6 +30,27 @@ Installs [capybara_accessibility_audit] and [capybara_accessible_selectors]
   [capybara_accessibility_audit]: https://github.com/thoughtbot/capybara_accessibility_audit
   [capybara_accessible_selectors]: https://github.com/citizensadvice/capybara_accessible_selectors
 
+### Factories
+
+Build test data with clarity and ease.
+
+This uses [FactoryBot] to help you define dummy and test data for your
+test suite. The `create`, `build`, and `build_stubbed` class methods are
+directly available to all tests.
+
+We recommend putting FactoryBot definitions in one `spec/factories.rb`
+(or `test/factories`) file, at least until it grows unwieldy. This helps reduce
+confusion around circular dependencies and makes it easy to jump between
+definitions.
+
+Supports the [default test suite] and [RSpec].
+
+`bin/rails g suspenders:factories`
+
+  [Factory Bot]: https://github.com/thoughtbot/factory_bot_rails
+  [default test suite]: https://guides.rubyonrails.org/testing.html
+  [RSpec]: https://rspec.info
+
 ### Inline SVG
 
 Render SVG images inline using the [inline_svg] gem, as a potential performance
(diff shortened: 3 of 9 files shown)


test/generators/suspenders/factory_bot_generator_test.rb
```diff
@@ -0,0 +1,215 @@
...
+
+module Suspenders
+  module Generators
+    class FactoryBotGenerator::DefaultTest < Rails::Generators::TestCase
+      include Suspenders::TestHelpers
+
+      tests Suspenders::Generators::FactoryBotGenerator
+      destination Rails.root
+      setup :prepare_destination
+      teardown :restore_destination
```
stevepolitodesign, line 12: `destination` is set to `test/dummy`, which is a blank Rails app.
`setup` and `teardown` handle adding or removing files needed for this specific test to run. For example, adding the `Gemfile`.

test/generators/suspenders/factory_bot_generator_test.rb
```diff
@@ -0,0 +1,215 @@
...
+          end
+        RUBY
+      end
+
+      def definition_file
+        <<~RUBY
+          FactoryBot.define do
+          end
+        RUBY
+      end
```
stevepolitodesign, line 134: I guess things like this could be placed in the `fixtures` directory, but for now I figured I'd just keep them inline. Maybe they should actually be set in the setup phase of the associated tests thought?

lib/generators/templates/factory_bot/dev.rake
```diff
@@ -0,0 +1,12 @@
...
+
+  namespace :dev do
+    desc "Sample data for local development environment"
+    task prime: "db:setup" do
+      include FactoryBot::Syntax::Methods
+
+      # create(:user, email: "user@example.com", password: "password")
+    end
+  end
+end
```
stevepolitodesign, line 12: Is this still an organizational convention? Would now be a good time to add this to the `./bin/setup` script too?

stevepolitodesign: Here’s some [context](https://thoughtbot.com/blog/priming-the-pump) as to why we use this.

Maybe now is a good opportunity to update the task description and link to the article.

lib/generators/suspenders/factory_bot_generator.rb
```diff
@@ -0,0 +1,70 @@
...
+        copy_file "dev.rake", "lib/tasks/dev.rake"
+      end
+
+      def remove_fixture_definitions
+        if default_test_helper_present?
+          comment_lines "test/test_helper.rb", /fixtures :all/
+        end
+      end
+
+      def create_linting_test
```
crackofdusk, line 45: Should we also lint factories when using RSpec?

lib/generators/templates/factory_bot/dev.rake
crackofdusk: I wouldn't expect the `dev:prime` task to come from the `factory_bot` generator, as you could very well have dev data built through a different mechanism. If it does exist, though, it would be nice to add the FactoryBot helper methods.

test/generators/suspenders/factory_bot_generator_test.rb
stevepolitodesign: I ended up moving these inline: 2bdf0d3cb21e13ea4ff4c558b55289e696910b69

lib/generators/suspenders/factory_bot_generator.rb
stevepolitodesign: Yes! Added here: 12b367f0af45fa20a1c7fc1c2a8ecc250b040214

lib/generators/templates/factory_bot/dev.rake
stevepolitodesign: I agree that this needs more thought. I was porting this over from the existing generator, but wonder if now is a good opportunity to re-think this.

For example, on my client project, we rely solely on the `./bin/setup` script, which simply uses `db/seeds.rb`. That file looks something like this:

```rb
require "factory_bot_rails"
require "webmock"

class Seeder
  include FactoryBot::Syntax::Methods

 def cal
  # create :user
 end

Seeder.new.call
```

stevepolitodesign: > The handling of development data feels out of scope in this pull request.

I agree. It feels like we could have a separate generator for development data as well as an improved `./bin/setup` script.

stevepolitodesign: 62eccbe0bd86318eb22d9f5ddb50c6bc384e4ad1

lib/generators/suspenders/factory_bot_generator.rb
```diff
@@ -0,0 +1,69 @@
+module Suspenders
+  module Generators
+    class FactoryBotGenerator < Rails::Generators::Base
+      source_root File.expand_path("../../templates/factory_bot", __FILE__)
+      desc "Installs and configures factory_bot_rails. Supports the default test suite and RSpec."
```
stevepolitodesign, line 5: I should use [this description](https://github.com/thoughtbot/suspenders/blob/main/templates/descriptions/factories.md)

mike-burns:
Why is this a new generator instead of a modification/replacement for `suspenders:factories`?

lib/generators/suspenders/factory_bot_generator.rb
stevepolitodesign: 6e793a9e3a48d387bb3fdb514c8ffab5ec21734c

stevepolitodesign:
> Why is this a new generator instead of a modification/replacement for `suspenders:factories`?

@mike-burns that was an oversite on my part. Fixed here: 1455dc140092bba9e0c5c454550b63c577837e5e

lib/generators/suspenders/factories_generator.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      def set_up_factory_bot
+        if default_test_helper_present?
+          insert_into_file Rails.root.join("test/test_helper.rb"), after: "class TestCase" do
+            "\n    include FactoryBot::Syntax::Methods"
+          end
+        elsif rspec_test_helper_present?
+          copy_file "factory_bot_rspec.rb", "spec/support/factory_bot.rb"
+          insert_into_file Rails.root.join("spec/rails_helper.rb") do
```
mike-burns, line 35: This is a common line in `rails_helper.rb`. Does `insert_into_file` avoid duplicating lines? Otherwise we might want to check that this line isn't already in there.

lib/generators/suspenders/factories_generator.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      private
+
+      def default_test_suite?
+        File.exist? Rails.root.join("test")
+      end
+
+      def rspec_test_suite?
+        File.exist? Rails.root.join("spec")
```
mike-burns, line 70: In the `suspenders:ci` generator, we [check against](https://github.com/thoughtbot/suspenders/blob/55cb5c246fe53aea4cf53436c7c208b40ad13e85/lib/suspenders/generators/ci_generator.rb#L29) `spec/spec_helper.rb` instead. I don't know whether one is correct, but I would like to keep this consistent.

stevepolitodesign: Good call out. Let's go with `spec/spec_helper.rb`. In an effort to keep things consistent, I extracted that into a helper.

6f436bf

lib/generators/suspenders/factories_generator.rb
stevepolitodesign: Looks like `insert_into_file` avoids duplication 👍 

e80f643

NEWS.md
```diff
@@ -3,6 +3,8 @@ Unreleased
 * Remove `suspenders` system executable
 * Introduce `suspenders:accessibility` generator
 * Introduce `Suspenders::Generators::APIAppUnsupported` module and concern
+* Introduce `suspenders:factories` generator
+* Introduce `Suspenders::Generators::Helpers` module
```
mike-burns, line 7: I read NEWS.md as an end user curious about what an upgrade gets me. Is the `Helpers` module something an end user would be interested in?

mike-burns (approved):

NEWS.md
stevepolitodesign: That's a good distinction. Addressed here: 7c84f1d442a3120d6af2b721a10a5c2ad4ac1938

I'll create a separate commit to remove `Suspenders::Generators::APIAppUnsupported`

PR #1137, opened 2023-10-31
Introduce `suspenders:accessibility` generator

stevepolitodesign:
Introduce `suspenders:accessibility` generator

Ported over from #1105

Installs [capybara_accessibility_audit] and
[capybara_accessible_selectors].

`./bin/rails g suspenders:accessibility`

Introduces `Suspenders::Generators::APIAppUnsupported` module for
generators that cannot be run in an [API only][] application.

This uses a [concern][] to ensure we raise an error before the generator
including the module invokes any of its methods.

## How to manually review this pull-request

If you want to manually run this generator, simply do the following:

1. `cd test/dummy`
1. `touch Gemfile`
1. `./bin/rails g suspenders:accessibility`

Confirm the `Gemfile` and `Gemfile.lock` contains the gems.

To test that it won't work on an API only application, comment out or remove
`config.api_only = true` from `test/dummy/config/application.rb` and run the
steps above

[capybara_accessibility_audit]:
https://github.com/thoughtbot/capybara_accessibility_audit
[capybara_accessible_selectors]:
https://github.com/citizensadvice/capybara_accessible_selectors
[API only]: https://guides.rubyonrails.org/api_app.html
[concern]: https://api.rubyonrails.org/classes/ActiveSupport/Concern.html

### DIFF
--- a/lib/generators/suspenders/accessibility_generator.rb
+++ b/lib/generators/suspenders/accessibility_generator.rb
@@ -1,0 +1,17 @@
+module Suspenders
+  module Generators
+    class AccessibilityGenerator < Rails::Generators::Base
+      include Suspenders::Generators::APIAppUnsupported
+
+      desc "Installs capybara_accessibility_audit and capybara_accessible_selectors"
+
+      def add_capybara_gems
+        gem_group :test do
+          gem "capybara_accessibility_audit"
+          gem "capybara_accessible_selectors", github: "citizensadvice/capybara_accessible_selectors"
+        end
+        Bundler.with_unbundled_env { run "bundle install" }
+      end
+    end
+  end
+end
--- a/lib/suspenders/generators.rb
+++ b/lib/suspenders/generators.rb
@@ -1,0 +1,30 @@
+require "active_support/concern"
+
+module Suspenders
+  module Generators
+    module APIAppUnsupported
+      class Error < StandardError
+        def message
+          "This generator cannot be used on API only applications."
+        end
+      end
+
+      extend ActiveSupport::Concern
+
+      included do
+        def raise_if_api_only_app
+          if api_only_app?
+            raise Suspenders::Generators::APIAppUnsupported::Error
+          end
+        end
+      end
+
+      private
+
+      def api_only_app?
+        File.read(Rails.root.join("config/application.rb"))
+          .match?(/^\s*config\.api_only\s*=\s*true/i)
+      end
+    end
+  end
+end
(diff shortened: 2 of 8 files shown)


stevepolitodesign:
I think I'll want to no-op this generator for API only applications.

lib/generators/suspenders/accessibility_generator.rb
```diff
@@ -0,0 +1,21 @@
+module Suspenders
+  module Generators
+    class AccessibilityGenerator < Rails::Generators::Base
+      desc "Installs capybara_accessibility_audit and capybara_accessible_selectors"
+
+      def raise_if_api_only_app
+        if File.read(Rails.root.join("config/application.rb")).match?(/^\s*config\.api_only\s*=\s*true/i)
+          raise Suspenders::Generators::APIUnsupportedError
+        end
+      end
```
stevepolitodesign, line 10: I wonder if we should extract this?

lib/suspenders/generators.rb
```diff
@@ -0,0 +1,9 @@
+module Suspenders
+  module Generators
+    class APIUnsupportedError < StandardError
```
stevepolitodesign, line 3: How do we feel about this name?

lib/suspenders/generators.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      extend ActiveSupport::Concern
+
+      included do
+        def raise_if_api_only_app
+          if api_only_app?
+            raise Suspenders::Generators::APIAppUnsupported::Error
+          end
+        end
+      end
```
stevepolitodesign, line 20: This will run first when we call `include Suspenders::Generators::APIAppUnsupported` in a generator. This is important because we don't want to invoke any of the generator's methods.

test/generators/suspenders/accessibility_generator_test.rb
```diff
@@ -0,0 +1,89 @@
...
+      tests Suspenders::Generators::AccessibilityGenerator
+      destination Rails.root
+      setup :prepare_destination
+      teardown :restore_destination
+
+      test "generator runs without errors" do
+        assert_nothing_raised do
+          run_generator
+        end
+      end
```
seanpdoyle, line 18: I'm always skeptical of the value of tests that use `assert_nothing_raised`. The absence of an error a fairly weak assertion to make outright, especially when there are other tests in the same file that make assertions about behavior and side effects that bake in this assertion, since they execute and pass without Exceptions.

seanpdoyle (approved):

PR #1107, opened 2022-11-22
Lint HTML Templates with ERBLint and BetterHtml

seanpdoyle:
Configure the [erb_lint][] utility to apply our `standard`-backed RuboCop linting tool to our application's `.erb` files.

The initial `.erb-lint.yml` file is configured according to the gem's [README.md][]. Additionally, there are some key RuboCop rules that we're disabling [according to some feedback from the community of `erb_lint` users][issue].

The Rails tasks defined in `lib/tasks/erblint.rake` are inspired by those defined in [standard/rake.rb][]. They define the `erblint:autocorrect` and `erblint` tasks to execute the `erblint` command with and without the `--autocorrect` flag. Additional arguments can be passed in with the `ERBLINTOPTS` environment variable.

By default, executing `rails standard` will execute `rails erblint`, and `rails standard:fix` will execute `rails erblint:autocorrect`.

Also depend on the [erblint-github][] gem to provide additional, accessibility-focused rules for how we statically analyze ERB templates.

Finally, the [ERBLint][] tool depends on [BetterHtml][], and claims to execute BetterHtml's [ERBSafety][] test suite. In our experience, this isn't the case, and [requires manual configuration][better_html_tests].

This commit also introduces the `config/better_html.yml` configuration file, and makes sure that `BetterHtml` and `ERBLint` consistently read its contents.

[erb_lint]: https://github.com/Shopify/erb-lint
[README.md]: https://github.com/Shopify/erb-lint#configuration
[standard/rake.rb]: https://github.com/testdouble/standard/blob/main/lib/standard/rake.rb
[exe/erblint]: https://github.com/Shopify/erb-lint/blob/main/exe/erblint
[issue]: Shopify/erb-lint#222 (comment)
[erblint-github]: https://github.com/github/erblint-github
[ERBLint]: https://github.com/Shopify/erb-lint
[ERBSafety]: https://github.com/Shopify/erb-lint#erbsafety
[BetterHtml]: https://github.com/Shopify/better-html
[better_html_tests]: https://github.com/Shopify/better-html#testing-for-valid-html-and-erb

PR #1138, opened 2023-11-06
Introduce `suspenders:advisories` generator

stevepolitodesign:
Uses the [bundler-audit][] gem to update the local security database and
show any relevant issues with the app's dependencies. This generator is
only responsible for installing the gem.

The [original implementation][] was written in 2014, and is no longer
relevant. This is because the gem ships [with a Rake task][] that can be
set as the default task, which will be addressed in #1144

Also exposes `backup_file` and `restore_file` test helpers into the
public API.

[bundler-audit]: https://github.com/rubysec/bundler-audit
[original implementation]: https://github.com/thoughtbot/suspenders/commit/e23157ede1477a6e3e1163516281cfda10b479c6
[with a Rake task]: https://github.com/rubysec/bundler-audit#rake-tasks

### DIFF
--- a/lib/generators/suspenders/advisories_generator.rb
+++ b/lib/generators/suspenders/advisories_generator.rb
@@ -1,0 +1,27 @@
+module Suspenders
+  module Generators
+    class AdvisoriesGenerator < Rails::Generators::Base
+      source_root File.expand_path("../../templates/advisories", __FILE__)
+      desc(<<~TEXT)
+        Show security advisories during development.
+
+        Uses the `bundler-audit` gem to update the local security database and
+        show any relevant issues with the app's dependencies via a Rake task.
+      TEXT
+
+      def add_bundler_audit
+        gem_group :development, :test do
+          gem "bundler-audit", ">= 0.7.0", require: false
+        end
+        Bundler.with_unbundled_env { run "bundle install" }
+      end
+
+      def modify_rakefile
+        insert_into_file "Rakefile", "\nrequire \"bundler/audit/task\"",
+          after: 'require_relative "config/application"'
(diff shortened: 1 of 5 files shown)


lib/generators/suspenders/advisories_generator.rb
```diff
@@ -0,0 +1,26 @@
...
+        gem_group :development, :test do
+          gem "bundler-audit", ">= 0.7.0", require: false
+        end
+        Bundler.with_unbundled_env { run "bundle install" }
+      end
+
+      def configurea_rake_task
+        copy_file "bundler_audit.rake", "lib/tasks/bundler_audit.rake"
+        append_file "Rakefile", %(\ntask default: "bundle:audit"\n)
+      end
```
stevepolitodesign, line 23: This might require more thought. What if an application already has a default task? Additionally, what if we want to modify what we consider to be our default task in the future? I could imagine we'd want to run `standard` as part of our default task too?

stevepolitodesign: Now that these generators can be run independently, it can be challenging to piece them together. I wonder if instead we want a `rake` generator instead? Something that does the following:

- Runs `bundle:audit`
- Runs `standard`
  - [Lints HTML and ERB](https://github.com/thoughtbot/suspenders/pull/1107) is it's not an API application

This could also dovetail nicely with an CI generator, since I think there's value in keeping parity with the default Rake task and what's run in CI.

stevepolitodesign: We'll address this in #1144. For now, this generator will be responsible for installing the gem only.

bd01ede2bcebeff932ba19ed9907b4dbda36085e

PR #1142, opened 2023-11-10
Add arm64-darwin-22 platform to Gemfile.lock

dorianmarie:


Gemfile.lock
```diff
@@ -231,6 +231,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-21
+  arm64-darwin-22
```
stevepolitodesign, line 234: Just out of curiosity, what was the inspiration for this change?

crackofdusk: Running `bundle install` adds this line. I had the same behaviour as Dorian. I think it depends on the macOS version and/or laptop version.

It's a little annoying to have to skip this change when you commit any other gem change, so we decided to add the platform.

I think there's a way to avoid all of this by adding a more generic `ruby` platform.

crackofdusk: Confirmed: if I manually change the platforms list to the following, `bundle install` does not modify it.

```
PLATFORMS
  ruby
```

stevepolitodesign: Thank you for debugging that. Do you think it's worth just setting the `PLATFORMS` to `ruby` then?

crackofdusk: Let's do that. It will reduce churn.

dorianmarie: I updated it to just ruby, makes sense

stevepolitodesign (approved):
Thank you! I tested this locally by removing `Gemfile.lock` and running `bin/setup` and then running `bundle exec rake`. Everything still works.

PR #1144, opened 2023-11-11
Introduce `suspenders:rake` generator

stevepolitodesign:
Add necessary files to make the [plugin][] an [engine][], which
automatically loads Rake tasks located in `lib/tasks`. This means when
the `suspenders` gem is installed, the consumer can run `bundle exec
rake suspenders:rake`, and any future tasks.

Because `suspenders:lint` and `suspenders:advisories` may not
necessarily have been invoked, we need to check if those gems are
installed.

[plugin]: https://guides.rubyonrails.org/plugins.html
[engine]: https://guides.rubyonrails.org/engines.html

## How to review this pull request

1. Spin up a new Rails application.

    ```sh
    rails new rake_demo
    ```

1. Scaffold a domain in order to generate passing tests.

    ```sh
    bin/rails g scaffold post title
    bin/rails db:migrate
    ```

1. Install suspenders on this branch.

    ```ruby
    group :development, :test do
      gem "suspenders", github: "thoughtbot/suspenders", branch: "suspenders-3-0-0-rake-generator"
    end
    ```

1. Run the following generators.

    ```sh
    bin/rails g suspenders:advisories
    bin/rails g suspenders:lint
    bin/rails g suspenders:rake
    ```

1. Run Rake and note the output. It should run the test suite, run `standard`
   and run `bundler-audit`.

    ```sh
    bundle exec rake
    ```

Co-authored-by: Mike Burns <mburns@thoughtbot.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,16 @@ Installs [Sidekiq][] for background job processing and configures ActiveJob for 
 
   [Sidekiq]: https://github.com/sidekiq/sidekiq
 
+### Rake
+
+Configures the default Rake task to audit and lint the codebase with
+[bundler-audit][] and [standard][] in addition to running the test suite.
+
+`bin/rails g suspenders:rake`
+
+  [bundler-audit]: https://github.com/rubysec/bundler-audit
+  [standard]: https://github.com/standardrb/standard
+
 ## Contributing
 
 See the [CONTRIBUTING] document.
--- a/lib/generators/suspenders/rake_generator.rb
+++ b/lib/generators/suspenders/rake_generator.rb
@@ -1,0 +1,15 @@
+module Suspenders
+  module Generators
+    class RakeGenerator < Rails::Generators::Base
+      source_root File.expand_path("../../templates/rake", __FILE__)
+      desc(<<~TEXT)
+        Configures the default Rake task to audit and lint the codebase with
+        `bundler-audit` and `standard`, in addition to running the test suite.
+      TEXT
+
+      def configure_default_rake_task
+        append_to_file "Rakefile", %(task default: "suspenders:rake")
+      end
+    end
+  end
+end
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@ Unreleased
 * Introduce `suspenders:styles` generator
 * Introduce `suspenders:jobs` generator
 * Introduce `suspenders:lint` generator
+* Introduce `suspenders:rake` generator
 
 20230113.0 (January, 13, 2023)
 
--- a/bin/rails
+++ b/bin/rails
@@ -3,6 +3,7 @@
 # installed from the root of your application.
 
 ENGINE_ROOT = File.expand_path("..", __dir__)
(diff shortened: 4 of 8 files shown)


lib/generators/suspenders/rake_generator.rb
```diff
@@ -0,0 +1,25 @@
...
+          gem "standard"
+        end
+        Bundler.with_unbundled_env { run "bundle install" }
+      end
+
+      def configurea_rake_task
+        copy_file "bundler_audit.rake", "lib/tasks/bundler_audit.rake"
+
+        insert_into_file "Rakefile", "\nrequire \"standard/rake\"", after: "require_relative \"config/application\""
+        append_file "Rakefile", %(\ntask default: %i[bundle:audit standard]\n)
```
stevepolitodesign, line 21: Not sure if it's worth accounting for RSpec like we [currently](https://github.com/thoughtbot/suspenders/blob/55cb5c246fe53aea4cf53436c7c208b40ad13e85/lib/suspenders/app_builder.rb#L231) do?

mike-burns:
I think we're running into a flexibility issue.

What if we made a default rake task that looks for whether bundler-audit is installed and run it if so, ignore it otherwise? I have no idea whether that's possible, but it would allow for the kind of extensibility we need.

in lib/suspenders/tasks.rb

```ruby
task :suspenders do
  if Bundler.rubygems.find_name("bundle-audit").any?
    Rake::Task[:"bundle:audit"].run
  end

  if Bundler.rubygems.find_name("standard").any?
    Rake::Task[:standard].run
  end
end
```

generated Rakefile

```ruby
require 'suspenders/tasks'
default: [:suspenders]
```

lib/generators/suspenders/rake_generator.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      def add_bundler_audit_and_standard
+        gem_group :development, :test do
+          gem "bundler-audit", ">= 0.7.0", require: false
+          gem "standard"
+        end
+        Bundler.with_unbundled_env { run "bundle install" }
+      end
+
+      def configurea_rake_task
```
crackofdusk, line 17:
```suggestion
      def configure_a_rake_task
```
?

lib/generators/suspenders/rake_generator.rb
stevepolitodesign: We don't need to include RSpec in the Rakefile. I tested with a simple app using RSpec, and by default, running `bundle exec rake` runs RSpec. This was still true after having invoked `suspenders:rake`, except that `standard` and `bundler-audit` were also invoked.

README.md
```diff
@@ -30,6 +30,7 @@ Installs [capybara_accessibility_audit] and [capybara_accessible_selectors]
   [capybara_accessibility_audit]: https://github.com/thoughtbot/capybara_accessibility_audit
   [capybara_accessible_selectors]: https://github.com/citizensadvice/capybara_accessible_selectors
 
+<<<<<<< HEAD
```
crackofdusk, line 33: Ther's a leftover marker from a rebase or a merge here.

PR #1145, opened 2023-11-22
Introduce `suspenders:styles` generator

stevepolitodesign:
Configures applications to use [PostCSS][1] or [Tailwind][2] via
[cssbundling-rails][3]. Defaults to PostCSS with [modern-normalize][8], with
the option to override via `--css=tailwind`. These options were pulled from the
[supported list of options][4] in Rails.

Also creates additional stylesheets if using PostCSS.

We choose to use [cssbundling-rails][4] instead of [dartsass-rails][5] or
[tailwindcss-rails][6] (or even just css) because we want to rely on Node to
process the CSS. Although we could have chosen to avoid using Node altogether,
we feel it's better to support it since we'll need it for additional generators,
like [StyleLintGenerator][7], and to support [modern-normalize][8].

Updates `within_api_only_app` by allowing support to conditionally comment out
the api configuration. This provided and opportunity to clean up existing setup
steps.

[1]: https://postcss.org
[2]: https://tailwindcss.com
[3]: https://github.com/rails/cssbundling-rails
[4]: https://github.com/rails/rails/blob/438cad462638b02210fc48b700c29dcd0428a8b7/railties/lib/rails/generators/app_base.rb#L22
[5]: https://github.com/rails/dartsass-rails
[6]: https://github.com/rails/tailwindcss-rails
[7]: https://github.com/thoughtbot/suspenders/blob/main/lib/suspenders/generators/stylelint_generator.rb
[8]: https://github.com/sindresorhus/modern-normalize
[9]: https://tailwindcss.com/docs/functions-and-directives#layer

### DIFF
--- a/lib/generators/suspenders/styles_generator.rb
+++ b/lib/generators/suspenders/styles_generator.rb
@@ -1,0 +1,60 @@
+module Suspenders
+  module Generators
+    class StylesGenerator < Rails::Generators::Base
+      include Suspenders::Generators::APIAppUnsupported
+
+      CSS_OPTIONS = %w[tailwind postcss].freeze
+
+      class_option :css, enum: CSS_OPTIONS, default: "postcss"
+      desc <<~TEXT
+        Configures applications to use PostCSS or Tailwind via cssbundling-rails.
+        Defaults to PostCSS with modern-normalize, with the option to override via
+        --css=tailwind.
+
+        Also creates a directory structure to store additional stylesheets if using
+        PostCSS.
+      TEXT
+
+      def add_cssbundling_rails_gem
+        gem "cssbundling-rails"
+
+        Bundler.with_unbundled_env { run "bundle install" }
+        run "bin/rails css:install:#{css}"
+      end
+
+      def build_directory_structure
+        return if is_tailwind?
+
+        create_file "app/assets/stylesheets/base.css"
+        create_file "app/assets/stylesheets/components.css"
+        create_file "app/assets/stylesheets/utilities.css"
+      end
+
+      # Modify if https://github.com/rails/cssbundling-rails/pull/139 is merged
+      def configure_application_stylesheet
+        return if is_tailwind?
+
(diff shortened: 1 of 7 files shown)


lib/generators/suspenders/styles_generator.rb
```diff
@@ -0,0 +1,52 @@
...
+      def build_directory_structure
+        empty_directory "app/assets/stylesheets/base"
+        create_file "app/assets/stylesheets/base/.gitkeep"
+
+        empty_directory "app/assets/stylesheets/components"
+        create_file "app/assets/stylesheets/components/.gitkeep"
+
+        empty_directory "app/assets/stylesheets/utilities"
+        create_file "app/assets/stylesheets/utilities/.gitkeep"
+      end
```
stevepolitodesign, line 43: Do we want to do this for applications running Tailwind?

lib/generators/suspenders/styles_generator.rb
stevepolitodesign, line 1: Should apps using `css` or `sass` leverage [modern-normalize](https://github.com/sindresorhus/modern-normalize)?

If "yes", how do we want to include it, since both options do not require NPM. We could use `curl` during installation, or we could just keep a local copy in the generator and copy it over. In either case, we run the risk of drifting.

stevepolitodesign: I think we'll want to use node since the [StylelintGenerator](https://github.com/thoughtbot/suspenders/blob/main/lib/suspenders/generators/stylelint_generator.rb) requires it anyways.

This means we might want to go back to `cssbundling_rails`

lib/generators/suspenders/styles_generator.rb
stevepolitodesign: We'll keep these since Tailwind supports [@layer directives](https://tailwindcss.com/docs/functions-and-directives#layer).

PR #1147, opened 2023-12-01
Introduce `suspenders:jobs` generator

laicuRoot:
Ports the existing generator to Suspenders 3.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,15 @@ Also creates additional stylesheets if using PostCSS.
   [cssbundling-rails]: https://github.com/rails/cssbundling-rails
   [modern-normalize]: https://github.com/sindresorhus/modern-normalize
 
+
+### Jobs
+
+Installs [Sidekiq][] for background job processing and configures ActiveJob for job queueing.
+
+`bin/rails g suspenders:jobs`
+
+  [Sidekiq]: https://github.com/sidekiq/sidekiq
+
 ## Contributing
 
 See the [CONTRIBUTING] document.
--- a/lib/generators/suspenders/jobs_generator.rb
+++ b/lib/generators/suspenders/jobs_generator.rb
@@ -1,0 +1,36 @@
+module Suspenders
+  module Generators
+    class JobsGenerator < Rails::Generators::Base
+      source_root File.expand_path("../../templates/active_job", __FILE__)
+      desc "Installs Sidekiq for background job processing."
+
+      def add_sidekiq_gem
+        gem "sidekiq"
+        Bundler.with_unbundled_env { run "bundle install" }
+      end
+
+      def initialize_active_job
+        copy_file "active_job.rb", "config/initializers/active_job.rb"
+      end
+
+      def configure_active_job
+        environment "config.active_job.queue_adapter = :sidekiq"
+        environment "config.action_mailer.deliver_later_queue_name = nil"
+        environment "config.action_mailbox.queues.routing = nil"
+        environment "config.active_storage.queues.analysis = nil"
+        environment "config.active_storage.queues.purge = nil"
+        environment "config.active_storage.queues.mirror = nil"
+        environment "config.active_job.queue_adapter = :inline", env: "test"
+      end
+
+      def configure_procfile
+        if Rails.root.join("Procfile.dev").exist?
+          append_to_file "Procfile.dev", "worker: bundle exec sidekiq"
+        else
+          say "Add default Procfile.dev"
+          create_file "Procfile.dev", "worker: bundle exec sidekiq"
+        end
+      end
+    end
+  end
+end
--- a/lib/generators/templates/active_job/active_job.rb
+++ b/lib/generators/templates/active_job/active_job.rb
@@ -1,0 +1,14 @@
+require "active_job/logging"
+require "active_job/log_subscriber"
+
+ActiveSupport::Notifications.unsubscribe("enqueue.active_job")
+
+module ActiveJob
+  module Logging
+    class EnqueueLogSubscriber < LogSubscriber
+      define_method :enqueue, instance_method(:enqueue)
+    end
+  end
+end
+
+ActiveJob::Logging::EnqueueLogSubscriber.attach_to(:active_job)
--- a/test/generators/suspenders/jobs_generator_test.rb
+++ b/test/generators/suspenders/jobs_generator_test.rb
@@ -1,0 +1,122 @@
+require "test_helper"
+require "generators/suspenders/jobs_generator"
+
+module Suspenders
+  module Generators
+    class JobsGeneratorTest < Rails::Generators::TestCase
+      include Suspenders::TestHelpers
+
+      tests Suspenders::Generators::JobsGenerator
+      destination Rails.root
+      setup :prepare_destination
+      teardown :restore_destination
+
+      test "adds gems to Gemfile" do
+        expected_output = <<~RUBY
+          gem "sidekiq"
+        RUBY
+
+        run_generator
+
+        assert_file app_root("Gemfile") do |file|
+          assert_match(expected_output, file)
+        end
+      end
+
+      test "installs gems with Bundler" do
+        output = run_generator
+
+        assert_match(/bundle install/, output)
+      end
+
+      test "generator has a description" do
+        description = "Installs Sidekiq for background job processing."
+
+        assert_equal description, Suspenders::Generators::JobsGenerator.desc
+      end
+
+      test "configures ActiveJob logging" do
+        expected_configuration = <<~RUBY
+          require "active_job/logging"
+          require "active_job/log_subscriber"
+
+          ActiveSupport::Notifications.unsubscribe("enqueue.active_job")
+
+          module ActiveJob
+            module Logging
+              class EnqueueLogSubscriber < LogSubscriber
+                define_method :enqueue, instance_method(:enqueue)
+              end
+            end
+          end
+
+          ActiveJob::Logging::EnqueueLogSubscriber.attach_to(:active_job)
+        RUBY
+
+        run_generator
+
+        assert_file app_root("config/initializers/active_job.rb") do |file|
+          assert_equal(expected_configuration, file)
+        end
+      end
+
+      test "adds ActiveJob configuration to the application file" do
+        run_generator
+
+        assert_file app_root("config/application.rb") do |file|
+          assert_match(/config.active_job.queue_adapter = :sidekiq/, file)
+          assert_match(/config.action_mailer.deliver_later_queue_name = nil/, file)
+          assert_match(/config.action_mailbox.queues.routing = nil/, file)
+          assert_match(/config.active_storage.queues.analysis = nil/, file)
+          assert_match(/config.active_storage.queues.purge = nil/, file)
+          assert_match(/config.active_storage.queues.mirror = nil/, file)
+        end
+      end
+
+      test "adds ActiveJob configuration to the test environment file" do
+        run_generator
+
+        assert_file app_root("config/environments/test.rb") do |file|
+          assert_match(/config.active_job.queue_adapter = :inline/, file)
+        end
+      end
+
+      test "creates a Procfile.dev with Sidekiq configuration" do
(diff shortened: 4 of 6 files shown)


lib/generators/suspenders/jobs_generator.rb
```diff
@@ -0,0 +1,55 @@
+module Suspenders
+  module Generators
+    class JobsGenerator < Rails::Generators::Base
+      include Suspenders::Generators::APIAppUnsupported
```
stevepolitodesign, line 4: We can actually configure API only applications to use Sidekiq, since [Active Job](https://edgeguides.rubyonrails.org/active_job_basics.html) does not depend on views. For example, Active Job is responsible for [delivering emails](https://edgeguides.rubyonrails.org/active_job_basics.html#action-mailer), which is something an API only application is capable of doing.

lib/generators/suspenders/jobs_generator.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      def configure_active_job
+        configure_application_file("config.active_job.queue_adapter = :sidekiq")
+        configure_application_file("config.action_mailer.deliver_later_queue_name = nil")
+        configure_application_file("config.action_mailbox.queues.routing = nil")
+        configure_application_file("config.active_storage.queues.analysis = nil")
+        configure_application_file("config.active_storage.queues.purge = nil")
+        configure_application_file("config.active_storage.queues.mirror = nil")
+        configure_environment "test", "config.active_job.queue_adapter = :inline"
```
stevepolitodesign, line 24: I think we can leverage the [environment](https://guides.rubyonrails.org/rails_application_templates.html#environment-application-data-nil-options-block) command here, which would alleviate us from having to create a custom method.

test/generators/suspenders/jobs_generator_test.rb
```diff
@@ -0,0 +1,176 @@
...
+      end
+
+      test "installs gems with Bundler" do
+        Bundler.stubs(:with_unbundled_env).yields
+        generator.expects(:run).with("bundle install").once
+
+        capture(:stdout) do
+          generator.add_sidekiq_gem
+        end
+      end
```
stevepolitodesign, line 57: [Minor]

After exploring the [Rails codebase](https://github.com/rails/rails/blob/37833debc669bd3435a0d70b1d294bda2eb14070/railties/test/generators/app_generator_test.rb#L647-L651) for inspiration on how they test bundle commands, I've [started to follow the same pattern](https://github.com/thoughtbot/suspenders/blob/73cdb80fb1f1bd3d11b5b87f5e00e0015206a541/test/generators/suspenders/styles_generator_test.rb#L36-L40). My hope is that we might be able to remove the `mocha` dependency.

```suggestion
      test "installs gems with Bundler" do
        output = run_generator
        
        assert_match /bundle install/, output
      end
```

stevepolitodesign (commented):
Thank you for taking this one on! I recognize this is a draft, but wanted to share some initial thoughts.

lib/generators/suspenders/jobs_generator.rb
laicuRoot: Thank you so much for the explanation @stevepolitodesign. 

I meant to write you a message in slack before creating the PR just saying that I was working on this generator by just following the patterns I was seeing in other generators and trying to learn along the way 🙂 but I got carried away.

lib/generators/templates/active_job/active_job.rb
```diff
@@ -0,0 +1,14 @@
+require "active_job/logging"
+require "active_job/log_subscriber"
+
+ActiveSupport::Notifications.unsubscribe("enqueue.active_job")
+
+module ActiveJob
+  module Logging
+    class EnqueueLogSubscriber < LogSubscriber
+      define_method :enqueue, instance_method(:enqueue)
```
crackofdusk, line 9: I think this doesn't work in Rails 7.1.

Have you tested the generator "manually" on a new application?

laicuRoot: Thank you for the comment @crackofdusk ! It has made me investigate more and try the generator in a new app. 
It looks like it’s working, but please let me know if you think otherwise. 

I took the following steps to test it:

- I created a new Rails 7.1.2 App
- I added the generator to this app.
- I added the template `active_job.rb`
- I ran in my terminal `bin/rails generate jobs`

- This has generated the following
    
    ```ruby
    gemfile  sidekiq
             run  bundle install from "."
    Bundle complete! 16 Gemfile dependencies, 85 gems now installed.
    Use `bundle info [gemname]` to see where a bundled gem is installed.
          create  config/initializers/active_job.rb
    Add default Procfile.dev
          create  Procfile.dev
    ```
- I can see the `sidekiq` gem in the Gemfile, the `active_job` file in initializers, and the `Procfile` with the expected configuration.
- Then I create a job - `MyJob`

I modified the initializer to: 
```
require "active_job/logging"
require "active_job/log_subscriber"


# ActiveSupport::Notifications.unsubscribe("enqueue.active_job")
Rails.logger.info "Custom Active Job initializer is running"

module ActiveJob
  module Logging
    class EnqueueLogSubscriber < LogSubscriber
      puts "ActiveJob::Logging::EnqueueLogSubscriber"
      puts "I'm running"
      define_method :enqueue, instance_method(:enqueue)
    end
  end
end

ActiveJob::Logging::EnqueueLogSubscriber.attach_to(:active_job)
```

I ran `rails console` and I can see:
```
ActiveJob::Logging::EnqueueLogSubscriber
I'm running
```
and when I run the job `MyJob.perform_later`, I can see in the logs:
```
Custom Active Job initializer is running
[ActiveJob] Enqueued MyJob (Job ID: 02a76977-c7a1-47c6-8366-787c097b83f4) to Sidekiq(default)
[ActiveJob] ↳ (irb):1:in `<main>'
[ActiveJob] Enqueued MyJob (Job ID: 02a76977-c7a1-47c6-8366-787c097b83f4)
```

Let me know what you think 😃

test/generators/suspenders/jobs_generator_test.rb
```diff
@@ -0,0 +1,140 @@
...
+        application_file = <<~RUBY
+          config.active_job.queue_adapter = :sidekiq
+          config.action_mailer.deliver_later_queue_name = nil
+          config.action_mailbox.queues.routing = nil
+          config.active_storage.queues.analysis = nil
+          config.active_storage.queues.purge = nil
+          config.active_storage.queues.mirror = nil
+        RUBY
+        File.write(config_path, application_file)
+
```
stevepolitodesign, line 74: I'm guessing you needed to do this because we're backing up the file in the `prepare_destination` step, correct? When we back up a file, we create a blank one. Until now, this hasn't been a problem. However, in order for the `environment` command to work, I think the corresponding file needs to have a `Rails::Application` declaration.

We can fix this by modifying the `backup` method.

```diff
diff --git a/test/test_helper.rb b/test/test_helper.rb
index 4fcaa7e..0495cd1 100644
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -77,8 +77,7 @@ module Suspenders::TestHelpers
   end
 
   def backup_file(file)
-    FileUtils.mv app_root(file), app_root("#{file}.bak")
-    touch file
+    FileUtils.copy app_root(file), app_root("#{file}.bak")
   end
 
   def restore_file(file)
``` 

This will alleviate us from having to populate the file in the setup step.

```suggestion

```

test/generators/suspenders/jobs_generator_test.rb
```diff
@@ -0,0 +1,140 @@
...
+          config.active_job.queue_adapter = :inline
+        RUBY
+        File.write(config_path, application_file)
+
+        run_generator
+
+        assert_file app_root("config/environments/test.rb") do |file|
+          assert_match(/config.active_job.queue_adapter = :inline/, file)
+        end
+      end
```
stevepolitodesign, line 99: Same here:

```suggestion
      test "adds ActiveJob configuration to the test environment file" do
        run_generator

        assert_file app_root("config/environments/test.rb") do |file|
          assert_match(/config.active_job.queue_adapter = :inline/, file)
        end
      end
```

test/generators/suspenders/jobs_generator_test.rb
```diff
@@ -0,0 +1,140 @@
...
+        assert_match(/bundle install/, output)
+      end
+
+      test "generator has a description" do
+        description = "Installs sidekiq for background job processing."
+
+        assert_equal description, Suspenders::Generators::JobsGenerator.desc
+      end
+
+      test "configures raising an error when ActiveJob is not found" do
```
stevepolitodesign, line 38: Does this test description make sense? I think we're configuring logging?

lib/generators/templates/active_job/active_job.rb
crackofdusk: Thank you for testing! I had a problem with this patch when upgrading an application from Rails 7.0 to 7.1. The error could have been caused by something else. For me it's enough if this generator works a new Rails application.

stevepolitodesign (commented):
We'll also want to update the `README` and `NEWS` too, before merging.

test/generators/suspenders/jobs_generator_test.rb
laicuRoot: Nice @stevepolitodesign thank you for this! Makes the test much cleaner.

test/generators/suspenders/jobs_generator_test.rb
laicuRoot: Yes, you are right, I forgot to modify the description. I will address this in the next commit 👍

test/generators/suspenders/jobs_generator_test.rb
```diff
@@ -0,0 +1,122 @@
...
+            end
+          end
+
+          ActiveJob::Logging::EnqueueLogSubscriber.attach_to(:active_job)
+        RUBY
+
+        run_generator
+
+        assert_file app_root("config/initializers/active_job.rb") do |file|
+          assert_match(expected_configuration, file)
```
stevepolitodesign, line 59: [Minor] Does this work?

```suggestion
          assert_equal(expected_configuration, file)
```

lib/generators/suspenders/jobs_generator.rb
```diff
@@ -0,0 +1,36 @@
+module Suspenders
+  module Generators
+    class JobsGenerator < Rails::Generators::Base
+      source_root File.expand_path("../../templates/active_job", __FILE__)
+      desc "Installs sidekiq for background job processing."
```
stevepolitodesign, line 5: [Minor] I think it's capitalized according to [their official website](https://sidekiq.org)

```suggestion
      desc "Installs Sidekiq for background job processing."
```

README.md
```diff
@@ -87,6 +87,15 @@ Also creates additional stylesheets if using PostCSS.
   [cssbundling-rails]: https://github.com/rails/cssbundling-rails
   [modern-normalize]: https://github.com/sindresorhus/modern-normalize
 
+
+### Jobs
+
+Installs sidekiq for background job processing and configures ActiveJob for job queueing.
```
stevepolitodesign, line 93: [Minor]

```suggestion
Installs [Sidekiq][] for background job processing and configures ActiveJob for job queueing.
```

lib/generators/templates/active_job/active_job.rb
stevepolitodesign, line 1: Here's the commit that introduced this for context.

https://github.com/thoughtbot/suspenders/commit/38b530c0a6a86175791c95a8f793d31b07b97c2d

stevepolitodesign (approved):
Great work. Thank you!

test/generators/suspenders/jobs_generator_test.rb
laicuRoot: Yes, it works 👍 I will update it.

lib/generators/suspenders/jobs_generator.rb
laicuRoot: Nice, thank you. I'm going to update all of them.

PR #1143, opened 2023-11-10
WIP Introduce `suspenders:lint` generator

dorianmarie:
- [x] Add standard gem
- [x] Add rake task
- [ ] Add standard to CI
- [ ] Test on new app

PR #1148, opened 2023-12-01
Introduce `suspenders:lint` generator

stevepolitodesign:
Closes #1107
Closes #1143

Creates a holistic linting solution that covers JavaScript, CSS, Ruby
and ERB.

Introduces `eslint` and `stylelint` NPM commands that leverage
[@thoughtbot/eslint-config][] and [@thoughtbot/stylelint-config][].

```
yarn run eslint
yarn run stylelint
```

Also introduces `.prettierrc` based off of our [Guides][].

[@thoughtbot/eslint-config]: https://github.com/thoughtbot/eslint-config
[@thoughtbot/stylelint-config]: https://github.com/thoughtbot/stylelint-config
[Guides]: https://github.com/thoughtbot/guides/blob/main/javascript/README.md#formatting

Introduces `rake standard` which also runs `erblint` to lint ERB files
via [better_html][], [erb_lint][] and [erblint-github][].

[better_html]: https://github.com/Shopify/better-html
[erb_lint]: https://github.com/Shopify/erb-lint
[erblint-github]: https://github.com/github/erblint-github

A future commit will ensure these linting rules are run during CI.

It should be noted that we deliberately permit this generator to be
invoked on API only applications, because those applications can still
contain views, like ones used for mailers.

Also improves the developer experience by introducing `with_test_suite`
helper, allowing the caller to run the generator in an application using
minitest or RSpec.

## To do

- [x] Run scripts that fix violations as part of the generator to ensure codebase is in a linted state.
- [x] Update `styles` generator by including a comment in the base styles to avoid `Unexpected empty source` linting error.

## Testing manually

1. Install the gem from this branch.

    ```ruby
    group :development, :test do
      gem "suspenders", github: "thoughtbot/suspenders", branch: "suspenders-3-0-0-stylelint"
    end
    ```

1. Run the generator.

    ```
    bin/rails g suspenders:styles # <- This ensures we're linting the correct files 
    bin/rails g suspenders:lint
    ```

1. Run the scripts and note the output.

    ```
    yarn run lint
    bundle exec rake standard
    ```

### DIFF
--- a/lib/generators/suspenders/lint_generator.rb
+++ b/lib/generators/suspenders/lint_generator.rb
@@ -1,0 +1,77 @@
+module Suspenders
+  module Generators
+    class LintGenerator < Rails::Generators::Base
+      include Suspenders::Generators::Helpers
+
+      source_root File.expand_path("../../templates/lint", __FILE__)
+      desc "Creates a holistic linting solution that covers JavaScript, CSS, Ruby and ERB."
+
+      def install_dependencies
+        run "yarn add stylelint@^15.10.1 eslint @thoughtbot/stylelint-config@3.0.0 @thoughtbot/eslint-config npm-run-all prettier --dev"
+      end
+
+      def install_gems
+        gem_group :development, :test do
+          gem "better_html", require: false
+          gem "erb_lint", require: false
+          gem "erblint-github", require: false
+          gem "standard"
+        end
+        Bundler.with_unbundled_env { run "bundle install" }
+      end
+
+      def configure_stylelint
+        copy_file "stylelintrc.json", ".stylelintrc.json"
+      end
+
+      def configure_eslint
+        copy_file "eslintrc.json", ".eslintrc.json"
+      end
+
+      def configure_prettier
+        copy_file "prettierrc", ".prettierrc"
+      end
+
+      def configure_erb_lint
+        copy_file "erb-lint.yml", ".erb-lint.yml"
+        copy_file "config_better_html.yml", "config/better_html.yml"
+        copy_file "config_initializers_better_html.rb", "config/initializers/better_html.rb"
+        copy_file "erblint.rake", "lib/tasks/erblint.rake"
+        template "rubocop.yml.tt", ".rubocop.yml"
+
+        if default_test_suite?
+          copy_file "better_html_test.rb", "test/views/better_html_test.rb"
+        elsif rspec_test_suite?
+          copy_file "better_html_spec.rb", "spec/views/better_html_spec.rb"
+        end
+      end
+
+      def update_package_json
+        content = File.read package_json
+        json = JSON.parse content
+        json["scripts"] ||= {}
+
+        json["scripts"]["lint"] = "run-p lint:eslint lint:stylelint lint:prettier"
+        json["scripts"]["lint:eslint"] = "eslint --max-warnings=0 --no-error-on-unmatched-pattern 'app/javascript/**/*.js'"
+        json["scripts"]["lint:stylelint"] = "stylelint 'app/assets/stylesheets/**/*.css'"
+        json["scripts"]["lint:prettier"] = "prettier --check '**/*' --ignore-unknown"
+        json["scripts"]["fix:prettier"] = "prettier --write '**/*' --ignore-unknown"
+
+        File.write package_json, JSON.pretty_generate(json)
+      end
+
+      # This needs to be the last method definition to ensure everything is
+      # properly configured
+      def fix_violations
+        run "yarn run fix:prettier"
+        run "bundle exec rake standard:fix_unsafely"
+      end
+
+      private
+
+      def package_json
+        Rails.root.join("package.json")
+      end
+    end
+  end
+end
--- a/lib/generators/templates/lint/erb-lint.yml
+++ b/lib/generators/templates/lint/erb-lint.yml
@@ -1,0 +1,63 @@
+---
+glob: "app/views/**/*.{html,turbo_stream}{+*,}.erb"
+
+linters:
+  AllowedScriptType:
+    enabled: true
+    allowed_types:
+      - "module"
+      - "text/javascript"
+  ErbSafety:
+    enabled: true
+    better_html_config: "config/better_html.yml"
+  GitHub::Accessibility::AvoidBothDisabledAndAriaDisabledCounter:
+    enabled: true
+  GitHub::Accessibility::AvoidGenericLinkTextCounter:
+    enabled: true
+  GitHub::Accessibility::DisabledAttributeCounter:
+    enabled: true
+  GitHub::Accessibility::IframeHasTitleCounter:
+    enabled: true
+  GitHub::Accessibility::ImageHasAltCounter:
+    enabled: true
+  GitHub::Accessibility::LandmarkHasLabelCounter:
+    enabled: true
+  GitHub::Accessibility::LinkHasHrefCounter:
+    enabled: true
+  GitHub::Accessibility::NestedInteractiveElementsCounter:
+    enabled: true
+  GitHub::Accessibility::NoAriaLabelMisuseCounter:
+    enabled: true
+  GitHub::Accessibility::NoPositiveTabIndexCounter:
+    enabled: true
+  GitHub::Accessibility::NoRedundantImageAltCounter:
+    enabled: true
+  GitHub::Accessibility::NoTitleAttributeCounter:
+    enabled: true
+  GitHub::Accessibility::SvgHasAccessibleTextCounter:
+    enabled: true
+  Rubocop:
+    enabled: true
+    rubocop_config:
+      inherit_from:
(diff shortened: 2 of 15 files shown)


lib/generators/suspenders/lint_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+          gem "erb_lint", require: false
+          gem "erblint-github", require: false
+          gem "standard"
+        end
+        Bundler.with_unbundled_env { run "bundle install" }
+      end
+
+      def configure_stylelint
+        copy_file "stylelintrc.json", ".stylelintrc.json"
+      end
```
stevepolitodesign, line 25: I wonder if we need this for API only applications? Mailers requires styles, but those are _usually_ inline.

lib/generators/templates/lint/eslint.json
stevepolitodesign, line 1: Delete this

lib/generators/suspenders/lint_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      def configure_eslint
+        copy_file "eslintrc.json", ".eslintrc.json"
+      end
```
stevepolitodesign, line 29: Would an API only ever have JavaSscript?

lib/generators/suspenders/lint_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+        end
+      end
+
+      # TODO: Consider extracting this into Rails
+      def update_package_json
+        content = File.read package_json
+        json = JSON.parse content
+        json["scripts"] ||= {}
+
+        json["scripts"]["eslint"] = "npx eslint 'app/javascript/**/*.js'"
```
stevehanson, line 54: This is awesome 🙌! I'm curious what your thoughts would be on aligning the script names and actions more closely with what is defined in the thoughtbot [ESLint-config blog post](https://thoughtbot.com/blog/eslint-configs-for-react-react-native#running-eslint)? That post matches how we often define these on JavaScript projects, including in [thoughtbelt](https://github.com/thoughtbot/thoughtbelt) (the React Native version of Suspenders). In particular:

* defining linter scripts with names `lint:{linter}` (eg. `lint:eslint`, `lint:stylelint`, `lint:prettier`)
* creating an additional `lint` script that runs _all_ linters, since we will most commonly want to run all linters.
* Adding a `lint:prettier` script, eg. `prettier --check '**/*' --ignore-unknown`
  * Do we want Prettier to run on JSON, CSS, and Markdown? It's my preference, but if not, we can restrict the glob a bit. The JSON formatting is definitely nice. I'm not sure offhand if we'll need CSS since we have stylelint. Markdown formatting is nice because it formats code in code blocks, but it can be a little annoying with tables, since it tries to line everything up.
* Running `eslint` with the `--max-warnings=0` flag. You can find more info about this in this [section of the blog post](https://thoughtbot.com/blog/eslint-configs-for-react-react-native#philosophy), but basically, we want warnings to also fail the build 
* No need for `npx` in the `eslint` and `stylelint` commands. Yarn or whatever package manager is used will find these since they're dev dependencies.
* I like to also have a `fix:prettier` command that runs Prettier and fixes any errors. CI would run `lint:prettier`. I rarely have to run `fix:prettier` in practice since my editor formats with Prettier on save, but it can be nice when we need to format files that were created by generators or for folks who don't have their editors configured with Prettier.

stevehanson (commented):
Amazing work! This will be such a valuable addition to Suspenders 🙌  I just did a quick review focused mostly on the JS linting setup. I left a few suggestions for how we name and execute the scripts. Let me know if you have any questions or wanted to pair on making those changes.

test/test_helper.rb
```diff
@@ -76,6 +82,21 @@ class Application < Rails::Application
     restore_file "config/application.rb"
   end
 
+  # TODO: Refactor existing tests to use this
+  def with_test_suite(test_suite = :minitest, &block)
```
mike-burns, line 86: I'll write an essay about this some day, but it does not make sense to use default args in private methods (in this case, package-private). You control each call site, so you can pass `:minitest`.

In this case, passing `:minitest` adds clarity at the call site.

lib/generators/suspenders/lint_generator.rb
```diff
@@ -0,0 +1,67 @@
+module Suspenders
+  module Generators
+    class LintGenerator < Rails::Generators::Base
+      include Suspenders::Generators::Helpers
+
+      source_root File.expand_path("../../templates/lint", __FILE__)
+      desc "Creates a holistic linting solution that covers JavaScript, CSS, Ruby and ERB."
+
+      def install_dependencies
+        run "yarn add stylelint eslint @thoughtbot/stylelint-config @thoughtbot/eslint-config --dev"
```
mike-burns, line 10: I got lost in the Yarn vs NPM battle. Do Rails projects default to Yarn now?

lib/generators/templates/lint/erb-lint.yml
```diff
@@ -0,0 +1,63 @@
...
+      Layout/TrailingEmptyLines:
+        Enabled: false
+      Layout/TrailingWhitespace:
+        Enabled: false
+      Layout/LeadingEmptyLines:
+        Enabled: false
+      Style/FrozenStringLiteralComment:
+        Enabled: false
+      Style/MultilineTernaryOperator:
+        Enabled: false
```
mike-burns, line 58: Why do we allow this?

lib/generators/templates/lint/erb-lint.yml
```diff
@@ -0,0 +1,63 @@
...
+      Lint/UselessAssignment:
```
mike-burns, line 59: Why do we allow this?

lib/generators/suspenders/lint_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+        copy_file "erblint.rake", "lib/tasks/erblint.rake"
+
+        if default_test_suite?
+          copy_file "better_html_test.rb", "test/views/better_html_test.rb"
+        elsif rspec_test_suite?
+          copy_file "better_html_spec.rb", "spec/views/better_html_spec.rb"
+        end
+      end
+
+      # TODO: Consider extracting this into Rails
```
stevepolitodesign, line 48: https://github.com/rails/rails/pull/50264

lib/generators/suspenders/lint_generator.rb
stevepolitodesign: Rails actually [defaults to Import Maps](https://guides.rubyonrails.org/working_with_javascript_in_rails.html#choosing-between-import-maps-and-a-javascript-bundler), but with the introduction of #1145 we've opted into use a JavaScript Runtime.

 However, it seems Rails still favors `yarn` over `npm` based on how they install packages in cssbundling-rails:

https://github.com/rails/cssbundling-rails/blob/66635235a96909f688627789b40cc81795917f7f/lib/install/helpers.rb#L4-L6

test/test_helper.rb
```diff
@@ -76,6 +82,21 @@ class Application < Rails::Application
...
+    if test_suite == :minitest
+      mkdir "test"
+    elsif test_suite == :rspec
+      mkdir "spec"
+      touch "spec/spec_helper.rb"
+    end
```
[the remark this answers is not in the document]
stevepolitodesign: Addressed here: e3bc6d0b07a86f1a411557331ff543c2814be5a6

test/test_helper.rb
stevepolitodesign: I want to read this essay!

a46f6dd05ff41f92440d712ac613312c19409ea3

stevepolitodesign:
@stevehanson after exhaustively testing this with a real application, I landed on this:

1. 520834a24875247c0971b2dca40aebee9f5e0a5a introduces your recommendations.
2. c8888320b35961cf402ba4402b9f7079541ad77d adds `--no-error-on-unmatched-pattern 'app/javascript/**/*.js'` since new Rails applications don't necessarily ship with any `.js` files. Without this, we raise an error when running `eslint`
3. 639056464b06a53b4018e1b1a5e1f268e1dccc1a fixes violations where the linter could not determine where the imported file exists.

    ```js
    // app/javascript/application.js
    // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
    import '@hotwired/turbo-rails';
    import 'controllers';
    ```

4. f3b85a4c4aabbaea40827535f90bbb9325f70e09 fixes a conflict between our stylelint config and prettier.
5. 5fe8cc4c7d2915b52c3cd4043a881163c81b8127 accounts for this [issue](https://github.com/thoughtbot/stylelint-config/issues/46).

lib/generators/templates/lint/erb-lint.yml
stevepolitodesign: I might be misunderstanding, but we are not allowing this since `Enabled ` is set to `false`.

PR #1151, opened 2023-12-12
Introduce `suspenders:production:environment` generator

crackofdusk:
Creates generator to configuration the production environment. For now,
this means [requiring a master key][master key].

Drops configuration for [asset_host][], since that should be an
infrastructure decision.

In an effort to distinguish between the development environment
configuration in #1149, we use the `production` namespace.

[master key]: https://guides.rubyonrails.org/configuring.html#config-require-master-key
[asset_host]: https://guides.rubyonrails.org/configuring.html#config-asset-host

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,7 @@ Unreleased
 * Introduce `suspenders:prerequisites` generator
 * Introduce `suspenders:ci` generator
 * Introduce `suspenders:cleanup:organize_gemfile` task
+* Introduce `suspenders:production:environment` generator
 
 20230113.0 (January, 13, 2023)
 
--- a/README.md
+++ b/README.md
@@ -197,6 +197,20 @@ Creates CI files for GitHub Actions.
 bin/rails g suspenders:ci
 ```
 
+### Environments
+
+#### Production
+
+Configures the production environment.
+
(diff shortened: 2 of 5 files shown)


stevepolitodesign:
Noting we might want to set:

```
config.require_master_key = true
```

test/generators/suspenders/production_environment_generator_test.rb
```diff
@@ -0,0 +1,35 @@
...
+            /^ +config.asset_host = ENV.fetch("ASSET_HOST", ENV.fetch("APPLICATION_HOST"))$/,
+            file
+          )
+        end
+      end
+
+      private
+
+      def prepare_destination
+        backup_file "config/environments/produciton.rb"
```
jho406, line 27:
```suggestion
        backup_file "config/environments/production.rb"
```

test/generators/suspenders/production_environment_generator_test.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def restore_destination
+        restore_file "config/environments/produciton.rb"
```
jho406, line 31:
```suggestion
        restore_file "config/environments/production.rb"
```

PR #1152, opened 2023-12-12
Introduce `suspenders:install:web` and application template

stevepolitodesign:
Create generator to invoke all necessary generators. We add it to the `install`
namespace to provide flexibility should we add other installation options, such
as ones for API Only applications.

Introduces [template][] as a means to invoke `suspenders:install:web` when
creating a new application. This serves as an alternative to the system
executable that was removed.

We call `db:prepare` in the template to ensure `bin/setup` works as expected.
This is because we overrode that file to use `dev:prime`, which assume the
database has been created.

[template]: https://guides.rubyonrails.org/rails_application_templates.html

## How to test this pull request

1. Generate a new Rails application:

    ```sh
    rails new suspenders_qa \
    --skip-test \
    -d=postgresql \
    -m=https://raw.githubusercontent.com/thoughtbot/suspenders/suspenders-3-0-0-web-generator/lib/install/web.rb
    ```

1. Run the setup script

    ```sh
    cd suspenders_qa bin/setup
    ```

1. Generate a scaffold to create data:

    ```sh
    bin/rails g scaffold post title
    bin/rails suspenders:db:migrate
    ```

1. Run rake and lint and note the output:

    ```sh
    bundle exec rake
    yarn run lint
    ```
    
**To Do**
- [x] Add `README` and other [templates](https://github.com/thoughtbot/suspenders/commit/8630fed66e248264721909e8eef50d1ccac2129f)
- [x] Ensure generated application has a passing CI in GitHub.
- [x] Ensure generated application can be deployed.
- [x] Consider creating a "clean-up" task that organizes the Gemfile.
- [x] [Enforce prerequisites](https://github.com/stevepolitodesign/rails_mvp/blob/main/template.rb)
- [x] Ensure setup script works.

### DIFF
--- a/lib/generators/suspenders/install/web_generator.rb
+++ b/lib/generators/suspenders/install/web_generator.rb
@@ -1,0 +1,67 @@
+module Suspenders
+  module Generators
+    module Install
+      class WebGenerator < Rails::Generators::Base
+        include Suspenders::Generators::APIAppUnsupported
+        include Suspenders::Generators::DatabaseUnsupported
+
+        desc <<~MARKDOWN
+          Invokes all necessary generators for new Rails applications generated with Suspenders.
+
+          This generatator is intended to be invoked as part of an [application template][].
+
+          ```
+          rails new suspenders_qa \
+          --skip-test \
+          -d=postgresql \
+          -m=https://raw.githubusercontent.com/thoughtbot/suspenders/main/lib/install/web.rb
+          ```
+
+          [application template]: https://guides.rubyonrails.org/rails_application_templates.html
+        MARKDOWN
+
+        def invoke_generators
+          # This needs to go first, since it configures `.node-version`
+          generate "suspenders:prerequisites"
+
+          generate "suspenders:accessibility"
+          generate "suspenders:advisories"
+          generate "suspenders:email"
+          generate "suspenders:factories"
+          generate "suspenders:inline_svg"
+          generate "suspenders:lint"
+          generate "suspenders:rake"
+          generate "suspenders:setup"
+          generate "suspenders:tasks"
+          generate "suspenders:testing"
+          generate "suspenders:views"
+
+          # suspenders:jobs needs to be invoked before suspenders:styles, since
+          # suspenders:styles generator creates Procfile.dev
+          generate "suspenders:styles"
+          generate "suspenders:jobs"
+
+          # Needs to run after other generators, since some touch the
+          # configuration files.
+          generate "suspenders:environments:test"
+          generate "suspenders:environments:development"
+          generate "suspenders:environments:production"
+
+          # Needs to be run last since it depends on lint, testing, and
+          # advisories
+          generate "suspenders:ci"
+        end
+
+        def cleanup
+          rake "suspenders:cleanup:organize_gemfile"
+          rake "suspenders:cleanup:generate_readme"
+        end
+
+        def lint
+          run "yarn run fix:prettier"
+          run "bundle exec rake standard:fix_unsafely"
+        end
+      end
+    end
+  end
+end
--- a/lib/install/web.rb
+++ b/lib/install/web.rb
@@ -1,0 +1,32 @@
+def apply_template!
+  if options[:database] == "postgresql" && options[:skip_test]
+    after_bundle do
+      gem_group :development, :test do
+        # TODO: Update once in `main`
+        gem "suspenders", github: "thoughtbot/suspenders", branch: "suspenders-3-0-0"
+      end
+
+      run "bundle install"
+
+      generate "suspenders:install:web"
+      rails_command "db:prepare"
+
+      say "\nCongratulations! You just pulled our suspenders."
+    end
+  else
+    message = <<~ERROR
+
+
+      === Please use the correct options ===
+
+      rails new <app_name> \\
+      --skip-test \\
+      -d=postgresql \\
+      -m=https://raw.githubusercontent.com/thoughtbot/suspenders/main/lib/install/web.rb
+    ERROR
+
+    fail Rails::Generators::Error, message
+  end
+end
+
+apply_template!
--- a/test/generators/suspenders/install/web_generator_test.rb
+++ b/test/generators/suspenders/install/web_generator_test.rb
@@ -1,0 +1,43 @@
+require "test_helper"
+require "generators/suspenders/install/web_generator"
+
+module Suspenders
+  module Generators
+    module Install
+      class WebGeneratorTest < Rails::Generators::TestCase
+        include Suspenders::TestHelpers
+
+        tests Suspenders::Generators::Install::WebGenerator
+        destination Rails.root
+        setup :prepare_destination
+        teardown :restore_destination
+
+        test "raises if API only application" do
+          within_api_only_app do
+            assert_raises Suspenders::Generators::APIAppUnsupported::Error do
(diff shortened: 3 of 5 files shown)


lib/generators/suspenders/install/web_generator.rb
```diff
@@ -0,0 +1,40 @@
...
+          Suspenders::Generators::StylesGenerator.new([], css: css).invoke_all
+          Suspenders::Generators::AdvisoriesGenerator.new.invoke_all
+          Suspenders::Generators::InlineSvgGenerator.new.invoke_all
+          Suspenders::Generators::FactoriesGenerator.new.invoke_all
+          Suspenders::Generators::JobsGenerator.new.invoke_all
+          Suspenders::Generators::RakeGenerator.new.invoke_all
+
+          # Needs to be invoked last, since it fixes any liting violations
+          # caused by the previous generators.
+          Suspenders::Generators::LintGenerator.new.invoke_all
```
stevepolitodesign, line 29: There's a case to be made that we could fix any violations in the source.

lib/generators/suspenders/styles_generator.rb
```diff
@@ -3,9 +3,7 @@ module Generators
     class StylesGenerator < Rails::Generators::Base
       include Suspenders::Generators::APIAppUnsupported
 
-      CSS_OPTIONS = %w[tailwind postcss].freeze
-
-      class_option :css, enum: CSS_OPTIONS, default: "postcss"
+      class_option :css, enum: Generators::CSS_OPTIONS, default: "postcss"
```
stevepolitodesign, line 6: Since we don't provide a default for the `WebGenerator`, I wonder if we should remove it here? Or, is not having a default a smell?

My intent with not having a default is to force the consumer to pick. Not all projects have the same requirements, so I want this to be flexable.

lib/install/web.rb
```diff
@@ -0,0 +1,13 @@
+css = ask("What CSS framwork do you want to use? [postcss, tailwind]")
+
+raise ArgumentError if ["postcss","tailwind"].exclude? css
+
+after_bundle do
+  gem_group :development, :test do
+    gem "suspenders", github: "thoughtbot/suspenders", branch: "suspenders-3-0-0-web-generator"
```
stevepolitodesign, line 7: We'll need to update this once we release a new version and merge into `main`.

lib/generators/suspenders/styles_generator.rb
stevepolitodesign, line 1: I should probably add a description.

test/generators/suspenders/install/web_generator_test.rb
```diff
@@ -0,0 +1,87 @@
...
+          styles_generator_mock.expects(:invoke_all).once
+          advisories_generator_mock.expects(:invoke_all).once
+          inline_svg_generator_mock.expects(:invoke_all).once
+          facories_generator_mock.expects(:invoke_all).once
+          jobs_generator_mock.expects(:invoke_all).once
+          lint_generator_mock.expects(:invoke_all).once
+          rake_generator_mock.expects(:invoke_all).once
+
+          generator_class.new([], css: "tailwind").invoke_generators
+        end
```
stevepolitodesign, line 58: I do everything in one test to avoid actually invoking the generators, since each one is stubbed. Otherwise, the `prepare_destination` and `restore_destination` logic would be too much, since it would need to account for all generators.

lib/install/web.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  run "bundle install"
+
+  generate "suspenders:install:web --css=#{css}"
```
stevepolitodesign, line 12: We might want to then `cd` into the directory and call `bin/setup` once that generator is created.

lib/install/web.rb
stevepolitodesign, line 1: Once #1156 is merged, we might want to consider raising if `--skip-test` was not passed.

lib/generators/suspenders/install/web_generator.rb
stevepolitodesign, line 1: This approach of invoking generators manually is helpful to driving out behavior via tests, but I think we'll want to update this to user the more familiar [generate](https://api.rubyonrails.org/classes/Rails/Generators/Actions.html#method-i-generate) syntax.

lib/generators/suspenders/install/web_generator.rb
stevepolitodesign: Maybe instead of worrying about the linting here, we do that in the template by manually running `yarn lint` and `standardrb—fix` as the very last step.

lib/install/web.rb
stevepolitodesign: Same with `-d=postgresql`

lib/suspenders/generators.rb
```diff
@@ -2,6 +2,9 @@
 
 module Suspenders
   module Generators
+
+    CSS_OPTIONS = %w[tailwind postcss].freeze
```
stevepolitodesign, line 6: This can be removed

lib/generators/suspenders/install/web_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+        def invoke_generators
+          # This needs to go first, since it configures `.node-version`
+          generate "suspenders:prerequisites"
+
+          generate "suspenders:accessibility"
+          generate "suspenders:advisories"
+          generate "suspenders:email"
+          generate "suspenders:factories"
+          generate "suspenders:inline_svg"
+          generate "suspenders:lint"
```
stevepolitodesign, line 32: Need to figure this out:

```
❯ yarn run lint
yarn run v1.22.17
$ run-p lint:eslint lint:stylelint lint:prettier
$ eslint --max-warnings=0 --no-error-on-unmatched-pattern 'app/javascript/**/*.js'
$ prettier --check '**/*' --ignore-unknown
$ stylelint 'app/assets/stylesheets/**/*.css'
Checking formatting...
.github/workflows/ci.yaml
Oops! Something went wrong! :(

ESLint: 9.0.0

Error: Could not find config file.
    at locateConfigFileToUse (/Users/polito/Desktop/suspenders_qa_0/node_modules/eslint/lib/eslint/eslint.js:349:21)
    at async calculateConfigArray (/Users/polito/Desktop/suspenders_qa_0/node_modules/eslint/lib/eslint/eslint.js:384:49)
    at async ESLint.lintFiles (/Users/polito/Desktop/suspenders_qa_0/node_modules/eslint/lib/eslint/eslint.js:814:25)
    at async Object.execute (/Users/polito/Desktop/suspenders_qa_0/node_modules/eslint/lib/cli.js:461:23)
    at async main (/Users/polito/Desktop/suspenders_qa_0/node_modules/eslint/bin/eslint.js:165:22)
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
app/assets/stylesheets/application.postcss.cssERROR: "lint:eslint" exited with 2.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

stevepolitodesign: Relates to https://github.com/eslint/eslint/issues/18287

stevepolitodesign: I think we can merge this commit without a solution for the above, since that can be addressed separately, and is not related to with commit.

PR #1153, opened 2023-12-12
Fix stylelint violations when using Tailwind

stevepolitodesign:
When we introduced #1148 we did not test it against applications that invoked
`suspenders:styles --css=tailwind`.


stevepolitodesign (commented):
Tested this manually and liting passed.

PR #1154, opened 2024-01-04
Introduce `suspenders:views` generator

stevepolitodesign:
Configures flash messages, page titles via the [title][] gem, and sets the
document [lang][].

[title]: https://github.com/calebhearth/title
[lang]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang

It should be noted that running `rails gscaffold` generates [views][] that
contain the flash. Although we could have [overridden][] this generator, that
would risk drift between our overridden generator and the one in Rails core.

Additionally, we decided to remove the `FlashesHelper` introduced in
[6c562b95e2e3c7ea31a04e6ef71c6a44b0b9ab2f][] since that is not a pattern we
currently use.

[views]: https://github.com/rails/rails/blob/main/railties/lib/rails/generators/erb/scaffold/templates/index.html.erb.tt
[overridden]: https://guides.rubyonrails.org/generators.html#overriding-rails-generator-templates
[6c562b95e2e3c7ea31a04e6ef71c6a44b0b9ab2f]: https://github.com/thoughtbot/suspenders/commit/6c562b95e2e3c7ea31a04e6ef71c6a44b0b9ab2f

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,7 @@ Unreleased
 * Introduce `suspenders:jobs` generator
 * Introduce `suspenders:lint` generator
 * Introduce `suspenders:rake` generator
+* Introduce `suspenders:views` generator
 
 20230113.0 (January, 13, 2023)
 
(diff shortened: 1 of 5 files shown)


stevepolitodesign (commented):
I wonder if we still want to encourage the use over CSS and JavaScript overrides?

https://github.com/thoughtbot/suspenders/blob/69ad8f0a1e3513fac4b06042349baabbfa224c25/templates/suspenders_layout.html.erb.erb#L18-L19

PR #1157, opened 2024-02-06
Introduce `suspenders:setup` generator

stevepolitodesign:
Replaces the [default setup script][] provided by Rails. The trade-off being
that the implementation is simple, but the cost is we risk drifting from Rails.
After attempting to use [gsub_file][] and [insert_into_file][], it felt simpler
to just override the file completely.

Clobber and precompile assets during setup if using [Sprockets][]. This
is helpful in cases where one might switch branches frequently, which
can result in broken assets.

Additionally, we [setup the test environment][] to avoid issues with asset
compilation in tests.

[default setup script]: https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/bin/setup.tt
[gsub_file]: https://rubydoc.info/gems/thor/Thor/Actions#gsub_file-instance_method
[insert_into_file]: https://rubydoc.info/gems/thor/Thor/Actions#insert_into_file-instance_method
[Sprockets]: https://github.com/rails/sprockets
[setup the test environment]: https://github.com/rails/rails/pull/47719/files

### DIFF
--- a/lib/generators/templates/setup/bin_setup.rb
+++ b/lib/generators/templates/setup/bin_setup.rb
@@ -41,11 +41,7 @@ FileUtils.chdir APP_ROOT do
   system("yarn install --check-files") if using_node?
 
   puts "\n== Preparing database and adding development seed data =="
-  if ENV["WIPE_DATABASE"] == true
-    system! "bin/rails db:reset dev:prime"
-  else
-    system! "bin/rails dev:prime"
-  end
+  system! "bin/rails dev:prime"
 
   # https://github.com/rails/rails/pull/47719/files
   puts "\n== Setup test environment =="
--- a/lib/generators/templates/setup/bin_setup.rb
+++ b/lib/generators/templates/setup/bin_setup.rb
@@ -20,7 +20,11 @@ FileUtils.chdir APP_ROOT do
   system("yarn install --check-files") if using_node?
 
   puts "\n== Preparing database and adding development seed data =="
-  system! "bin/rails dev:prime"
+  if File.exist? "lib/tasks/dev.rake"
+    system! "bin/rails dev:rake"
+  else
+    system! "bin/rails db:prepare"
+  end
 
   if Bundler.rubygems.find_name("sprockets")
     puts "\n== Generating assets =="
--- a/lib/generators/templates/setup/bin_setup.rb
+++ b/lib/generators/templates/setup/bin_setup.rb
@@ -1,6 +1,7 @@
 #!/usr/bin/env ruby
 require "fileutils"
 require "open3"
+require "bundler"
 
 # path to your application root.
 APP_ROOT = File.expand_path("..", __dir__)
@@ -42,6 +43,11 @@ FileUtils.chdir APP_ROOT do
 
   puts "\n== Preparing database and adding development seed data =="
   system! "bin/rails dev:prime"
+
+  if Bundler.rubygems.find_name("sprockets")
+    puts "\n== Generating assets =="
+    system! "bin/rails assets:clobber assets:precompile"
+  end
 
   # https://github.com/rails/rails/pull/47719/files
   puts "\n== Setup test environment =="
--- a/lib/generators/templates/setup/bin_setup.rb
+++ b/lib/generators/templates/setup/bin_setup.rb
@@ -1,0 +1,59 @@
+#!/usr/bin/env ruby
+require "fileutils"
+require "open3"
+
+# path to your application root.
+APP_ROOT = File.expand_path("..", __dir__)
+
+def system!(*args)
+  system(*args, exception: true)
+end
+
+def using_node?
+  File.exist? "package.json"
+end
+
+FileUtils.chdir APP_ROOT do
+  if ENV["CI"].nil?
+    puts "\n== Installing Homebrew Bundle from Brewfile =="
+    system! "brew bundle"
+
+    puts "\n== Starting postgresql@15 =="
+    system! "brew services restart postgresql@15"
+
+    puts "\n== Starting redis =="
+    system! "brew services restart redis"
+
+    _, _, status = Open3.capture3("which asdf")
+    if status.success?
+      puts "\n== Installing tool dependecies via asdf =="
+      system! "asdf plugin-add ruby https://github.com/asdf-vm/asdf-ruby"
+      system! "asdf plugin-add nodejs https://github.com/asdf-vm/asdf-nodejs" if using_node?
+      system! "asdf plugin-add yarn https://github.com/twuni/asdf-yarn" if using_node?
+      system! "asdf plugin-update --all"
+      system! "asdf install"
+    end
+  end
+
+  puts "\n== Installing dependencies =="
+  system! "gem install bundler --conservative"
+  system("bundle check") || system!("bundle install")
+  system("yarn install --check-files") if using_node?
+
+  puts "\n== Preparing database and adding development seed data =="
+  if ENV["WIPE_DATABASE"] == true
+    system! "bin/rails db:reset dev:prime"
+  else
+    system! "bin/rails dev:prime"
+  end
+
+  # https://github.com/rails/rails/pull/47719/files
+  puts "\n== Setup test environment =="
+  system! "bin/rails test:prepare"
+
+  puts "\n== Removing old logs and tempfiles =="
+  system! "bin/rails log:clear tmp:clear"
+
+  puts "\n== Restarting application server =="
+  system! "bin/rails restart"
+end
--- a/lib/generators/templates/setup/bin_setup.rb
+++ b/lib/generators/templates/setup/bin_setup.rb
@@ -1,6 +1,5 @@
 #!/usr/bin/env ruby
 require "fileutils"
-require "open3"
 require "bundler"
 
 # path to your application root.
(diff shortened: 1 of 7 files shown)


lib/generators/templates/setup/bin_setup.rb
```diff
@@ -0,0 +1,56 @@
...
+    puts "\n== Installing Homebrew Bundle from Brewfile =="
+    system! "brew bundle"
+
+    puts "\n== Starting postgresql@15 =="
+    system! "brew services restart postgresql@15"
+
+    puts "\n== Starting redis =="
+    system! "brew services restart redis"
+
+    _, _, status = Open3.capture3("which asdf")
```
stevepolitodesign, line 23: There are several ways to do this, but I landed on this. I like how explicit it is, but curious what others think.

lib/generators/templates/setup/bin_setup.rb
```diff
@@ -0,0 +1,56 @@
...
+    if status.success?
+      puts "== Installing tool dependecies via asdf =="
+      system! "asdf plugin-add nodejs https://github.com/asdf-vm/asdf-nodejs"
+      system! "asdf plugin-add ruby https://github.com/asdf-vm/asdf-ruby"
+      system! "asdf plugin-add yarn https://github.com/twuni/asdf-yarn"
```
stevepolitodesign, line 28: There's a case to be made we should conditionally install `nodejs` and `yarn`, since we can't guarantee when this generator will be run. 
```suggestion
      system! "asdf plugin-add nodejs https://github.com/asdf-vm/asdf-nodejs" if File.exist?("package.json")
      system! "asdf plugin-add ruby https://github.com/asdf-vm/asdf-ruby"
      system! "asdf plugin-add yarn https://github.com/twuni/asdf-yarn" if File.exist?("package.json")
```

lib/generators/templates/setup/bin_setup.rb
```diff
@@ -0,0 +1,56 @@
...
+      system! "asdf plugin-update --all"
+      system! "asdf install"
+    end
+  end
+
+  puts "== Installing dependencies =="
+  system! "gem install bundler --conservative"
+  system("bundle check") || system!("bundle install")
+  system("yarn install --check-files") if File.exist?("package.json")
```
stevepolitodesign, line 37: It's possible [this line will already exist](https://github.com/rails/rails/blob/f14401aa17c02cfad2213b5bb92a650f43ee4b86/railties/lib/rails/generators/rails/app/templates/bin/setup.tt#L21), so we might want to check for it first?

lib/generators/templates/setup/brewfile
```diff
@@ -0,0 +1,9 @@
+# https://formulae.brew.sh/formula/vips
+brew "vips"
+
+# https://formulae.brew.sh/formula/redis
+brew "redis"
```
stevepolitodesign, line 5: I wonder if these should be conditionally loaded based on the Gemfile?

lib/generators/templates/setup/brewfile
```diff
@@ -0,0 +1,9 @@
...
+
+# https://formulae.brew.sh/formula/postgresql@15
+# Change version to match production
+brew "postgresql@15"
```
stevepolitodesign, line 9: Should we raise early if we're not using Postgres? I ask because we've decided as an organization we prefer PostgreSQL.

https://github.com/thoughtbot/dotfiles/pull/728

lib/generators/templates/setup/bin_setup.rb
stevepolitodesign: Actually, that's not necessary since we're overriding the file and have a conditional.

lib/generators/templates/setup/brewfile
stevepolitodesign: After giving it some thought, I think it's acceptable to leave these all in. We're focusing on new applications, and most new applications will need image processing and Redis.

lib/generators/templates/setup/dev_prime.rb
```diff
@@ -0,0 +1,12 @@
+if Rails.env.development? || Rails.env.test?
+  require "factory_bot" if Bundler.rubygems.find_name("factory_bot_rails").any?
+
+  namespace :dev do
+    desc "Sample data for local development environment"
+    task prime: "db:setup" do
```
stevepolitodesign, line 6: This seemed to be wiping the database, even when `WIPE_DATABASE` was not set.

lib/generators/templates/setup/dev_prime.rb
```diff
@@ -0,0 +1,12 @@
+if Rails.env.development? || Rails.env.test?
+  require "factory_bot" if Bundler.rubygems.find_name("factory_bot_rails").any?
```
stevepolitodesign, line 2: I wonder if instead we conditionally should run `suspenders:factories` if the gem is missing?

lib/generators/templates/setup/dev_prime.rb
stevepolitodesign: It seems `db:setup` is wipes everything. d28306358f097b7b73f4cff56bd6a5e616f3dcdb

lib/generators/templates/setup/bin_setup.rb
```diff
@@ -0,0 +1,61 @@
...
+def system!(*args)
+  system(*args, exception: true)
+end
+
+def using_node?
+  File.exist? "package.json"
+end
+
+FileUtils.chdir APP_ROOT do
+  if ENV["CI"].nil?
```
mike-burns, line 18: Is the idea that those of us in #not-macos need to identify as CI?

lib/generators/templates/setup/bin_setup.rb
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env ruby
```
mike-burns, line 1: I admit, the idea of writing this shell script in Ruby never made sense to me.

lib/generators/templates/setup/dev_prime.rb
mike-burns: Can `suspenders:factories` add this file instead?

mike-burns (commented):
This feels like it conflates system setup with app setup. I'm not sold on mixing the two.

(And I say that as someone who doesn't use Homebrew.)

lib/generators/templates/setup/bin_setup.rb
stevepolitodesign: I might be misunderstanding, but this is to prevent this block of code from running during GitHub actions, since `CI` is a [default environment variable](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables).

Is the concern that some operating systems ship with a `CI` value set?

lib/generators/templates/setup/dev_prime.rb
stevepolitodesign: I was toying with that idea when I made the `suspenders:factories` generator, but then the conditional logic would need to happen in the setup script.

```ruby
if File.exist? "lib/tasks/dev.rake"
  system! "bin/rails dev:prime"
else
    system! "bin/rails dev:setup"
end
```

I guess that's simpler, though.

lib/generators/templates/setup/bin_setup.rb
mike-burns: Those of us on Linux don't use Homebrew.

lib/generators/templates/setup/dev_prime.rb
mike-burns: Or the `suspenders:factory` generator could `gsub("rails dev:setup", "rails dev:prime")`.

lib/generators/templates/setup/bin_setup.rb
stevepolitodesign: Ahh, thank you for clarifying. It sounds like we should check to see if Homebrew is also installed too.

stevepolitodesign:
@mike-burns 

> This feels like it conflates system setup with app setup. I'm not sold on mixing the two.

Is the concern that this setup script seems to favor MacOS? Or, even if it were more agnostic, would you rather see the Hombrew and asdf sections removed?

Should those be the responsibility of something like [Laptop](https://github.com/thoughtbot/laptop)

mike-burns:
> Should those be the responsibility of something like Laptop

The system setup has traditionally been the responsibility of the Laptop script, yeah. Not sure why we would duplicate it here.

lib/generators/templates/setup/bin_setup.rb
stevepolitodesign: No longer relevant: 97068cd

lib/generators/templates/setup/dev_prime.rb
stevepolitodesign: d283063

lib/generators/templates/setup/dev_prime.rb
stevepolitodesign: I ended up opting for the conditional, since I don't like the idea of any other another generator affecting the setup script.

I'll need to either modify `suspenders:factory` to create the `dev:prime` task, or create a separate generator for this.

b074e46

lib/generators/templates/setup/bin_setup.rb
stevepolitodesign: Let's keep this application specific for now.

97068cd

lib/generators/templates/setup/bin_setup.rb
```diff
@@ -0,0 +1,43 @@
...
+
+FileUtils.chdir APP_ROOT do
+  puts "== Installing dependencies =="
+  system! "gem install bundler --conservative"
+  system("bundle check") || system!("bundle install")
+  system("yarn install --check-files") if using_node?
+
+  puts "\n== Preparing database and adding development seed data =="
+  if File.exist? "lib/tasks/dev.rake"
+    system! "bin/rails dev:rake"
```
mike-burns, line 24: `dev:rake`?

stevepolitodesign: Whoops!

c010e4a8876d6ea598c4570e844fabcb37a9014a

mike-burns (approved):
Nice!

PR #1162, opened 2024-02-27
Introduce `suspenders:email` generator

stevepolitodesign:
Creates [email interceptor][] to ensure emails are sent to preconfigured
addresses in non-production environments. Drops support for [Recipient
Interceptor][], since this functionality is supported in Rails Core.

In order to enable the interceptor, set `EMAIL_RECIPIENTS` to a
comma-separated list.

```sh
INTERCEPTOR_ADDRESSES="user_1@example.com,user_2@example.com" bin/rails s
```

Configures [default_url_options][] in `test` and `development`
environments so that `*_url` helpers will work.

Note that we do not configure `production`, since we've determined the
`delivery_method` is something each project will configure
independently. This means we'll just use the [default configuration][],
which is `smtp`.

[email interceptor]: https://guides.rubyonrails.org/action_mailer_basics.html#intercepting-emails
[Recipient Interceptor]: https://github.com/croaky/recipient_interceptor
[default_url_options]: https://guides.rubyonrails.org/action_mailer_basics.html#generating-urls-in-action-mailer-views
[default configuration]: https://guides.rubyonrails.org/configuring.html#config-action-mailer-delivery-method

### DIFF
--- a/lib/generators/suspenders/email_generator.rb
+++ b/lib/generators/suspenders/email_generator.rb
@@ -1,0 +1,52 @@
+module Suspenders
+  module Generators
+    class EmailGenerator < Rails::Generators::Base
+      source_root File.expand_path("../../templates/email", __FILE__)
+      desc <<~TEXT
+        Intercepts emails in non-production environments by setting `INTERCEPTOR_ADDRESSES`.
+
+        ```sh
+        INTERCEPTOR_ADDRESSES="user_1@example.com,user_2@example.com" bin/rails s
+        ```
+
+        Configures `default_url_options` in `test` and `development`.
+      TEXT
+
+      def create_email_interceptor
+        copy_file "email_interceptor.rb", "app/mailers/email_interceptor.rb"
+      end
+
+      def create_email_interceptor_initializer
+        initializer "email_interceptor.rb", <<~RUBY
+          Rails.application.configure do
+            if ENV["INTERCEPTOR_ADDRESSES"].present?
+              config.action_mailer.interceptors = %w[EmailInterceptor]
+            end
+          end
+        RUBY
+      end
+
+      def configure_email_interceptor
+        environment do
+          %(
+            config.to_prepare do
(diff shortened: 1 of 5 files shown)


lib/generators/suspenders/email_generator.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+      end
+
+      def configure_development_environment
+        environment %(config.action_mailer.default_url_options = { host: "localhost", port: 3000 }), env: "development"
+      end
+
+      def configure_test_environment
+        environment %(config.action_mailer.default_url_options = { host: "www.example.com" }), env: "test"
+      end
```
stevepolitodesign, line 41: These might be able to be removed if https://github.com/rails/rails/pull/51191 is merged.

stevepolitodesign, line 41: Also noting I didn't need to [config.asset_host](https://guides.rubyonrails.org/configuring.html#config-asset-host). Everything worked without issue in `test` and `development`.

lib/generators/templates/email/email_interceptor.rb
```diff
@@ -0,0 +1,11 @@
+class EmailInterceptor
+  include ActiveSupport::Configurable
+
+  config_accessor :interceptor_addresses, default: []
+
+  def self.delivering_email(message)
+    to = interceptor_addresses
```
stevepolitodesign, line 7: Need to investigate if I should send an email per address in this list.

stevepolitodesign: Using an array works: https://guides.rubyonrails.org/action_mailer_basics.html#sending-email-to-multiple-recipients

lib/generators/suspenders/email_generator.rb
stevepolitodesign: We can remove once there's a new release.

mike-burns (approved):

PR #1164, opened 2024-03-07
Disable Turbo's InstantClick

stevepolitodesign:
Turbo 8 introduced [InstantClick][]. An unintended side-effect of this
feature is that page requests are [not debounced][] when a user hovers
over a link. This is a concern for two reasons:

1. It creates an unnecessary burden on the server, which could affect
   rate limiting.
2. It can artificially inflate our metrics.

[InstantClick]: https://turbo.hotwired.dev/handbook/drive#prefetching-links-on-hover
[not debounced]:https://github.com/hotwired/turbo/pull/1181#issuecomment-1936505362

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,10 +129,13 @@ Configures the default Rake task to audit and lint the codebase with
 ### Views
 
 Configures flash messages, page titles via the [title][] gem, and sets the
-document [lang][].
+document [lang][]. Disables Turbo's [InstantClick][] to account for page
+requests that are [not debounced][].
 
 [title]: https://github.com/calebhearth/title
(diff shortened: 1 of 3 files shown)


davidalejandroaguilar:
Just want to mention that it is debounced in some way:

> To avoid prefetching links that the user is briefly hovering, Turbo waits 100ms after the user hovers over the link before prefetching it.

I agree you might want to have more control and be able to configure the delay, which is what [I originally intended](https://github.com/hotwired/turbo/pull/1101#issuecomment-1899319337). However, [the subsequent PR](https://github.com/hotwired/turbo/pull/1164) to add this received push back until this kind of customization was in high demand.

So, if you think being able to tweak the delay would be useful for you, feel free to chime in that PR.

PR #1168, opened 2024-03-21
Testing Generator: Improvements

stevepolitodesign:
Follow-up to #1156

Creates parity with Rails' decision to [use a headless driver by default][headless].

This will be fixed in an [upcoming release][rspec] of rspec-rails, but I felt it was important to capture here. Additionally, it ensures the `screen_size` is the same as what is set in Rails.

Removes `webdrivers` dependency in favor of `selenium-webdriver`. This generator assumes the app was generated with the `--skip-test` flag, which means we need to add the `selenium-webdriver` and `capybara` gems.

Updates `action_dispatch-testing-integration-capybara` dependency to the most recent tagged release in an effort to suppress Dependabot notifications.

Ensure all files under `spec/support` are loaded by uncommenting a line generated by the RSpec installation script.

[headless]: https://github.com/rails/rails/pull/50512
[rspec]: https://github.com/rspec/rspec-rails/pull/2746


README.md
```diff
@@ -178,17 +178,10 @@ bin/rails g suspenders:email
 
 ### Testing
 
-Set up the project for an in-depth test-driven development workflow.
-
-Installs and configures [rspec-rails][],
-[action_dispatch-testing-integration-capybara][], [shoulda-matchers][],
-[webdrivers][] and [webmock][].
+Set up the project for an in-depth test-driven development workflow via
+[rspec-rails][] and friends.
```
stevepolitodesign, line 182: I felt it was easier to keep this simple.

PR #1172, opened 2024-03-22
Introduce `suspenders:ci` generator

stevepolitodesign:
Creates CI template to be run via [GitHub Actions][ga] based on a
[similar template][ci template] that will be generated in an upcoming
Release of Rails.

Also create a [Dependabot][dependabot] file based off the [the upcoming
release][ci template].

Raises if the application is not using PostgreSQL, since our CI template
assumes that adapter.

Because this generator can be run in an existing application, we add
conditional checks for some jobs. However, this generator is intended
to be run as part of our holistic `suspenders:install:web` which will be
introduced in #1152.

Once Rails is released to contain a CI template, we will need to
consider how we want to handle conflicts between its file and ours, but
for now, we do not need to worry about that.

[ga]: https://docs.github.com/en/actions
[ci template]: https://github.com/rails/rails/pull/50508
[dependabot]: https://docs.github.com/en/code-security/dependabot/working-with-dependabot

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,14 @@ Configures prerequisites. Currently Node.
 bin/rails g suspenders:prerequisites
 ```
 
+### CI
+
+Creates CI files for GitHub Actions.
+
+```
+bin/rails g suspenders:ci
+```
+
 ## Contributing
 
 See the [CONTRIBUTING] document.
--- a/lib/generators/suspenders/ci_generator.rb
+++ b/lib/generators/suspenders/ci_generator.rb
@@ -1,0 +1,52 @@
+module Suspenders
+  module Generators
+    class CiGenerator < Rails::Generators::Base
+      include Suspenders::Generators::DatabaseUnsupported
+      include Suspenders::Generators::Helpers
+
+      source_root File.expand_path("../../templates/ci", __FILE__)
+      desc "Creates CI files for GitHub Actions"
+
+      def ci_files
+        empty_directory ".github/workflows"
+        template "ci.yml", ".github/workflows/ci.yaml"
+        template "dependabot.yml", ".github/dependabot.yaml"
+      end
+
+      private
+
+      def scan_ruby?
+        has_gem? "bundler-audit"
+      end
+
+      def scan_js?
+        File.exist?("bin/importmap") && using_node?
+      end
+
+      def lint?
+        using_node? && has_gem?("standard") && has_yarn_script?("lint")
+      end
+
+      def using_node?
+        File.exist? "package.json"
+      end
+
+      def has_gem?(name)
+        Bundler.rubygems.find_name(name).any?
+      end
+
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -188,6 +188,14 @@ Configures prerequisites. Currently Node.
 bin/rails g suspenders:prerequisites
 ```
 
+### CI
+
+CI
```
stevepolitodesign, line 193: Add description.

lib/generators/suspenders/ci_generator.rb
```diff
@@ -0,0 +1,12 @@
+module Suspenders
+  module Generators
+    class CiGenerator < Rails::Generators::Base
+      source_root File.expand_path("../../templates/ci", __FILE__)
+
+      def cifiles
+        empty_directory ".github/workflows"
+        template "ci.yml", ".github/workflows/ci.yml"
```
stevepolitodesign, line 8: I wonder if we should add a [dependabot](https://github.com/rails/rails/pull/50508/files#diff-1182a749a11dc300cbc5af5e664598ab031c4b152cd2828ecae02d2d4f09168d) script too?

lib/generators/templates/ci/ci.yml.tt
```diff
@@ -0,0 +1,146 @@
...
+      - name: Install modules
+        run: yarn install
+
+      - name: Scan for security vulnerabilities in JavaScript dependencies
+        run: |
+          bin/importmap audit
+          yarn audit
+<% end -%>
+
+<%- if Bundler.rubygems.find_name("standard").any? && File.exist?(".prettierrc") -%>
```
stevepolitodesign, line 57: Checking for `.prettierrc` doesn't guarantee that  `yarn lint` exists. I wonder if we want to parse `package.json`?

lib/generators/templates/ci/ci.yml.tt
```diff
@@ -0,0 +1,146 @@
...
+          ruby-version: .ruby-version
+          bundler-cache: true
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .node-version
+
+      - name: Install modules
+        run: yarn install
```
stevepolitodesign, line 123: This should be wrapped in a conditional.

lib/suspenders/generators.rb
```diff
@@ -44,5 +44,33 @@ def api_only_app?
...
+
+        def database_unsupported?
+          configuration = File.read(Rails.root.join("config/database.yml"))
+          configuration = YAML.load(configuration, aliases: true)
+          adapter = configuration["default"]["adapter"]
+
+          adapter != "postgresql"
+        end
+      end
+    end
```
stevepolitodesign, line 74: I might want to add a unit test for this Class.

PR #1181, opened 2024-03-30
Introduce `suspenders:cleanup:organize_gemfile` task

stevepolitodesign:
Introduce `Suspenders::Cleanup::OrganizeGemfile` class and corresponding task in an effort to reduce duplicate groups in the modified Gemfile.

This is because the [gem_group][] method does not modify existing groups. I've opened [#49512][] in an effort to fix this, but until then, this task will suffice.

This class is designed to be run after `suspenders:install:web`, and does not account for all edge cases. For example, it assumes gems are grouped by symbols (i.e. :test and not "test"), and does not account for inline syntax:

```ruby
gem 'my-gem', group: [:cucumber, :test]
```

We could consider extracting this into a Gem (with a fun name, of course. Maybe "Polish"), but for now, I think this simple procedural code if just fine.

Additionally, this commit removes duplicate Rake task that was generated with the plugin.

[gem_group]: https://api.rubyonrails.org/classes/Rails/Generators/Actions.html#method-i-gem_group
[#49512]: https://github.com/rails/rails/pull/49512

### DIFF
--- a/lib/suspenders/cleanup/organize_gemfile.rb
+++ b/lib/suspenders/cleanup/organize_gemfile.rb
@@ -1,0 +1,134 @@
+module Suspenders
+  module Cleanup
+    class OrganizeGemfile
+      def self.perform(gemfile)
+        new(gemfile).perform
+      end
+
+      attr_reader :gemfile, :current_lines, :new_lines, :new_line_markers,
+        :current_group, :gem_groups
+
+      def initialize(gemfile)
+        @gemfile = gemfile
+
+        @current_lines = File.read(gemfile).lines
+        @new_lines = []
+        @new_line_markers = []
+
+        @current_group = nil
+        @gem_groups = {}
+      end
+
+      def perform
+        remove_line_breaks
+        sort_gems_and_groups
+        add_gem_groups_to_gemfile
+        add_line_breaks
+        cleanup
+
+        File.open(gemfile, "w+") { _1.write new_lines.join }
+      end
+
+      private
+
+      def remove_line_breaks
+        current_lines.delete("\n")
+      end
+
+      def sort_gems_and_groups
+        current_lines.each do |line|
+          if line.starts_with?(/group/)
+            @current_group = line
+          end
+
+          # Consolidate gem groups
+          if current_group
+            if line.starts_with?(/end/)
+              @current_group = nil
+            elsif !line.starts_with?(/group/)
+              gem_groups[current_group] ||= []
+              gem_groups[current_group] << line
+            end
+          # Add non-grouped gems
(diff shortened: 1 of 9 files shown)


test/suspenders/cleanup/organize_gemfile_test.rb
```diff
@@ -0,0 +1,21 @@
...
+        modified = file_fixture("gemfile_clean").read
+        File.open("tmp/Gemfile", "w+") { _1.write original }
+
+        Suspenders::Cleanup::OrganizeGemfile.perform("tmp/Gemfile")
+
+        gemfile = File.read("tmp/Gemfile")
+        assert_equal modified, gemfile
+      ensure
+        File.delete("tmp/Gemfile")
+      end
```
stevepolitodesign, line 18: I originally tried using `TempFile` for this, but found it challenging, since the file was closed when I tried to make assertions.

test/fixtures/files/gemfile_messy
stevepolitodesign, line 1: This is exactly how `suspenders:install:web` modifies the Gemfile.

lib/suspenders/cleanup/organize_gemfile.rb
```diff
@@ -0,0 +1,119 @@
...
+        @current_lines = File.read(gemfile).lines
+        @new_lines = []
+        @new_lines_markers = []
+
+        @current_group = nil
+        @gem_groups = {}
+      end
+
+      def perform
+        # Remove line breaks to ensure consistent spacing
```
mike-burns, line 23: What do you think of something like:

```ruby
def perform
  group_gems
  add_consolidated_gem_groups
  add_relevant_newlines
  final_cleanup
end
```

This way we get a high-level view of what it means to organize a Gemfile.

lib/suspenders/cleanup/organize_gemfile.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      attr_reader :gemfile, :current_lines, :new_lines, :new_lines_markers,
+        :current_group, :gem_groups
+
+      def initialize(gemfile)
+        @gemfile = gemfile
+
+        @current_lines = File.read(gemfile).lines
+        @new_lines = []
+        @new_lines_markers = []
```
stevepolitodesign, line 16: Should probably be:

```suggestion
        @new_line_markers = []
```

lib/suspenders/cleanup/organize_gemfile.rb
stevepolitodesign: Addressed here: fd6db90f91ff08fe849528c35482afbc0f71c5f1

PR #1191, opened 2024-04-19
Styles Generator: Install `postcss-url`

### LINKED ISSUE #1190: PostCSS fonts
When including styles from node_modules, sometimes the font faces refer to the packages font files. When PostCSS process that, it doesn't copy the font files over.
For example with primereact installed at `node_modules/primereact/resources/themes/lara-dark-amber/theme.css`

```
@font-face {
  font-family: "Inter var";
  font-weight: 100 900;
  font-display: swap;
  font-style: normal;
  font-named-instance: "Regular";
  src: url("./fonts/Inter-roman.var.woff2?v=3.19") format("woff2");
}

```

Suspenders didn't add additional postcss modules, but this is what I had to do to get that working:

```
module.exports = {
  plugins: [
    require('postcss-import'),
    require('postcss-nesting'),
    require('autoprefixer'),
    require('postcss-url')({
      url: 'copy',
      assetsPath: 'static'
    })
  ],
}
```

stevepolitodesign:
Closes #1190

When including styles from `node_modules` via `@import`, we discovered that static assets referenced in those style sheets are not loaded, resulting in console errors.

Since [cssbundling-rails][] ships with [postcss-import][], we are already encouraging the use of importing styles from `node_modules`.

This commit aims to solve this problem by installing and configuring [postcss-url][]. Because `postcss-url` needs a directory to copy assets to, we create `app/assets/static` as part of the generator.

One thing to note is that we override the `postcss.config.js` generated by the cssbundling-rails installation script, which assumes that the following plugins have been installed.

- `postcss-import`
- `postcss-nesting`
- `autoprefixer`

Should that change, this file would be invalid, and those packages would need to be installed.

[cssbundling-rails]: https://github.com/rails/cssbundling-rails
[postcss-import]: https://github.com/postcss/postcss-import
[postcss-url]: https://github.com/postcss/postcss-url

### DIFF
--- a/lib/generators/suspenders/styles_generator.rb
+++ b/lib/generators/suspenders/styles_generator.rb
@@ -3,6 +3,7 @@ module Suspenders
     class StylesGenerator < Rails::Generators::Base
       include Suspenders::Generators::APIAppUnsupported
 
+      source_root File.expand_path("../../templates/styles", __FILE__)
       desc <<~MARKDOWN
         Configures application to use PostCSS via cssbundling-rails.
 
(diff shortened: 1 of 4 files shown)


lib/generators/suspenders/styles_generator.rb
```diff
@@ -39,6 +40,25 @@ def configure_application_stylesheet
           TEXT
         end
       end
+
+      def install_postcss_url
+        run "yarn add postcss-url"
+      end
+
+      def configures_postcss
+        File.delete(postcss_config) if File.exist?(postcss_config)
```
stevepolitodesign, line 49: Unfortunately, `copy_file` does not override existing files. So, we need to delete it first.

PR #1165, opened 2024-03-14
Call dynamic-readme reusable workflow

stefannibrasil:
We want to have a way to edit our README footer at one place and have the changes from there be propagated to our repos.

By adding this snippet in the README, we call this reusable workflow: https://github.com/thoughtbot/templates/blob/main/.github/workflows/dynamic-readme.yaml that renders and updates the README footer dynamically.

PR #1200, opened 2024-04-29
Update License

Tresor11:
Update the License date to the current year.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ Thank you, [contributors]!
 
 ## License
 
-Suspenders is Copyright © 2008-2017 thoughtbot.
+Suspenders is Copyright © 2008-2024 thoughtbot.
 It is free software,
 and may be redistributed under the terms specified in the [LICENSE] file.
 


stevepolitodesign:
Thank you for opening this. In #1165 we explored using our new [dynamic README](https://github.com/thoughtbot/templates/blob/main/.github/workflows/dynamic-readme.yaml), but ended up closing that pull request because of #1135.

Now that #1135  is for review, I wonder if instead we should simply [remove the date altogether](https://github.com/thoughtbot/suspenders/pull/1165/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R203) and base off of the `suspenders-3-0-0` branch?

That will make re-opening #1165 simpler when #1135 is merged into `main`.

Tresor11:
Thanks for the review @stevepolitodesign 
I think it makes sense to remove the date altogether so we don't have to remember to update it each year😄

I'll close this PR

PR #1202, opened 2024-05-08
Prerequisites: Raise if Node version unsupported

stevepolitodesign:
Follow-up to #1201

It's not enough to ensure Node is installed. We also need to ensure the consumer has the supported minimum version installed. Otherwise, subsequent generators will raise errors like so:

```
error stylelint-config-standard-scss@13.1.0: The engine "node" is incompatible with this module. Expected version ">=18.12.0". Got "18.0.0"
error Found incompatible module.
```

We select `v20.0.0` as our minimum supported version because it is slated for Active LTS, but is not bleeding edge at [this time][]

We also raise when calling the template to avoid unnecessarily generating a new Rails application.

[this time]: https://nodejs.org/en/about/previous-releases

### DIFF
--- a/lib/install/web.rb
+++ b/lib/install/web.rb
@@ -1,4 +1,17 @@
+extend Suspenders::Generators::Helpers
+
 def apply_template!
+  if node_not_installed? || node_version_unsupported?
+    message = <<~ERROR
+
+
+    === Node version unsupported ===
+
+    Suspenders requires Node >= #{Suspenders::MINIMUM_NODE_VERSION}
+    ERROR
+
+    fail Rails::Generators::Error, message
+  end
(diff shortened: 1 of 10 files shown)


lib/install/web.rb
```diff
@@ -1,4 +1,27 @@
...
+  ENV["NODE_VERSION"] || `node --version`[/\d+\.\d+\.\d+/]
+end
+
+def node_not_installed?
+  !node_version.present?
+end
+
+def node_version_unsupported?
+  node_version < "20.0.0"
+end
```
stevepolitodesign, line 11: I needed to re-declare these since this template will be treated as a static file when invoked. I can't import Suspenders.

PR #1206, opened 2024-05-15
Prepare for "Tailored" release

stevepolitodesign:
With the merge of #1135 and some subsequent follow-ups, we are ready to
officially release the next version of Suspenders.

Because we moved to [calver][] in #1106, we need to continue using
calver, since the latest release `20230113.0` is greater than `3.0.0`,
which was set in haste in ab3eb9735eabd3c3827c7b6a9cce7d59e079559e

In an effort to better brand this release, we give it the code name
"Tailored".

[calver]: https://calver.org

### DIFF
--- a/lib/suspenders/version.rb
+++ b/lib/suspenders/version.rb
@@ -1,5 +1,5 @@
 module Suspenders
-  VERSION = "3.0.0".freeze
+  VERSION = "20240516.0".freeze
   RAILS_VERSION = "~> 7.0".freeze
   MINIMUM_RUBY_VERSION = ">= 3.1".freeze
   MINIMUM_NODE_VERSION = "20.0.0".freeze
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    suspenders (3.0.0)
+    suspenders (20240516.0)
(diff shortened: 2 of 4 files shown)


lib/suspenders/version.rb
```diff
@@ -1,5 +1,5 @@
 module Suspenders
-  VERSION = "3.0.0".freeze
+  VERSION = "20240516.0".freeze
```
stevepolitodesign, line 2: @mike-burns in #1106 we moved over to [calver](https://calver.org), but now I'm wondering what the ramifications would be if we switched back to [semver](https://semver.org)?

I ask because I've been refering to this next release as `3.0.0` 😅

mike-burns: Release numbers are, like, vibes, ya know?

But since `20230516.0` > `3.0`, I think we're stuck with calver.

stevepolitodesign: > But since 20230516.0 > 3.0, I think we're stuck with calver.

Yeah, I was thinking the same thing. Maybe I can give this release a code-name for better branding. Maybe "tailored".

mike-burns: Or we can hold the release until the year 3000.

I like your code name.

PR #1207, opened 2024-05-16
Application Template: Used released version

stevepolitodesign:
Follow-up to #1206

Now that we've released `20240516.0`, we update the application template
to reference the released version on RubyGems.

This change doesn't require a new release, since this change just
affects the application template, and not the Gem itself.


stevepolitodesign (commented):
I tested this locally with:

```
rails new app_name \
--skip-test
-d=postgresql \
-m=https://raw.githubusercontent.com/thoughtbot/suspenders/sp-update-template/lib/install/web.rb
```

It used the released version, and the app was created successfully.

PR #1208, opened 2024-05-17
Call dynamic-readme reusable workflow

stefannibrasil:
We want to have a way to edit our README footer at one place and have the changes from there be propagated to our repos.

By adding this snippet in the README, we call this reusable workflow: https://github.com/thoughtbot/templates/blob/main/.github/workflows/dynamic-readme.yaml that renders and updates the README footer dynamically.

More details about this approach can be found here: https://thoughtbot.com/blog/automate-processes-with-reusable-github-workflows

### DIFF
--- a/.github/workflows/dynamic-readme.yml
+++ b/.github/workflows/dynamic-readme.yml
@@ -1,0 +1,19 @@
+name: update-templates
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


stevepolitodesign (approved):
Thank you!

PR #1235, opened 2024-10-29
Add node_modules to .gitignore

louis-antonopoulos:
**This PR:**
- Adds `node_modules` to `.gitignore`


stevepolitodesign (approved):
I just ran into this today because I'm on a new machine. My old machine's dotfiles are configured to already ignore `node_modules`, so I never noticed this.

PR #1236, opened 2024-10-30
Use Gem::Version to compare versions

louis-antonopoulos:
**This PR:**
- Uses [Gem::Version](https://docs.ruby-lang.org/en/3.3/Gem/Version.html) to determine if Node is the correct version
- Fixes a potential warning in `WebGeneratorTest`

String comparison of version numbers can produce incorrect results. For example, `"20.10.0" < "20.2.0"` resolves to `true`.


test/generators/suspenders/install/web_generator_test.rb
```diff
@@ -48,10 +48,34 @@ class WebGeneratorTest < Rails::Generators::TestCase
...
+        end
+
         private
 
         def prepare_destination
           touch "Gemfile"
+
+          File.open("test/dummy/Gemfile", "w") do |gemfile|
+            gemfile.write('source "https://rubygems.org"')
+          end
```
louis-antonopoulos, line 78: While working in this file, if nothing was raised when a `run_generator` command ran, then there was a big slowdown and the following message would appear multiple times:

```
...[DEPRECATED] This Gemfile does not include an explicit global source. Not using an explicit global source may result in a different lockfile being generated depending on the gems you have installed locally before bundler is run. Instead, define a global source in your Gemfile like this: source "https://rubygems.org".
```

louis-antonopoulos: Adding `source "https://rubygems.org"` to the `Gemfile` resolved this warning, making it easier to test the tests in the file.

Note that `run_generator` still fails for different reasons, but it's a faster fail with less noise.

lib/install/web.rb
```diff
@@ -1,13 +1,19 @@
 def node_version
-  ENV["NODE_VERSION"] || `node --version`[/\d+\.\d+\.\d+/]
+  version = ENV["NODE_VERSION"] || `node --version`[/\d+\.\d+\.\d+/]
+
+  return if version.blank?
+
+  Gem::Version.new(version)
```
louis-antonopoulos, line 6: Do you think these two lines are clearer as written or combined like this?

```
Gem::Version.new(version) unless version.blank?
```

lib/install/web.rb
```diff
@@ -1,13 +1,19 @@
...
 end
 
 def node_not_installed?
-  !node_version.present?
+  node_version.blank?
```
louis-antonopoulos, line 10: I changed this to avoid the double negative of `not_installed` and `!(...).present?`.

lib/install/web.rb
purinkle, line 6: Would the following work?

```suggestion
  if version = ENV.fetch(ENV["NODE_VERSION"], `node --version`[/\d+\.\d+\.\d+/])
    Gem::Version.new(version)
  else
    ""
  end
```

It solves a couple of issues.

1. Deals with `#[]` returning something falsy but not nil
2. Gets rid of the trailing conditional and makes the method more idiomatic
3. Ensures the method always returns a string

Issue #1233, opened 2024-10-25
Requiring --skip_rubocop in command line parameters causes script failure when Rails < 7.2


louis-antonopoulos:
The [web.rb](https://github.com/thoughtbot/suspenders/blob/7cb8ebdbf28f768b1f45d80c7aea7b39595d80fb/lib/install/web.rb#L25) script fails when running versions of Rails before 7.2 because `--skip_rubocop` is not a valid option, and so it is not parsed.

#1223 added a requirement to specify the `--skip_rubocop` parameter to resolve #1221.

louis-antonopoulos:
(identified while pairing with @stevepolitodesign)

stevepolitodesign:
I think we have a few options here:

- We could update the README by encouraging users to use the latest version of Rails.
- We could return early in the [application template](https://github.com/thoughtbot/suspenders/blob/main/lib/install/web.rb) if the detected version is below `7.2`

louis-antonopoulos:
I'm going to be opening a PR to return early as you suggest!

PR #1237, opened 2024-10-31
Require Rails 7.2 in web template

louis-antonopoulos:
**This PR:**
- Closes [Issue 1233: Requiring --skip_rubocop in command line parameters causes script failure when Rails < 7.2](https://github.com/thoughtbot/suspenders/issues/1233)


lib/install/web.rb
```diff
@@ -10,6 +10,20 @@ def node_version_unsupported?
...
+  return if version.blank?
+
+  Gem::Version.new(version)
+end
+
+def rails_version_unsupported?
+  minimum_rails_version = Gem::Version.new("7.2.0")
+
+  rails_version < minimum_rails_version
+end
```
stevepolitodesign, line 25: I experimented with this locally, and we have access to [`Rails#gem_version`](https://api.rubyonrails.org/classes/Rails.html#method-c-gem_version) in the template. This is neat because it returns the [version](https://api.rubyonrails.org/classes/Rails.html#method-c-version) as an instance of `Gem::Version`.

```suggestion
def rails_version_unsupported?
  Rails.gem_version < Gem::Version.new("7.2.0")
end
```

louis-antonopoulos: That's great and cleaned this up a lot! `Rails.gem_version` implemented in 62105edb3fc797c0f6ec983d50988ad4176f9b96.

louis-antonopoulos: Make that 67f7a30fa6d0ec9055edce5f15bac1f5d41f0526.

slickepinne (commented):
What do you think of also updating the README to name `Rails ~> 7.2.0` instead of `Rails ~> 7.0` as a requirement for new applications? Or a note that this is the minimum version for new applications?

PR #1239, opened 2024-11-29
Add support for Rails 8.0

### LINKED ISSUE #1213: Prepare for Rails 8
With the upcoming release of [Rails 8](https://github.com/rails/rails/issues?q=is%3Aopen+is%3Aissue+milestone%3A8.0.0), we might want to consider re-generating the [Engine](https://guides.rubyonrails.org/engines.html) to extract the [dummy app](https://guides.rubyonrails.org/engines.html#testing-an-engine). This way, our tests are running against a Rails 8 dummy app, and might catch regressions.

nickcharlton:
Here, we loosen the constraints to allow us to use Rails 8.0, and update the `dummy` app so that we're testing against Rails 8.0.

Closes #1213.


stevepolitodesign:
Just noting we might want to make this a feature branch, as I suspect we'll need to create several commits to complete this effort. Similar to how we handled the re-write in #1135

stevepolitodesign:
Noting, we'll also want to update the README.

nickcharlton:
I'd apparently forgotten that this had failing lint and test runs, so I've just fixed those.

As this is, it'd be helpful to squash these commits into one, well-written one and then merge it. There are other changes we know we want to make to Suspenders, some of which would be worth doing before cutting a release, but doing it that way will be much more straightforward than doing another approach.

stevepolitodesign (approved):

PR #1241, opened 2024-12-05
Add option to install local build

stevepolitodesign:
In the process of supporting Rails 8 via #1213, we realized we needed to
install a local build.

This commit simply introduces a `--suspenders-local` option, and
documents how to build locally.

Co-Authored-By: Nick Charlton <nick@nickcharlton.net>


lib/install/web.rb
```diff
@@ -27,6 +27,8 @@ def apply_template!
       gem_group :development, :test do
         if ARGV.include?("--suspenders-main")
           gem "suspenders", github: "thoughtbot/suspenders", branch: "main"
+        elsif ARGV.include?("--suspenders-local") && ENV["SUSPENDERS_LOCAL_PATH"]
+          gem "suspenders", path: ENV.fetch("SUSPENDERS_LOCAL_PATH")
```
stevepolitodesign, line 31: This feels a little redundant. Would this be better?

  ```suggestion
        elsif ENV["SUSPENDERS_LOCAL_PATH"].present?
          gem "suspenders", path: ENV.fetch("SUSPENDERS_LOCAL_PATH")
```

Then, we could omit `--suspenders-local` altogether.

 ```
 SUSPENDERS_LOCAL_PATH=/path/to/suspenders \
  rails new app_name \
  --skip-rubocop \
  --skip-test \
  -d=postgresql \
 ```

Comment (nickcharlton) [reply lib/install/web.rb:31]: Could we just let it use any path you give it to the template? Do you know if we have the value of `-m` at this point?

I'm thinking you could do something like this then:

```
value_of_m = "https://raw.githubusercontent.com/thoughtbot/suspenders/main/lib/install/web.rb"
uri = URI.parse(value_of_m)

if uri.kind_of? URI::Generic
  gem "suspenders", path: File.expand_path(value_of_m)
end
```

[This is just playing around with a URL parsing example](https://stackoverflow.com/a/13041565)

stevepolitodesign:
This is no longer relevant thanks to the [latest release.](https://github.com/thoughtbot/suspenders/releases/tag/v20251219.0). We can run `bin/qa` locally.

PR #1244, opened 2024-12-06
Generate Gemfile entry with version constraint

seanpdoyle:
When generating a new application by loading the template over HTTP, specifying `gem "suspenders"` without a version constraint results in the resolution of an extremely outdated version ([v0.2.4][]).

When omitting the `--suspenders-main` argument, include a version constraint to use a version that is at least `3.0.0`.

[v0.2.4]: https://github.com/thoughtbot/suspenders/tree/v0.2.4


lib/install/web.rb
```diff
@@ -28,7 +28,7 @@ def apply_template!
         if ARGV.include?("--suspenders-main")
           gem "suspenders", github: "thoughtbot/suspenders", branch: "main"
         else
-          gem "suspenders"
+          gem "suspenders", "> 20240101"
```
stevepolitodesign, line 31: I _think_ I tried (and failed) to see if I could do this dynamically by including `Suspenders::Version` in this file, and then doing something like this:

```suggestion
          gem "suspenders", "> #{Suspenders::VERSION}"
```

The problem is that when this file is invoked, it's outside the context of Suspenders, so if Suspenders is not installed on the system, it will error.

seanpdoyle: Right, this is a Ruby file executed in its own environment without access to Bundler or other dependencies. I chose `20240101` because its a lower limit on the possible values (versions starting with the `2024`-prefix), not necessarily because it's a valid version that has a release.

stevepolitodesign: > this is a Ruby file executed in its own environment without access to Bundler or other dependencies

I wonder if we could do something similar to the [Rails reproduction scripts](https://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#create-an-executable-test-case), but for now, this works.

stevepolitodesign: As an aside, I thought [leaving off the version meant we'd download the latest version](https://thoughtbot.com/blog/a-healthy-bundle#versionless)?

seanpdoyle: > I wonder if we could do something similar to the [Rails reproduction scripts](https://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#create-an-executable-test-case)

While that might increase the runtime of `rails new`, that might be worthwhile, since it's (ideally) only executed once per project.

> As an aside, I thought [leaving off the version meant we'd download the latest version](https://thoughtbot.com/blog/a-healthy-bundle#versionless)?

Ideally, that would be true. In practice, `rails new` would consistently install `v0.2.4`.

test/generators/suspenders/install/web_generator_test.rb
```diff
@@ -12,6 +12,14 @@ class WebGeneratorTest < Rails::Generators::TestCase
         setup :prepare_destination
         teardown :restore_destination
 
+        test "generates Gemfile entry with version constraint" do
+          with_database "postgresql" do
+            run_generator
+          end
+
+          assert_file "Gemfile", /gem "suspenders", "> 20240101"/
+        end
```
stevepolitodesign, line 21: Were you able to get this to pass locally? I'm having some local dependency issues, so I can't verify this locally.

That being said, the file we're modifying in this commit is `lib/install/web.rb`, which is not a generator, but the application template.

This test is  responsible for testing [`Suspenders ::Generators::Install:: WebGenerator`](https://github.com/thoughtbot/suspenders/blob/include-gemfile-version/lib/generators/suspenders/install/web_generator.rb), which doesn't install Suspenders.

stevepolitodesign: Unfortunately, we don't have any coverage for `lib/install/web.rb`.

Issue #1243, opened 2024-12-05
New Rails 8 applications generate two CI files


stevepolitodesign:
Relates to #1213 

We create files that end in `.yaml`, and Rails created files that end in `.yml`

https://github.com/thoughtbot/suspenders/blob/bad1d016796223729d713d06261dd722c5dae56f/lib/generators/suspenders/ci_generator.rb#L17-L18

stevepolitodesign:
I think we can rely on the `dependabot.yml` file generated by Rails, but need to replace the `ci.yml` file generated by Rails.

nickcharlton:
Oh yeah, weird. I _thought_ I always used the longer form, but I seem to use the three letter variant.

I also prefer to use either `main.yml` or `tests.yml` instead of `ci.yml` for the name. My thinking here is that "CI" is meaningless in this context, all the files are "CI" as we've come to understand that term.

Then I'll have different files so that we break out the feedback you get on the GitHub statuses view:

* tests
* linting
* bundle audit
* etc

stevepolitodesign:
We could call `--skip-ci` when running [rails new](https://edgeguides.rubyonrails.org/command_line.html#rails-new) to make it simpler for use to maintain these files.

> I also prefer to use either main.yml or tests.yml instead of ci.yml for the name. My thinking here is that "CI" is meaningless in this context, all the files are "CI" as we've come to understand that term.
>
> Then I'll have different files so that we break out the feedback you get on the GitHub statuses view:

It looks like having one file will still break out each task into its own status. Below is a screenshot from a personal project that is using the stock CI file from Rails. 

![CleanShot 2024-12-06 at 09 27 23](https://github.com/user-attachments/assets/ec93f5cf-e284-4ec7-b36b-5137ca559208)

nickcharlton:
Yeah, I'd expect it break things out like that. Much of my intention from breaking out the files is to not have horribly long YAML files which are tricky to understand.

nickcharlton:
I've opened #1246, #1247 to try and address both of the cases mentioned here.

It doesn't look like we've got any more YAML files, so merging both of those should wrap this issue up.

PR #1246, opened 2025-01-31
Switch to using yml for generated yaml files

nickcharlton:
See: #1243


stevepolitodesign (approved):
We might need to remove the [`github/workflows/ci.yml`](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt) that ships with Rails, but that can be task for a separe commit.

PR #1247, opened 2025-01-31
Remove dependabot config in favour of upstream

nickcharlton:
Rails has started providing a `dependabot.yml` file, so instead of maintaining our own we can rely on that instead.

See: #1243


stevepolitodesign (approved):
Thank you!

PR #1248, opened 2025-02-12
Add support for SQLite

### LINKED ISSUE #1245: We don't just prefer, but require the use of Postgres
I've been working on [updating a little script I have for testing out Administrate](https://github.com/nickcharlton/dotfiles/blob/main/bin/make-administrate-demo), where I'll go about creating a bunch of small Rails applications for testing. This has one problem of leaving a lot of unused, potentially clashing Postgres databases sitting around.

To work around this problem, I considered using SQLite — [especially as there's been significant renewed interest in it](https://fractaledmind.github.io/tags/sqlite/) up to and including Rails 8.0.

Whilst [we prefer Postgres](https://github.com/thoughtbot/dotfiles/pull/728) (and I don't see a need to generally diverge from this opinion), the way we've implemented `lib/install/web.rb` stops us from using another database type, for example:

```sh
$ rails new app_name \
 --suspenders-local \
 --skip-rubocop \
 --skip-test \
 -d=postgresql \
 -m=~/projects/thoughtbot/suspenders/lib/install/web.rb

[… snip …]

remove  config/initializers/cors.rb
      remove  config/initializers/new_framework_defaults_8_0.rb
       apply  /Users/nickcharlton/projects/thoughtbot/suspenders/lib/install/web.rb
  in the local project!
The template [/Users/nickcharlton/projects/thoughtbot/suspenders/lib/install/web.rb] could not be loaded. Error:

=== Please use the correct options ===

# Use the latest suspenders release:
rails new <app_name> \
--skip-rubocop \
--skip-test \
-d=postgresql \

# OR use the current (possibly unreleased) `main` branch of suspenders:
rails new <app_name> \
--suspenders-main \
--skip-rubocop \
--skip-test \
-d=postgresql \
```

The implementation of `lib/install/web.rb` looks like this:  https://github.com/thoughtbot/suspenders/blob/a4a4af047526c3f4f79b0bed40c7610b185789da/lib/install/web.rb#L25-L33

Can anyone think of a better way to do that test in `options[:database]`, etc? The intention is to see if we should setup and run our suspenders generator, but it's doing more than that.

nickcharlton:
With increasing interest in SQLite across the Rails community as a whole, it's valuable for us to start exploring it too. This expands the existing checks for PostgreSQL to also include SQLite so that where we want to try out SQLite we're able to. It also adjusts the CI workflow to exclude PostgreSQL when generating a SQLite-based application.

https://rubyonrails.org/2024/11/7/rails-8-no-paas-required
https://fractaledmind.github.io/tags/sqlite/

Closes #1245

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ This approach uses an [application template][] to generate a new Rails
 application with Suspenders.
 
 We skip the [default test framework][] in favor of [RSpec][], and [prefer
-PostgreSQL][] as our database.
+PostgreSQL][] as our database, although we do support SQLite.
 
 We skip [RuboCop rules by default][] in favor of our [holistic linting rules][].
 
--- a/lib/generators/templates/ci/ci.yml.tt
+++ b/lib/generators/templates/ci/ci.yml.tt
@@ -86,6 +86,7 @@ jobs:
     runs-on: ubuntu-latest
 
     services:
+      <%- if using_postgres? -%>
       postgres:
         image: postgres
         env:
@@ -94,6 +95,7 @@ jobs:
         ports:
           - 5432:5432
         options: --health-cmd="pg_isready" --health-interval=10s --health-timeout=5s --health-retries=3
+      <%- end %>
 
       # redis:
       #   image: redis
@@ -127,7 +129,9 @@ jobs:
       - name: Run tests
         env:
           RAILS_ENV: test
+          <%- if using_postgres? -%>
           DATABASE_URL: postgres://postgres:postgres@localhost:5432
+          <%- end -%>
           # REDIS_URL: redis://localhost:6379/0
         <%- if using_rspec? -%>
         run: bin/rails db:setup spec
--- a/test/generators/suspenders/ci_generator_test.rb
+++ b/test/generators/suspenders/ci_generator_test.rb
@@ -21,7 +21,7 @@ module Suspenders
 
       test "raises if PostgreSQL is not the adapter" do
         with_database "unsupported" do
-          assert_raises Suspenders::Generators::DatabaseUnsupported::Error, match: "This generator requires PostgreSQL" do
+          assert_raises Suspenders::Generators::DatabaseUnsupported::Error, match: "This generator requires either PostgreSQL or SQLite" do
             run_generator
 
             assert_no_file app_root(".github/workflows/ci.yaml")
--- a/lib/generators/suspenders/ci_generator.rb
+++ b/lib/generators/suspenders/ci_generator.rb
@@ -36,6 +36,10 @@ module Suspenders
         File.exist? "package.json"
       end
 
+      def using_postgres?
+        database_adapter == "postgresql"
+      end
+
       def has_gem?(name)
         Bundler.rubygems.find_name(name).any?
       end
--- a/lib/install/web.rb
+++ b/lib/install/web.rb
@@ -10,6 +10,10 @@ def node_version_unsupported?
   node_version < "20.0.0"
 end
 
+def database_supported?
+  ["postgresql", "sqlite"].include?(options[:database])
+end
+
 def apply_template!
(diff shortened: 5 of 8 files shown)


README.md
```diff
@@ -31,7 +31,7 @@ This approach uses an [application template][] to generate a new Rails
 application with Suspenders.
 
 We skip the [default test framework][] in favor of [RSpec][], and [prefer
-PostgreSQL][] as our database.
+PostgreSQL][] as our database, although we do support SQLite.
```
nickcharlton, line 34: I'm being a bit … careful (?) with my README changes here. I want to suggest it's possible, but we're just trying it out, and it's worth evaluating and not that you should necessarily switch away from Postgres.

That said, maybe we need an explicit example?

purinkle:
We are trying to keep Suspenders as close to Rails defaults as possible, but I still expect us to have an opinion. Otherwise, we should run `rails new` and be done with it. Are we trying out SQLite? Should team members start new Rails apps with SQLite? Suspenders should reflect our practices, not our aspirations.

MatheusRich:
> Are we trying out SQLite? Should team members start new Rails apps with SQLite? 

I think there are a number of (small?) projects that would benefit from a simpler setup. You just want to try out an idea or build a side project. That shouldn't require you to deploy 2 different servers

test/generators/suspenders/ci_generator_test.rb
```diff
@@ -21,7 +21,7 @@ class CiGeneratorTest < Rails::Generators::TestCase
 
       test "raises if PostgreSQL is not the adapter" do
```
louis-antonopoulos, line 22: 1) Should this test description reflect "PostgreSQL or SQLite"?

2) Should this or another test assert that `with_database "sqlite"` does not raise?

lib/generators/templates/ci/ci.yml.tt
```diff
@@ -86,6 +86,7 @@ jobs:
     runs-on: ubuntu-latest
 
     services:
+      <%- if using_postgres? -%>
```
louis-antonopoulos, line 89: Should this block be tested that it is present for Postgres and absent for SQLite?

lib/generators/templates/ci/ci.yml.tt
stevepolitodesign, line 1: We might want to borrow from the [Rails CI](https://github.com/rails/rails/blob/3b79de3df3d200acd0f291b1a08d9181ddee4605/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt#L66-L73) template.

README.md
stevepolitodesign: Yeah, this is tricky, and I'm not sure the best approach to take. Suspenders is about limiting the choices we need to make, but I recognize we're in an exploratory phase.

I don't know if we have an explicit example (yet), but I imagine it would relate to avoiding a Platform as a service.

stevepolitodesign:
> That being said, should we just run
> 
> ```shell
> rails db:system:change --to=sqlite3
> ```

This is interesting, and I'm curious to see how this works on a new app generated by Suspenders.I imagine it would still require manually editing the CI file, though.

@nickcharlton is this worth exploring?

README.md
nickcharlton: I was on the fence about putting a change in at all, but as you can't comment on things you haven't changed in a PR something had to do 😅 .

It's almost like it either needs to be missing, or come with an explanation and I'm not sure what I'd want to do there.

lib/generators/templates/ci/ci.yml.tt
nickcharlton: Oh, interesting!

I also note that they're actually using Valkey, which diverges in another way from what we usually do: 

https://github.com/rails/rails/blob/3b79de3df3d200acd0f291b1a08d9181ddee4605/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt#L66-L73

PR #1249, opened 2025-02-17
Force creation of GitHub Actions CI workflow

nickcharlton:
Since we switched to using `.yml` for YAML files in #1246, generating the CI config would clash and request the user intervene. We force the template to overwrite instead.

Completes #1243.

### DIFF
--- a/lib/generators/suspenders/ci_generator.rb
+++ b/lib/generators/suspenders/ci_generator.rb
@@ -12,8 +12,7 @@ module Suspenders
       MARKDOWN
 
       def ci_files
-        empty_directory ".github/workflows"
-        template "ci.yml", ".github/workflows/ci.yml"
+        template "ci.yml", ".github/workflows/ci.yml", force: true
       end
 
       private
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -35,7 +35,7 @@ module Suspenders::TestHelpers
   def mkdir(dir)
     path = app_root dir
 
-    FileUtils.mkdir path
+    FileUtils.mkdir_p path
   end
 
   def touch(file, **options)
--- a/test/generators/suspenders/ci_generator_test.rb
+++ b/test/generators/suspenders/ci_generator_test.rb
@@ -11,6 +11,9 @@ module Suspenders
       teardown :restore_destination
 
       test "generates CI files" do
+        mkdir(".github/workflows")
+        touch(".github/workflows/ci.yml")
+
         with_database "postgresql" do
           run_generator
 


nickcharlton:
I noticed this after generating a new application after merging a bunch of Rails 8.0 support PRs in. I don't know if there's actually a better way to approach this, or if there's a way we can test for it? Thoughts welcome!

We do no longer have any `.yaml` files (`find . -name "*.yaml"`) left though after this.

lib/generators/suspenders/ci_generator.rb
```diff
@@ -13,7 +13,7 @@ class CiGenerator < Rails::Generators::Base
 
       def ci_files
         empty_directory ".github/workflows"
-        template "ci.yml", ".github/workflows/ci.yml"
+        template "ci.yml", ".github/workflows/ci.yml", force: true
```
stevepolitodesign, line 16: Is it possible/worth adding test coverage for this by first adding a dummy `.github/workflows/ci.yml` in the setup, and ensuring our template overrides it?

lib/generators/suspenders/ci_generator.rb
```diff
@@ -13,7 +13,7 @@ class CiGenerator < Rails::Generators::Base
 
       def ci_files
         empty_directory ".github/workflows"
```
stevepolitodesign, line 15: I wonder if we need this anymore?

```suggestion
```

lib/generators/suspenders/ci_generator.rb
nickcharlton: Oh, yes! That'd do it (and make me happy.)

lib/generators/suspenders/ci_generator.rb
nickcharlton: Ooh it turns out we can.

lib/generators/suspenders/ci_generator.rb
nickcharlton: This did indeed work well, thanks for the suggestion!

test/test_helper.rb
```diff
@@ -38,6 +38,12 @@ def mkdir(dir)
     FileUtils.mkdir path
   end
 
+  def mkdir_p(dir)
```
nickcharlton, line 41: I feel like having both `mkdir` and `mkdir_p` is a bit unnecessary, but it's better than two calls.

stevepolitodesign: TIL: `FileUtils.mkdir_p`.

Could we just update the existing `mkdir` to call `FileUtils.mkdir_p`? It shouldn't break the existing behavior.

stevepolitodesign (approved):
Thank you! Just one minor comment worth considering, but not blocking.

test/test_helper.rb
nickcharlton: We could, yeah. I don't think it'd be surprising either, so I think I'll do that.

stevepolitodesign:
@nickcharlton I'm just now realizing we haven't been updating [`NEWS.md`](https://github.com/thoughtbot/suspenders/blob/main/NEWS.md) with these changes. We should update it to reflect all the work you've done.

nickcharlton:
Oh yeah, we've not. I'm not sure what we usually do here, but I typically fill out a CHANGELOG when I cut a new release.

PR #1251, opened 2025-03-10
Introduce `development:seed` and `development:seed:replant` tasks

stevepolitodesign:
Since `db:seeds.rb` is intended to ensure

> ... the existence of records required to run the application in every
environment (production, development, test)

we have relied on [`dev:prime`][1] for loading data necessary for users
to view most of the features of the app in development.

However, there were a few areas of improvement.

First, rename `dev` namespace to `development`, and rename [`prime`][1]
task to `seed` for improved clarity.

The, introduces `development:seed:replant` take to create parity with
the existing `db:seed:replant` task.

Introduces `Seeder` class to encapsulate seed data, which is called in
the `development:seed` and `development:seed:replant` tasks. The
advantage to this is that the class can be called independently from the
task, and can also be [unit tested][2].

[1]: https://thoughtbot.com/blog/priming-the-pump
[2]: https://thoughtbot.com/blog/seeds-of-destruction


lib/generators/templates/tasks/development.rake
```diff
@@ -0,0 +1,11 @@
+namespace :development do
+  desc "Loads seed data into development."
+  task seed: :environment do
+    Seeder.load_seeds
```
stevepolitodesign, line 4: If we don't care about unit testing this, we could always load the seeds directly in this file, like before.

Curious to hear what others think.

lib/generators/templates/tasks/development.rake
```diff
@@ -0,0 +1,11 @@
+namespace :development do
+  desc "Loads seed data into development."
+  task seed: :environment do
```
stevepolitodesign, line 3: I should consider wrapping this in the `db` namespace.

```
bin/rails development:db:seed
bin/rails development:db:seed:replant
```

nickcharlton:
I've found that people find `dev:prime` to be confusing over the years, so I think this is a good idea.

One of the nice things about `dev:prime` was that it read nicely when in scripts. I'd not call any other database tasks, just that one and it looked good:

https://github.com/thoughtbot/administrate/blob/9870b40ce7330e64acad04a6c1685299defff77b/bin/setup#L25

lib/generators/templates/setup/bin_setup.rb
```diff
@@ -20,11 +20,8 @@ def using_node?
   system("yarn install --check-files") if using_node?
 
   puts "\n== Preparing database and adding development seed data =="
-  if File.exist? "lib/tasks/dev.rake"
-    system! "bin/rails dev:prime"
-  else
-    system! "bin/rails db:prepare"
-  end
+  system! "bin/rails db:prepare"
+  system! "bin/rails development:seed" if File.exist? "lib/tasks/development.rake"
```
stevepolitodesign, line 24: @nickcharlton 

> One of the nice things about `dev:prime` was that it read nicely when in scripts. I'd not call any other database tasks, just that one and it looked good:
> 
> https://github.com/thoughtbot/administrate/blob/9870b40ce7330e64acad04a6c1685299defff77b/bin/setup#L25

Are you suggesting something like this?

```suggestion
  system! "bin/rails development:seed" if File.exist? "lib/tasks/development.rake"
```

nickcharlton: Right — you don't need to call `db:prepare`, because the task for seeding calls it in it's dependency chain.

stevepolitodesign: It does? I might be misunderstanding, but the newly introduced `development:seed` has no other dependencies other than the `:environment`. 

https://github.com/thoughtbot/suspenders/blob/04589ab9ccc8472031de30d8e0a714560d4be5ba/lib/generators/templates/tasks/development.rake#L3

nickcharlton: I mean that in comparison to the original (not the new one!), one of the "hidden" features of it was that it'd also setup the database for you. This is why Administrate's `bin/setup` is so small.

stevepolitodesign:
I think CI is failing because of this line:

```
# config/environments/test/.rb
config.eager_load = ENV["CI"].present?
```

PR #1255, opened 2025-04-11
Load application.css instead of `:app`

gabebw:
`:app` is new in Rails 8 and has a special meaning for Propshaft. It is not the same as `:application`.

In Propshaft, as shown [here][1], `:app` loads all of the CSS files in `app/assets`.. That causes an error with a the Suspenders setup, because it loads `application.css` then tries (and fails) to load all of the `@import`ed files as well.

The fix is to only load `"application.css"`.

[1]: https://github.com/rails/propshaft/blob/689e756689baa141291c7422f6a2df41b5ceead8/lib/propshaft/helper.rb#L13

This happens in both the released and the main branch.

## To replicate the bug

* Install rails: `gem install rails -v 8.0.2`
* Create new Rails app (include `--suspenders-main` if you like):
  ```
  rails _8.0.2_ new my_app_name \
    --skip-rubocop \
    --skip-test \
    -d=postgresql \
 ```
* `cd my_app_name`
* `rails g controller homes cool_test`
* `rails s`
* Now visit `http://localhost:3000/homes/cool_test` and you'll see 404 errors for various assets including `modern_normalize` in the web inspector

Tests will also fail. To see that, create `spec/system/something_spec.rb`:

```ruby
require "rails_helper"

RSpec.describe "Assets" do
  it "works" do
    visit homes_cool_test_path
  end
end
```

* Run `rake`
* Among the errors you will see:
  ```
  1.2) Failure/Error: raise ActionController::RoutingError, "No route matches [#{env['REQUEST_METHOD']}] #{env['PATH_INFO'].inspect}"

          ActionController::RoutingError:
            No route matches [GET] "/homes/modern-normalize"

          [Screenshot Image]: /Users/gabe/my_app_name/tmp/capybara/failures_r_spec_example_groups_assets_works_869.png
```

### CODE REVIEW

File: test/dummy/app/views/layouts/application.html.erb [hunk 1]
```diff
@@ -6,7 +6,8 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
-    <%= stylesheet_link_tag "application" %>
+    <%# Includes all stylesheet files in app/assets/stylesheets %>
+    <%= stylesheet_link_tag :app, "data-turbo-track": "reload" %>
```
gabebw, line 10: This is what Rails 8 generates. Should this PR remove `"data-turbo-track": "reload"` as well?

PR #1272, opened 2025-12-12
Introduce Strong Migrations 

laicuRoot:
Currently Suspenders does not include strong_migrations per default.
This is a thoughtbot recommendation in any new application.

In this PR, we introduce strong_migrations to be part of the defaults of
any new application created using suspenders.

[Strong Migrations](https://github.com/ankane/strong_migrations)


laicuRoot:
I've tested in a new application and it worked as expected. 


Some logs from the initialisation process:
image

Gemfile:
image

PR #1276, opened 2025-12-12
Improve CI script and fix broken layout

stevepolitodesign:
Prior to this commit, we had no specs to test.  This commit updates the CI script by adding a basic static page, and adding a corresponding system spec.

This revealed we had a broken layout, because we were including a partial that does not exist.


stevepolitodesign:
I deliberately made a failure in 2163de57c40f38e486587d93f04769679b2eb682 to make sure CI would fail, which it did.

PR #1352, opened 2026-04-27
Add simple AI Harness

stevepolitodesign:
Relates to https://github.com/thoughtbot/suspenders/discussions/1350 and
https://github.com/thoughtbot/suspenders/discussions/1351.

We introduced [thoughtbot agents
rules](https://github.com/thoughtbot/guides/pull/783) to our guides a
few months ago. This commit simply pulls those files into newly
generated Suspenders applications. There's a case to be made that these
should be kept separate, but I like having a single source of truth.

### DIFF
--- a/lib/templates/web.rb
+++ b/lib/templates/web.rb
@@ -1,4 +1,6 @@
 require "suspenders/version"
+require "net/http"
+require "json"
 
 # Methods like `copy_file` will accept relative paths to the template's location.
 def source_paths
@@ -69,6 +71,7 @@ after_bundle do
   # Finalization
   run_migrations
   update_readme
+  add_ai_harness
   lint_codebase
   commit_final_application_state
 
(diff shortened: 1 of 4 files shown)


lib/templates/web.rb
```diff
@@ -524,6 +527,24 @@ def lint_codebase
...
 end
 
+def add_ai_harness
+  base_url = "https://raw.githubusercontent.com/thoughtbot/guides/main/rails/ai-rules"
+  api_url = "https://api.github.com/repos/thoughtbot/guides/contents/rails/ai-rules/rules"
+
+  empty_directory ".claude"
+  empty_directory ".claude/rules"
+
+  get "#{base_url}/CLAUDE.md", ".claude/CLAUDE.md"
```
stevepolitodesign, line 559: One issue here is this is Claude-specific. That works for thoughtbot, but do we want to support other services?

jaredlt: FYI it's being discussed here https://github.com/thoughtbot/guides/pull/786#issuecomment-4325540433

jaredlt:
This looks great!

Annoyingly, this will include the [usage section](https://github.com/thoughtbot/guides/blob/main/rails/ai-rules/CLAUDE.md?plain=1#L6-L12) which we've effectively already done. I can't think of a non-fragile way to include these. Unless perhaps we add a README.md in the same guides folder with these instructions and then the CLAUDE.md is "ready" to use, they just need to update the app name and add the sentence on app context. @laicuRoot  what do you think?

stevepolitodesign:
@jaredlt @laicuRoot alternatively, would it be worth making Suspenders the source of truth for this? Then, we could update the Guides to link to these files?

laicuRoot:
Hey folks!

I think extracting the instructions to a README makes sense @jaredlt 👍 

> would it be worth making Suspenders the source of truth for this?

🤔 I'm not sure about this @stevepolitodesign. Shouldn't our guides be the source of truth for how we work and how we recommend writing code? In any case, if you think that would make the rules more "maintainable" I'm ok with it. 

By the way, now rails ships with an [AGENT.md file](https://github.com/rails/rails/blob/main/AGENTS.md).

jaredlt:
I believe the AGENTS.md file in the Rails repo is more about contributing to the rails repo, and not so much about developing apps using rails.

I can't decide what is best about the source of truth for where this should live :)

stevepolitodesign:
@jaredlt @laicuRoot I've opened https://github.com/thoughtbot/guides/pull/789 to address the concerns mentioned in this PR. Once that merges, we can merge this.

lib/templates/web.rb
```diff
@@ -524,6 +527,24 @@ def lint_codebase
...
+
+  response = Net::HTTP.get(URI(api_url))
+  files = JSON.parse(response)
+  files.each do |file|
+    next unless file["type"] == "file"
+
+    get "#{base_url}/rules/#{file["name"]}", ".claude/rules/#{file["name"]}"
+  end
```
stevepolitodesign, line 568: Noting we'll want to ignore the README once https://github.com/thoughtbot/guides/pull/789 merges.

stevepolitodesign:
@jaredlt @laicuRoot now that https://github.com/thoughtbot/guides/pull/789 has merged, I think we're in a good state. You can also manually test this by pulling down the project and running `bin/qa`..

We also have coverage in [CI](https://github.com/thoughtbot/suspenders/actions/runs/25113848968/job/73595135909?pr=1352).

jaredlt (approved):
Looks good!

I think we can explore supporting other agents etc in a future iteration.

laicuRoot (approved):
🦾

PR #1353, opened 2026-05-08
Seed development seeds during bin/setup

jaredlt:
Before, the development seeder was only reachable by manually running `bin/rails development:db:seed` after setup. Developers had to know to run this separately.

Now, bin/setup calls `development:db:seed` automatically after preparing the database. An environment guard skips the call in non-development environments.

---

As I was playing with a new suspenders app, I added development seeds to the seeder and then ran bin/setup expecting that the seeds would be populated for me. I didn't realise I needed to run the development seed command directly. I think this could be a nice quality of life improvement. Given data in the development seeder is expected to be seeded in development, then running bin/setup can help do this for us.

Let me know what you think.

### DIFF
--- a/lib/templates/web.rb
+++ b/lib/templates/web.rb
@@ -242,6 +242,14 @@ end
 def configure_development_seeder
   copy_file "lib/development/seeder.rb"
   copy_file "lib/tasks/development.rake"
+  inject_into_file "bin/setup", after: /system!.*db:reset.*\n/ do
+    "\n" + <<~RUBY.gsub(/^/, "  ")
+      if ENV.fetch("RAILS_ENV", "development") == "development"
+        puts "\\n== Loading development seed data =="
+        system! "bin/rails development:db:seed"
+      end
+    RUBY
+  end
 end
 
 def setup_test_environment
--- a/lib/templates/web.rb
+++ b/lib/templates/web.rb
@@ -244,7 +244,7 @@ def configure_development_seeder
   copy_file "lib/tasks/development.rake"
   inject_into_file "bin/setup", after: /system!.*db:reset.*\n/ do
     "\n" + <<~RUBY.gsub(/^/, "  ")
-      if ENV("RAILS_ENV") == "development"
+      if ENV["RAILS_ENV"] == "development"
         puts "\\n== Loading development seed data =="
         system! "bin/rails development:db:seed"
       end
--- a/lib/templates/web.rb
+++ b/lib/templates/web.rb
@@ -244,7 +244,7 @@ def configure_development_seeder
   copy_file "lib/tasks/development.rake"
   inject_into_file "bin/setup", after: /system!.*db:reset.*\n/ do
     "\n" + <<~RUBY.gsub(/^/, "  ")
-      if ENV["RAILS_ENV"] == "development"
+      if ENV.fetch("RAILS_ENV", "development") == "development"
         puts "\\n== Loading development seed data =="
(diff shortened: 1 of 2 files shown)


lib/templates/web.rb
```diff
@@ -241,6 +241,14 @@ def configure_inline_svg
 def configure_development_seeder
   copy_file "lib/development/seeder.rb"
   copy_file "lib/tasks/development.rake"
+  inject_into_file "bin/setup", after: /system!.*db:reset.*\n/ do
+    "\n" + <<~RUBY.gsub(/^/, "  ")
+      if ENV.fetch("RAILS_ENV", "development") == "development"
```
stevepolitodesign, line 247: I think we want this. Otherwise, if the value isn't set, it would default to `"development"`, which could be a problem in a deployed environment.

```suggestion
      if ENV("RAILS_ENV") == "development"
```

jaredlt: Ohhh yes, good catch. Thank you 👍

lib/templates/web.rb
```diff
@@ -241,6 +241,14 @@ def configure_inline_svg
 def configure_development_seeder
   copy_file "lib/development/seeder.rb"
   copy_file "lib/tasks/development.rake"
+  inject_into_file "bin/setup", after: /system!.*db:reset.*\n/ do
+    "\n" + <<~RUBY.gsub(/^/, "  ")
+      if ENV["RAILS_ENV"] == "development"
+        puts "\\n== Loading development seed data =="
+        system! "bin/rails development:db:seed"
+      end
```
stevepolitodesign: So, I was testing this locally, and it doesn't appear `ENV["RAILS_ENV"]` is set, which makes sense. But, then I realized that doesn't matter because [the task returns early](https://github.com/thoughtbot/suspenders/blob/f38407a4fe0f3cdfcd86defba88a026caed4682c/lib/templates/lib/tasks/development.rake#L1) if it's not in the `development` environment. 

```suggestion
      puts "\\n== Loading development seed data =="
      system! "bin/rails development:db:seed"
```

I'd like to test this locally first though just to make sure it won't break if run in a deployed environment.

jaredlt: Oh huh, you're right. Shame on me. I saw it generated the code I wanted but then didn't test it actually seeded a seed 🤦 

I actually added the `if ENV["RAILS_ENV"] == "development"` in `bin/setup` as double protection but I don't think it needs it. I think we could remove it and then rely on the checks in the rake task and the Seeder class. The one thing to change will be the [raise on non dev environments](https://github.com/thoughtbot/suspenders/blob/f38407a4fe0f3cdfcd86defba88a026caed4682c/lib/templates/lib/development/seeder.rb#L11), instead we would want some form of no-op I think.

I'll take another pass at this, maybe next week. But if you have any more thoughts let me know!

stevepolitodesign: @jaredlt sorry, I've gone full circle on this. Your original implementation was correct.

This mirrors how [Rails defaults its environment to `"development"`](https://github.com/rails/rails/blob/c69ec7ad963315b25f35c7156ecc4d0a68590189/railties/lib/rails.rb#L74-L76).

```suggestion
      if ENV.fetch("RAILS_ENV", "development") == "development"
        puts "\\n== Loading development seed data =="
        system! "bin/rails development:db:seed"
      end
```

jaredlt: What is old is new again :)

PR #1355, opened 2026-05-08
Add Letter Opener

jaredlt:
It is inevitable that an application will send email. In development, we don't want to actually send emails because a) we might accidentally send to a real person, and b) it's friction in the development flow to ensure you use an email you own, jump across to your email client, wait for it to be received etc.

This change introduces [letter_opener] in development only, so generated emails open automatically in the browser.

[letter_opener]: https://github.com/ryanb/letter_opener

---

I appreciate that Rails comes with [mailer interceptors](https://guides.rubyonrails.org/action_mailer_basics.html#intercepting-and-observing-emails) now but these need to be configured and the emails don't open in the browser which is a nice DX.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -8,6 +8,12 @@ Uses [Hotwire Spark][] for live reloading during development.
 
 [Hotwire Spark]: https://github.com/hotwired/spark
 
+### Letter Opener
+
+Uses [Letter Opener][] to preview emails in the browser instead of sending them.
+
+[Letter Opener]: https://github.com/ryanb/letter_opener
+
 ### Strong Migrations
 
 Uses [Strong Migrations][] to catch unsafe migrations in development.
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -91,6 +91,7 @@ PLATFORMS
   arm64-darwin-21
   arm64-darwin-23
   arm64-darwin-24
+  arm64-darwin-25
```
stevepolitodesign, line 94: Let's merge #1354 

```suggestion
```

stevepolitodesign (approved):
I think this is a nice addition. Just add this feature to `NEWS.md` before merging.

Gemfile.lock
jaredlt: Doh! How'd that sneak in here...!?

PR #1358, opened 2026-06-12
Add Prosopite for N+1 query detection

jaredlt:
Before, generated apps relied on `strict_loading_by_default` with the `:n_plus_one_only` mode for N+1 protection, which raises when an association is lazily loaded from a collection-loaded record. However, strict loading does not detect [all] N+1s.

Now, [Prosopite] scans web requests & Sidekiq jobs in development and test via its middleware. Detections are logged in development and raise `Prosopite::NPlusOneQueriesError` in test, so new apps fail their test suite on N+1 queries that strict loading can't see.

[all]: https://github.com/rails/rails/issues/56094
[Prosopite]: https://github.com/charkost/prosopite

Closes: https://github.com/thoughtbot/suspenders/discussions/1325

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -20,6 +20,13 @@ Uses [Strong Migrations][] to catch unsafe migrations in development.
 
 [Strong Migrations]: https://github.com/ankane/strong_migrations
 
+### Prosopite
+
+Uses [Prosopite][] to detect N+1 queries in development and test,
+complementing the built-in strict loading protection.
+
+[Prosopite]: https://github.com/charkost/prosopite
+
 ### Seed Data
 
 Follows [our guidance][seed-data-guide] for managing seed data. Use
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -22,8 +22,7 @@ Uses [Strong Migrations][] to catch unsafe migrations in development.
 
 ### Prosopite
 
-Uses [Prosopite][] to detect N+1 queries in development and test,
-complementing the built-in strict loading protection.
+Uses [Prosopite][] to detect N+1 queries in development and test.
 
 [Prosopite]: https://github.com/charkost/prosopite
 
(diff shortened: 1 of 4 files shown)


stevepolitodesign (commented):
I think we'll want to remove these lines too:

https://github.com/thoughtbot/suspenders/blob/25a6c4bb8d834400b52b2208b779fbdef9cec7d2/lib/templates/web.rb#L283-L284

FEATURES.md
stevepolitodesign, line 1: We'll also want to remove these lines

https://github.com/thoughtbot/suspenders/blob/25a6c4bb8d834400b52b2208b779fbdef9cec7d2/FEATURES.md?plain=1#L60-L61

We'll also wanna update `lib/templates/web.rb` by removing those lines too. 

https://github.com/thoughtbot/suspenders/blob/25a6c4bb8d834400b52b2208b779fbdef9cec7d2/lib/templates/web.rb#L410-L411

lib/templates/config/initializers/sidekiq.rb
```diff
@@ -7,6 +7,13 @@
 
 Sidekiq.configure_server do |config|
   config.redis = SIDEKIQ_REDIS_CONFIGURATION
+
+  unless Rails.env.production?
```
stevepolitodesign: Let's use `local?` instead. Otherwise, this code would run in a `staging` environment.

```suggestion
  if Rails.env.local?
```

lib/templates/web.rb
```diff
@@ -358,6 +381,12 @@ def update_readme
 
       [Strong Migrations]: https://github.com/ankane/strong_migrations
 
+      ### N+1 Query Detection
+
+      Uses [Prosopite][] to detect N+1 queries.
```
stevepolitodesign: [nit] Let's make this consistent with `FEATURES.md`


```suggestion
      Uses [Prosopite][] to detect N+1 queries in development and test.
```

stevepolitodesign (approved):
Sorry, just one more set of suggestions, then we're good!

jaredlt:
@stevepolitodesign thanks for taking another look at this! All feedback should be addressed now.

stevepolitodesign:
@jaredlt thank you for this. Just noting I need to cut a new release for this to actually take effect.

PR #1364, opened 2026-09-23
Add RuboCop plugins for linting specs in generated apps

DChavez18:
Generated apps only get Rails' default linting (rubocop-rails-omakase). This PR adds RuboCop plugins for RSpec, FactoryBot, and Capybara so specs get linted too.

Changes:
- Add rubocop-rspec, rubocop-factory_bot, and rubocop-capybara to generated apps
- Set NewCops: disable so plugin updates don't break CI
- Enable `FactoryBot/SyntaxMethods`
- Disable RSpec/ExampleLength and RSpec/MultipleExpectations.
- Add an expectation to the generated factories_spec.rb and fix typo
- Use before in the generated action_mailer support file
- Document the change in FEATURES.md, the generated README, and NEWS.md

I ran bin/qa: new apps and scaffolds lint clean, and specs pass.


stevepolitodesign:
I wonder if we should call `--skip-rubocop` (see this [PR](https://github.com/rails/rails/pull/50486)) so we're not even relying on that gem. That way, we have full control over our linting. I need to give it some thought though.

